PR #77, opened 2025-07-30
Optimize CI workflow by splitting into parallel jobs

cjroth:
## Summary
- Split frontend job into separate typescript and rust jobs for faster feedback
- Typescript job now runs independently without rust dependencies  
- Improved caching with actions/cache@v4 and built-in rust toolchain caching
- Added comprehensive cargo cache paths and OS-specific cache keys
- Added missing test steps (bun test, cargo test, cargo fmt check)

## Benefits
- Typescript failures will be caught in ~2-3 minutes instead of waiting for rust setup
- Rust caching is now much more efficient with proper paths and built-in toolchain caching
- Jobs run in parallel so total CI time should be reduced
- Better cache hit rates with OS-specific keys

## Test plan
- [x] Verify typescript job runs independently and fast
- [x] Verify rust job has proper caching
- [x] Verify all jobs pass successfully
- [x] Confirm overall CI time is reduced

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,9 +84,6 @@ jobs:
       - name: Type check
         run: bun run tsc --noEmit
 
-      - name: Run tests
-        run: bun test
-
   rust:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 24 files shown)


radishmouse (commented):
just chiming in that this looks :+1:

PR #92, opened 2025-08-13
THU-57 Background Processes

radishmouse:
Linear Task: https://linear.app/mozilla-thunderbolt/issue/THU-57/research-how-to-reliably-run-background-tasks-across-different

@cjroth here's the work so far on getting scheduled tasks working:

Initial dev done for linux:
* adds `schedules.tsx` component that appears under "Settings"
* adds command in `commands.rs` to run OS-appropriate shell script
* adds bash script for linux
* adds necessary `systemd` templates to `src-tauri/resources`
* configures Tauri to allow access to resources folder

Confirmed that:
1. clicking the "Install" button on the Schedules page calls the `run_schedule_installer` command with (hard-coded) arguments for description and time (hour/minute)
2. arguments are passed from `run_schedule_installer` to the shell script
3. shell script replaces placeholders with argument values and outputs result to user's `systemd` config folder

Next: 
1. update the shell script to run the commands to activate the new `systemd` units
2. implement script and create templates for macOS
3. implement script and create templates for Windows


cjroth:
Looking cool! Two things to put on your mind:

1. We'll probably want to gracefully hide this on platforms that don't support it such as web. That or somehow disable it and communicate to the user that it's not available in web. We might be able to pull in some design help to make this a good UX.
2. The Scheduled Tasks section that you've built will essentially be a list of the Automations that are set to run at a certain time. I've already added the UI for setting times for Automations but it is hidden at the moment (since it wasn't working yet). So basically there will be two places to configure/view these tasks (1) automations and (2) scheduled tasks.

Example of what I mean (but this is different because it's just a toggle whereas yours is  an entire section).
Screenshot 2025-08-13 at 11 23 06 AM

cjroth:
Closing this as it is getting stale - but would love to revisit it in the future when we have time :)

PR #95, opened 2025-08-14
Show confirmation dialog before clearing all chats or a single chat

### LINKED ISSUE #7: Show confirmation dialog before clearing all chats
*This issue was migrated from thunderbird/thunderbolt#105*

---

raivieiraadriano92:
## Description

This PR introduces **`DeleteAllChatsDialog `** and **`DeleteChatDialog `** components.

It is now used for:

- **Delete All Chats** — confirms before removing all chat history.  
- **Delete Single Chat** — confirms before removing an individual chat. *(This was not part of the original issue, but I added it for consistency and safety.)*  

The dialogs use **`useImperativeHandle`** to expose `open()` and `close()` methods, allowing parent components to trigger it imperatively. This approach avoids unnecessary re-renders in the parent when opening or closing the dialog.

---

## Render Preview

https://github.com/user-attachments/assets/13f089fb-a59d-4b8b-9f1c-ad5da02b0712

---

## Related Issue

Closes [#7](https://github.com/thunderbird/thunderbolt/issues/7)

### DIFF
--- a/src/components/delete-all-chats-dialog.tsx
+++ b/src/components/delete-all-chats-dialog.tsx
@@ -36,7 +36,7 @@ export const DeleteAllChatsDialog = forwardRef<DeleteAllChatsDialogRef, DeleteAl
       <AlertDialog open={open} onOpenChange={setOpen}>
         <AlertDialogContent>
           <AlertDialogHeader>
-            <AlertDialogTitle>Are you absolutely sure?</AlertDialogTitle>
+            <AlertDialogTitle>Delete all chats?</AlertDialogTitle>
             <AlertDialogDescription>This will permanently delete all your chats.</AlertDialogDescription>
           </AlertDialogHeader>
           <AlertDialogFooter>
--- a/src/components/delete-all-chats-dialog.tsx
+++ b/src/components/delete-all-chats-dialog.tsx
@@ -7,7 +7,7 @@ import {
   AlertDialogTitle,
   AlertDialogDescription,
   AlertDialogCancel,
-} from '@/components/ui/alert-dialog' // adjust import to your project
+} from '@/components/ui/alert-dialog'
 import { Button } from './ui/button'
 
 export type DeleteAllChatsDialogRef = {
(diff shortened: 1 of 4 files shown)


cjroth:
Nice first PR :)

Just thinking about this approach a bit. There will be other destructive dialogs in the app but I think that they will have different buttons and text, so I think this is more of an abstraction than necessary. I think we should just make individual alerts for each thing honestly. Also - just curious - why refs instead of state?

raivieiraadriano92:
> Nice first PR :)
> 
> Just thinking about this approach a bit. There will be other destructive dialogs in the app but I think that they will have different buttons and text, so I think this is more of an abstraction than necessary. I think we should just make individual alerts for each thing honestly. Also - just curious - why refs instead of state?

Hey @cjroth I replaced that DestructiveDialog with two individual components `DeleteAllChatsDialog` and `DeleteChatDialog`.

About the `ref` question, with this approach the `open` state lives in the dialog component avoiding re-render the parent component when it changes.

src/components/delete-all-chats-dialog.tsx
```diff
@@ -0,0 +1,56 @@
+import { forwardRef, useImperativeHandle, useState } from 'react'
+import {
+  AlertDialog,
+  AlertDialogContent,
+  AlertDialogHeader,
+  AlertDialogFooter,
+  AlertDialogTitle,
+  AlertDialogDescription,
+  AlertDialogCancel,
+} from '@/components/ui/alert-dialog' // adjust import to your project
```
cjroth: Remove comment

src/components/delete-all-chats-dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+    useImperativeHandle(ref, () => ({
+      open: () => setOpen(true),
+      close: () => setOpen(false),
+    }))
+
+    return (
+      <AlertDialog open={open} onOpenChange={setOpen}>
+        <AlertDialogContent>
+          <AlertDialogHeader>
+            <AlertDialogTitle>Are you absolutely sure?</AlertDialogTitle>
```
cjroth: How about just "Delete all chats?"

src/components/delete-all-chats-dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <AlertDialogDescription>
+              This action cannot be undone. This will permanently delete all your chats.
```
cjroth: Just "This will permanently delete all your chats." without the "This action cannot be undone."

src/components/delete-all-chats-dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+            </AlertDialogDescription>
+          </AlertDialogHeader>
+          <AlertDialogFooter>
+            <AlertDialogCancel onClick={handleCancel}>Cancel</AlertDialogCancel>
+            <Button variant="destructive" onClick={onConfirm}>
```
cjroth, line 44: "Delete All Chats"

cjroth (approved):
Looks good! Just a few minor copy changes / remove comment and then let's merge :)

src/components/delete-all-chats-dialog.tsx
raivieiraadriano92: done

src/components/delete-all-chats-dialog.tsx
raivieiraadriano92: done

src/components/delete-all-chats-dialog.tsx
raivieiraadriano92: done

src/components/delete-all-chats-dialog.tsx
raivieiraadriano92: done

PR #101, opened 2025-08-16
THU-25 Proof-of-concept sync with cr-sqlite + libsql

cjroth:
https://linear.app/mozilla-thunderbolt/issue/THU-25/create-proof-of-concept-sync-with-cr-sqlite-libsql

This adds https://github.com/vlcn-io/cr-sqlite to libsql in order to allow CRDT-based offline-first <> cloud database syncing. This is step one of many in order to get cross-device data syncing working.

This is experimental and may not get merged. The goal here is to see how well cr-sqlite works for data syncing compared to Turso's remote databases.

Tasks:
- [ ] Is there a way to end-to-end encrypt the data before syncing?
- [ ] Will this work with Turso's new "limbo" project (we plan on switching to that in the future)? If not, this could cause us to get locked into the old version of libsql which would be bad.


cjroth:
Commenting with a bit of an update here after more experimenting outside of this PR:
- We may want to use Powersync since it takes care of the backend + syncing part, unlike cr-sqlite
- If we do use cr-sqlite, we would need to figure out how to build cr-sqlite for turso-wasm in the browser, which might be difficult
- In my own experimenting, we will likely want to use cr-sqlite's wa-sqlite instead of sqlocal since it works out of the box with cr-sqlite enabled
- I believe 95% of the time, last-write-wins (LWW) is a reasonable conflict resolution approach, which is another reason why I think Powersync might be better than cr-sqlite here
- Futher, Powersync also has sdks beyond typescript, whereas whatever syncing logic we end up writing will need to be re-written in other languages if we ever build native apps

PR #111, opened 2025-08-21
Storybook setup

raivieiraadriano92:
## Description

This PR sets up **Storybook** in the Thunderbolt project to enable modular UI component development.  
With Storybook, front-end developers can now work on individual components in isolation, test across different screen sizes, and experiment with various property combinations without needing to recreate states in the main application.

Key updates:
- Installed and configured **Storybook** for the project.
- Added initial stories for the following components:
  - `components/ui/button`
  - `components/ui/input`
  - `components/delete-all-chats-dialog`
  - `components/delete-chat-dialog`
- Kept the default Storybook examples (`Example/Header`, `Example/Page`) to showcase **advanced usage** patterns.
- Updated the **README** with instructions on how to run and use Storybook.

Most of the files in this PR are the default Storybook setup files.

---

## Render Preview

https://github.com/user-attachments/assets/9ab2cc63-9975-437c-b422-aaec2a30c96a

---

## Related Issue

Closes [THU-23](https://linear.app/mozilla-thunderbolt/issue/THU-23/set-up-storybook)


vite.config.ts
```diff
@@ -1,10 +1,15 @@
+/// <reference types="vitest/config" />
 import tailwindcss from '@tailwindcss/vite'
 import react from '@vitejs/plugin-react'
 import path from 'path'
 import { defineConfig } from 'vite'
 import { analyzer } from 'vite-bundle-analyzer'
 import { bundleMigrations } from './src/db/bundle-migrations'
+import { fileURLToPath } from 'node:url'
```
cjroth, line 8: should this be `bun:url`?

src/stories/Page.stories.ts
```diff
@@ -0,0 +1,33 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
```
cjroth: Would it make sense to remove the examples?

cjroth (changes requested):
Looks good - basically just thinking we should remove the example stuff to avoid cluttering our repo

src/stories/Page.stories.ts
raivieiraadriano92: For sure

vite.config.ts
raivieiraadriano92: I don't think so.

`The bun:url package does not exist as a distinct, importable module in Bun. Instead, Bun provides a Node.js-compatible url module, which is part of its built-in Node.js compatibility layer.`

cjroth (approved):
Nice! Happy to have this :)

PR #120, opened 2025-08-21
Fix Google Drive Search

cjroth:


cjroth:
Closes https://linear.app/mozilla-thunderbolt/issue/THU-9/fix-failing-tool-calls-for-google-drive

PR #123, opened 2025-08-25
THU-41: Add telemetry settings toggle

raivieiraadriano92:
## Description

This PR introduces a **telemetry setting** in the Settings screen to allow users to control whether analytics data is collected via PostHog.  

Key updates:
- Added a new setting, `data_collection`, to enable/disable telemetry.  
- Implemented a simple **on/off toggle** in the Settings screen with explanatory copy about what data is collected and why.  
- Default state logic is `true` based on the team’s decision.  
- Telemetry toggle correctly controls PostHog’s enabled/disabled status using `usePostHog().is_capturing()`.  

Additionally:
- Created a new **Privacy Policy** page at `/privacy-policy`.  
  - ⚠️ A follow-up ticket is needed to provide and finalize the content for this page.  

---

## Render Preview

https://github.com/user-attachments/assets/bfa2e45d-50e8-46e1-9d1f-165978a01485

Screenshot 2025-08-25 at 19 19 21

---

## Related Issue

Closes [#THU-41](https://linear.app/mozilla-thunderbolt/issue/THU-41/add-telemetry-settings-toggle)


src/components/privacy-policy.tsx
```diff
@@ -0,0 +1,16 @@
+/**
```
cjroth: Actually we can remove this for now - I think we will likely create a website where this will go. For now, can you make it go to https://www.thunderbird.net/en-US/privacy/ and make sure it opens it in a new tab (for web) or opens user's browser (for tauri)?

src/settings/preferences.tsx
```diff
@@ -68,12 +77,15 @@ export default function PreferencesSettingsPage() {
...
       const lngData = await db.select().from(settingsTable).where(eq(settingsTable.key, 'location_lng'))
       const preferredNameData = await db.select().from(settingsTable).where(eq(settingsTable.key, 'preferred_name'))
+      const dataCollection = await db.select().from(settingsTable).where(eq(settingsTable.key, 'data_collection'))
 
       return {
         locationName: nameData[0]?.value || '',
         locationLat: latData[0]?.value || '',
         locationLng: lngData[0]?.value || '',
         preferredName: preferredNameData[0]?.value || '',
+        // if not explicitly set to false, default to true
```
cjroth: I think this code makes sense, probably don't need this comment

src/settings/preferences.tsx
```diff
@@ -186,6 +209,25 @@ export default function PreferencesSettingsPage() {
     },
   })
 
+  // Save data collection mutation
```
cjroth, line 210: I know this file is already full of comments like `// Save location mutation`. I'm trying to get AI to stop adding these comments since I think the code is self-explanatory and they don't really add much. So nothing wrong with this PR, just FYI. If you have a suggestion for updating cursor's AGENTS.md file or something like that, that would be cool

src/settings/preferences.tsx
```diff
@@ -433,6 +479,35 @@ export default function PreferencesSettingsPage() {
...
+          <form className="flex flex-col gap-2" onSubmit={(e) => e.preventDefault()}>
+            <FormField
+              control={privacyForm.control}
+              name="dataCollection"
+              render={({ field }) => (
+                <div className="flex-row flex items-center gap-4">
+                  <div>
+                    <label className="text-sm font-medium">Data Collection</label>
+                    <p className="text-sm text-muted-foreground">
+                      Help us improve the app by sending only anonymous usage info such as crashes, performance, and
```
cjroth: Now that I think about it... I think we should remove the word "only". It could be confusing

src/settings/preferences.tsx
```diff
@@ -433,6 +479,35 @@ export default function PreferencesSettingsPage() {
           </AlertDialog>
         </div>
       </SectionCard>
+
+      <div className="h-6" />
+
+      <SectionCard title="Privacy">
```
cjroth: Can we move the privacy section above the data reset section?

src/settings/preferences.tsx
```diff
@@ -433,6 +479,35 @@ export default function PreferencesSettingsPage() {
...
+                    <p className="text-sm text-muted-foreground">
+                      Help us improve the app by sending only anonymous usage info such as crashes, performance, and
+                      usage. No personal data is collected or stored. Read more about our{' '}
+                      <Link className="text-primary underline-offset-4 hover:underline" to="/privacy-policy">
+                        privacy policy
+                      </Link>
+                      .
+                    </p>
+                  </div>
+                  <Switch checked={field.value} onCheckedChange={handleDataCollectionToggle} />
```
cjroth: Not specific to this PR, but I would love to update the Switch component to show a `pointer` cursor instead of the arrow (default) one. It's weird to me that when you hover it, it doesn't appear with the pointer cursor and therefore almost feels like it's not clickable.

cjroth (changes requested):
Looks great, just a bunch of minor nitpicks basically. Nice work!

src/components/privacy-policy.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: created a symbolic link from CLAUDE.md to AGENTS.md

PR #130, opened 2025-08-27
THU-34: Token Tracking

cjroth:
https://linear.app/mozilla-thunderbolt/issue/THU-34/track-conversation-context-window-usage


src/chats/detail.tsx
```diff
@@ -55,34 +57,20 @@ export default function ChatDetailPage() {
...
-        .onConflictDoUpdate({
-          target: chatMessagesTable.id,
-          set: {
-            content: sql`excluded.content`,
-            parts: sql`excluded.parts`,
-            role: sql`excluded.role`,
-          },
-        })
+      // Fetch thread info to check if we need to generate a title
+      const thread = await db.select().from(chatThreadsTable).where(eq(chatThreadsTable.id, params.chatThreadId)).get()
```
raivieiraadriano92, line 64: In [this](https://github.com/thunderbird/thunderbolt/pull/131) it was implemented a new method called `getChatThreadById` in `lib/dal.ts`. I'll use that PR to refactor this line

src/components/context-overflow-modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+  isOpen: boolean
+  onClose: () => void
+  onNewChat: () => void
+  maxTokens?: number
+}
+
+/**
+ * Modal shown when a message would exceed the model's context window
+ */
+export const ContextOverflowModal = ({ isOpen, onClose, onNewChat, maxTokens }: ContextOverflowModalProps) => {
```
raivieiraadriano92, line 24: An interesting approach we could adopt for modals is use `forwardRef` and `useImperativeHandle` to handle with open/close state. This way we keep the state inside of the modal component - avoiding re-render in the parent component.
Example: src/components/delete-chat-dialog.tsx

src/lib/analytics.tsx
```diff
@@ -100,6 +100,7 @@ export const PostHogProvider = ({ children }: { children: ReactNode }) => {
 export type EventType =
   // Chat & Messaging
   | 'chat_send_prompt'
+  | 'chat_send_prompt_overflow'
```
raivieiraadriano92, line 103: Would be necessary to keep the TELEMETRY.md updated with all events we are tracking?

raivieiraadriano92 (commented):
LGTM

src/components/context-overflow-modal.tsx
cjroth: This is an interesting idea. I think I like that. I can see some pros and cons to both, but I think I like your idea better than this.

src/lib/analytics.tsx
cjroth: Good question! I forgot to update it - what if we actually don't store the list in TELEMETRY.md so that we don't have to keep it updated in 2 places, but instead, we can just link to this file from there

raivieiraadriano92: Agreed - the type itself already serves as a doc

PR #133, opened 2025-08-31
THU-45: Standardize code formatting

raivieiraadriano92:
Add automated code formatting and linting across the entire project.

### New Scripts
- `make format` - Format all code (frontend, Python, Rust)
- `make format-check` - Check formatting without changes
- `npm run format:python` - Format Python with Ruff
- `npm run format:rust` - Format Rust with cargo fmt

### Main Changes
- **Makefile**: Added formatting commands
- **package.json**: Added formatting scripts and pre-commit hooks
- **25+ code files**: Applied consistent formatting

### Key Feature
**Automatic pre-commit formatting**: Staged git code is automatically formatted before commits via lint-staged.

This ensures consistent code style across TypeScript, Python, and Rust codebases.


src/lib/oauth-webview.ts
```diff
@@ -80,39 +80,49 @@ export const startOAuthFlowWebview = async (
...
+
     const cleanup = async () => {
-      unlistenCallback()
-      unlistenNavigate()
+      if (unlistenCallback) unlistenCallback()
+      if (unlistenNavigate) unlistenNavigate()
       await window.destroy()
     }
 
     const handleCallback = (code: string | null, state: string | null, error: string | null) => {
```
cjroth: Can we use try/catch and await here and below? I think that would help make this more standard (also curious if there's a linting setting for that)

package.json
```diff
@@ -7,6 +7,9 @@
   "lint-staged": {
     "*.{ts,tsx,js,jsx,json,css,md}": [
       "bun run format"
+    ],
```
cjroth: Bun/npm/typescript stuff should be separate from python. I know it's a monorepo, but I think we should still try to separate the concerns here. I think the `Makefile` is the one place where we can centralize the code to call both frontend and backend - but I don't think package.json should have Python references in it personally

package.json
```diff
@@ -24,10 +27,12 @@
...
-    "format:python": "cd backend && uv run ruff format .",
-    "format:python-check": "cd backend && uv run ruff format --check .",
-    "lint:python": "cd backend && uv run ruff check --fix .",
-    "lint:python-check": "cd backend && uv run ruff check .",
+    "format:python": "cd backend && uv run ruff format . && cd ..",
+    "format:python-check": "cd backend && uv run ruff format --check . && cd ..",
+    "lint:python": "cd backend && uv run ruff check --fix . && cd ..",
+    "lint:python-check": "cd backend && uv run ruff check . && cd ..",
+    "format:rust": "cd src-tauri && cargo fmt && cd ..",
+    "format:rust-check": "cd src-tauri && cargo fmt --check && cd ..",
```
cjroth, line 19: I think putting rust here might be ok though since the rust IS part of the tauri / frontend stuff, unlike the python side

cjroth (changes requested):
This is great! I think it will definitely make our lives easier in code reviews going forward :) Just added a couple of minor things

cjroth:
Also question: where are the rust/python linting/formatting settings getting configured?

package.json
robbyrob42: yes, completely agree, using bun for this is novel and cool but I 100% agree, python should be kept completely separate and it should use `ruff` to format.

src/lib/oauth-webview.ts
raivieiraadriano92: Yes, there is an eslint rule called [no-async-promise-executor](https://eslint.org/docs/latest/rules/no-async-promise-executor) - due this rule I did the current change.

Commit: [ad0b82](https://github.com/thunderbird/thunderbolt/pull/133/commits/ad0b82f7b1518d6117c01fed40cde8a465712264)

We can disable that rule a rollback to the previous implementation which is already using wait.

cjroth: Makes sense! I think it's fine in that case. Although, a more readable way of creating promises imo is just to use async in the first place

raivieiraadriano92:
> Also question: where are the rust/python linting/formatting settings getting configured?

For Python we have the file `backend/pyproject.toml` -> line 38 `[tool.ruff.lint]`

For Rust: `src-tauri/rust-toolchain.toml`

package.json
raivieiraadriano92: Done.

src/lib/oauth-webview.ts
raivieiraadriano92: Done

cjroth (approved):
This is amazing!

PR #134, opened 2025-09-01
THU-(38|42): Add preview features settings section

raivieiraadriano92:
## Description

This PR adds a **Preview Features** section to the Settings screen, allowing users to opt into experimental features.  
- Includes toggles for **Tasks** and **Automations**, disabled by default.  
- If telemetry is disabled, enabling a preview feature triggers a modal requiring users to enable telemetry first.  
- Selecting "Enable telemetry" enables both telemetry and the chosen feature.  
- Navigation items for preview features are feature-gated by toggle state, while the underlying code remains intact.  

---

## Related Issue

Closes
- [THU-38](https://linear.app/mozilla-thunderbolt/issue/THU-38/add-preview-features-settings-section)
- [THU-42](https://linear.app/mozilla-thunderbolt/issue/THU-42/make-tasks-a-preview-feature)


cjroth:
This is a good start!

I just realized that the Linear ticket is a bit confusing. There is actually not meant to be a generic toggle for preview features - rather, the Automations feature is meant to be the first feature that has its own toggle. Now that this is built, I think it would make sense to also squeeze https://linear.app/mozilla-thunderbolt/issue/THU-42/make-tasks-a-preview-feature into the same PR. Sorry for the confusion here - I could have done a better job at writing this ticket to be more clear.

A couple other minor UI things:
- We should put this above the "Local Database" and "Privacy" sections IMO
- If the user turns off telemetry and has preview features enabled, I think we should show a modal warning them that the preview features will be disabled as well.
- Now that I see it visually, I think we should skip the yellow warning label - feels a bit much

Not related to this PR but would be nice to fix while we're working on this:
- We should make sure the spacing between each section is the same
- There should be a bit more padding at the bottom of the page so that the last section isn't so tight against the bottom of the viewport

Screenshot 2025-09-02 at 11 06 46 AM

src/app.tsx
```diff
@@ -79,8 +81,22 @@ function AppRoutes({ initData }: { initData: InitData }) {
         <Route element={<ChatLayout />}>
           <Route index element={<Navigate to={`/chats/${initData.initialThreadId}`} replace />} />
           <Route path="chats/:chatThreadId" element={<ChatDetailPage />} />
-          <Route path="tasks" element={<TasksPage />} />
-          <Route path="automations" element={<AutomationsPage />} />
+          <Route
```
cjroth: In the ideal world, I would want this to show a placeholder page that says something like "The <feature> feature is not enabled." with a button to go to the settings page and enable it. But I think this is fine for now!

src/settings/use-preview-features.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  return {
+    form,
+    savePreviewFeaturesMutation,
+    handleFeatureToggle,
+    handleBulkSave,
+    disableAllFeatures,
+    features: PREVIEW_FEATURES,
+  }
```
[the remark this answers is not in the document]
cjroth: I think this is a legitimate problem

src/settings/preview-features-section.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const PreviewFeaturesSection: React.FC<PreviewFeaturesSectionProps> = ({ settings, onFeatureToggle }) => {
+  const { form, features } = usePreviewFeatures(settings)
+
+  const handleToggle = async (featureName: keyof PreviewFeaturesFormData, value: boolean) => {
+    return await onFeatureToggle(featureName, value)
+  }
+
+  return (
+    <SectionCard title="Preview Features">
+      <p className="mb-4 text-sm text-muted-foreground">
```
cjroth: Let's just say `Try out experimental beta features.` I think the rest is implied by the fact that they are beta features.

src/settings/preferences.tsx
```diff
@@ -67,6 +69,9 @@ export default function PreferencesSettingsPage() {
   const [locations, setLocations] = React.useState<LocationData[]>([])
   const [isSearching, setIsSearching] = React.useState(false)
   const [isResetting, setIsResetting] = React.useState(false)
+  const [showTelemetryModal, setShowTelemetryModal] = React.useState(false)
```
cjroth: I just noticed this - not related to this PR, but we should use `useState` not `React.useState` to be consistent with the rest of the codebase here (this is my fault). Just wanted to say it while it's fresh on my mind.

src/settings/preferences.tsx
```diff
@@ -67,6 +69,9 @@ export default function PreferencesSettingsPage() {
...
+  const [showTelemetryWarningModal, setShowTelemetryWarningModal] = React.useState(false)
+  const pendingFeatureToggle = React.useRef<string | null>(null)
```
cjroth: Not sure that I understand why this one is a ref instead of useState.

src/settings/preferences.tsx
```diff
@@ -67,6 +69,9 @@ export default function PreferencesSettingsPage() {
   const [locations, setLocations] = React.useState<LocationData[]>([])
   const [isSearching, setIsSearching] = React.useState(false)
   const [isResetting, setIsResetting] = React.useState(false)
+  const [showTelemetryModal, setShowTelemetryModal] = React.useState(false)
+  const [showTelemetryWarningModal, setShowTelemetryWarningModal] = React.useState(false)
```
cjroth: I think this page's state would be much cleaner if we switched to useReducer instead of useState given how many useStates we have here. In my experience, prompting the AI to rewrite this with useReducer generally works pretty well on the first try

src/settings/preferences.tsx
```diff
@@ -526,6 +564,60 @@ export default function PreferencesSettingsPage() {
...
+          <AlertDialogHeader>
+            <AlertDialogTitle>Telemetry Required</AlertDialogTitle>
+            <AlertDialogDescription>
+              In order to use preview features, we ask that you help us improve the product by sharing telemetry data.
+            </AlertDialogDescription>
+          </AlertDialogHeader>
+          <AlertDialogFooter>
+            <AlertDialogCancel>Cancel</AlertDialogCancel>
+            <AlertDialogAction
+              onClick={async () => {
```
cjroth: This should be defined above the render section like `const handleEnableTelemetry = `

src/settings/preferences.tsx
```diff
@@ -526,6 +564,60 @@ export default function PreferencesSettingsPage() {
           </AlertDialog>
         </div>
       </SectionCard>
+
+      {/* Telemetry Required Modal */}
```
cjroth: I wouldn't mind if this were in a separate component file, but don't think it's bad like this either

src/settings/preferences.tsx
```diff
@@ -526,6 +564,60 @@ export default function PreferencesSettingsPage() {
...
+          <AlertDialogHeader>
+            <AlertDialogTitle>Preview Features Will Be Disabled</AlertDialogTitle>
+            <AlertDialogDescription>
+              Turning off telemetry will disable all preview features. Are you sure you want to continue?
+            </AlertDialogDescription>
+          </AlertDialogHeader>
+          <AlertDialogFooter>
+            <AlertDialogCancel>Cancel</AlertDialogCancel>
+            <AlertDialogAction
+              onClick={async () => {
```
cjroth: Same here - should be defined above

src/app.tsx
cjroth: Think a little more here... I think the current implementation is unnecessarily complicated. I think we can simplify this by just grabbing the status of the feature in a hook above and then wrapping it in an if statement. I'm not sure if I see the value in a new ExperimentalFeatureRouteWrapper component unless we are going to implement the page I described above

src/settings/use-preview-features.ts
```diff
@@ -0,0 +1,139 @@
+import React from 'react'
```
cjroth: Imo, the value that this hook adds doesn't justify the complexity. I feel like the saving logic should be handled in the settings page and doesn't require a hook. And I think the settings should just use `useBooleanSetting` directly.

That said, I DO think we should integrate this with PostHog's "early access features" feature like you described. If we do that, I do think it might make sense to abstract that logic since in that case maybe we are checking multiple locations for the source of truth.

src/settings/preview-features-config.ts
```diff
@@ -0,0 +1,27 @@
+export enum PreviewFeatureDatabaseKey {
```
cjroth: Similar to the above comment - I think this just feels like unnecessary abstraction. I don't think we will have enough features that I see a benefit to abstracting this when it could just be hardwired strings like the other settings.

cjroth (changes requested):
Looks good. Tbh I think I just want it to be a bit simpler overall, but it's the right idea. Nice work!

cjroth:
Oh, forgot to add also:

- After I disabled the Tasks feature, the LLM is still calling the "get_tasks" tool. It should not have access to the tasks tools anymore when the feature is disabled.

src/app.tsx
raivieiraadriano92: done

src/settings/preview-features-section.tsx
raivieiraadriano92: done

src/settings/preferences.tsx
raivieiraadriano92: Usually I use useRef when I need to hold a value which doesn't require a re-render - which is the case here.
However with the useReducer refactor that value was moved to the reducer state for better organization.

src/settings/preferences.tsx
raivieiraadriano92: Totally agree - done

src/settings/preferences.tsx
raivieiraadriano92: Agreed - it was implemented a new component. It's cleaner now

src/settings/preferences.tsx
raivieiraadriano92: as above - a new modal component was implemented

src/settings/preferences.tsx
raivieiraadriano92: I'll run AI in another branch to refactor the entire project - this way we avoid mix things here

src/settings/use-preview-features.ts
raivieiraadriano92: I'm working on PostHog's "early access features" implementation in a separate branch - so we keep this PR smaller

src/settings/preferences.tsx
```diff
@@ -230,6 +320,49 @@ export default function PreferencesSettingsPage() {
     },
   })
 
+  // Save preview features mutation
+  const savePreviewFeaturesMutation = useMutation({
+    mutationFn: async (values: z.infer<typeof previewFeaturesFormSchema>) => {
+      // Save each feature setting
+      await db
```
cjroth: I feel like we could make a DAL function like setBooleanSetting('experimental_feature_automations', false) or something like that

src/lib/dal.ts
```diff
@@ -151,13 +151,23 @@ export const getPreferencesSettings = async () => {
...
+    .where(eq(settingsTable.key, 'experimental_feature_tasks'))
 
   return {
     locationName: nameData[0]?.value || '',
     locationLat: latData[0]?.value || '',
     locationLng: lngData[0]?.value || '',
     preferredName: preferredNameData[0]?.value || '',
     dataCollection: dataCollection[0]?.value === 'false' ? false : true,
+    experimentalFeatureAutomations: experimentalFeatureAutomations[0]?.value === 'false' ? false : true,
+    experimentalFeatureTasks: experimentalFeatureTasks[0]?.value === 'false' ? false : true,
```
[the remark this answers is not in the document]
cjroth: I think I agree with this that they should default to false. We could do that as part of the next PR though

cjroth (approved):
Looks great!! Love this.

src/settings/preferences.tsx
raivieiraadriano92: Totally agree - to follow the naming pattern we current have it was implemented `updateBooleanSetting`

src/settings/preferences.tsx
```diff
@@ -230,6 +314,30 @@ export default function PreferencesSettingsPage() {
     },
   })
 
+  // Save preview features mutation
+  const savePreviewFeaturesMutation = useMutation({
+    mutationFn: async (values: z.infer<typeof previewFeaturesFormSchema>) => {
+      // Save each feature setting
+      updateBooleanSetting('experimental_feature_automations', values.experimentalFeatureAutomations)
+      updateBooleanSetting('experimental_feature_tasks', values.experimentalFeatureTasks)
+    },
```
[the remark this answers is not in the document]
raivieiraadriano92: done

PR #136, opened 2025-09-02
THU-10 Proxy fetch content tool for user privacy

robbyrob42:
draft pr


cjroth:
Just a heads up - we'll want to split the markdown docs stuff into a separate PR so we can review and merge it separately

robbyrob42:
> Just a heads up - we'll want to split the markdown docs stuff into a separate PR so we can review and merge it separately

ah ok, i'll separate it.

robbyrob42:
THU-10 Ready for Review, Thank you!

robbyrob42:
Added WebContentFetcher fallback to fetch-content endpoint. Uses Exa proxy when available, falls back to direct fetching when unavailable or failing. Tests added for both scenarios.

backend/pro/exa_content_fetcher.py
```diff
@@ -0,0 +1,128 @@
+"""
```
cjroth: I think we should just have one class for exa - not separate fro content and search.

src/ai/fetch.ts
```diff
@@ -203,13 +203,14 @@ export const aiFetchStreamingResponse = async ({
           console.groupEnd()
```
cjroth: We need to get this PR to only show the changes relevant to your branch - this is hard to review

cjroth:
Hey - I think this will be ready for review soon, but it's not there yet:
 - PR description should link to Linear task or be more descriptive
 - Checks should be passing
 - Merge conflicts should be resolved
 - There should only be changes that are part of this branch when I click on the "Files Changed" tab
 - Reviewer should be tagged in the top right
 - Should be marked as "Ready for review" so that it's not in draft state

cjroth (changes requested):

backend/pro/exa_content_fetcher.py
robbyrob42: Reasonable, I almost went that direction but wanted to get it working. I'm happy to refactor

src/ai/fetch.ts
robbyrob42: yeah. agreed, apologies, i'll get my process tightened up.

robbyrob42:
@cjroth all corrections implemented as requested. I grok the process and expectation now, and this should be easier for you to consume. I appreciate your feedback.

backend/pro/routes.py
```diff
@@ -70,14 +69,29 @@ async def search_exa_endpoint(request: SearchRequest) -> SearchResponse:
     async def fetch_content_endpoint(
         request: FetchContentRequest,
     ) -> FetchContentResponse:
-        """Fetch and parse content from a webpage URL"""
-        try:
-            ctx = SimpleContext()
-            content = await fetcher.fetch_and_parse(request.url, ctx)
```
cjroth, line 76: Just realized - we need to remove the old `WebContentFetcher` code since it's no longer used as well

cjroth (changes requested):

backend/pro/exa.py
```diff
@@ -106,8 +150,126 @@ async def search(
...
+                return f"Error: Content fetch service error ({e.response.status_code})"
+        except httpx.TimeoutException:
+            await ctx.error("Exa content fetch request timed out")
+            return "Error: The request timed out while trying to fetch the webpage."
+        except Exception as e:
+            await ctx.error(f"Unexpected error during content fetch: {str(e)}")
+            traceback.print_exc(file=sys.stderr)
+            return f"Error: An unexpected error occurred while fetching the webpage ({str(e)})"
+
+    def format_search_results_for_llm(self, results: list[dict[str, Any]]) -> str:
```
cjroth: Do we really want to keep this? I thought exa was basically intended for use by LLMs and would return results in more or less a good format for LLMs as it is?

backend/pro/exa.py
```diff
@@ -106,8 +150,126 @@ async def search(
...
+
+        Raises:
+            httpx.HTTPStatusError: For API authentication or rate limit errors
+            httpx.TimeoutException: If the request times out
+            Exception: For other unexpected errors during content fetching
+        """
+        try:
+            headers = {"x-api-key": self.api_key, "Content-Type": "application/json"}
+
+            # Exa contents request payload - only the URL is sent to Exa
```
cjroth: I think we should clean up all of these comments - I don't think they add much value but are adding visual clutter

cjroth (changes requested):

backend/pro/exa.py
```diff
@@ -1,129 +1,117 @@
...
 """
 
-import sys
-import traceback
-from dataclasses import dataclass
+import os
 from typing import Any
 
-import httpx
+from exa_py import Exa
```
cjroth, line 8: OH, good call. Idk why we didn't just do that in the first place haha

robbyrob42:
@cjroth moved to exa-labs/exa-py package. 
They: 
- format search results properly, 
- handle error handling and retry without effort from us, 
- enable unit tests without API key present so that we don't have to mock the api, and 
- have other tool classes and methods for future extension. 

sorry for the multiple reviews but this is clearly the correct ending location for us

backend/pro/exa.py
cjroth: Actually now that we're just using Exa's actual client - any reason we still need this exa class at all? I feel like we could just create and call the exa instance directly from the routes

robbyrob42: Yeah, this is me being too risk-averse to change while I was learning how the product works, but now that I am fully familiar, doing these things manually is clearly the most risky option. 

I'm open to removing the abstraction. I'll review this closely and see if there's any significant benefit to staying abstracted away from routes.py and either make the change or comeback for a chat

robbyrob42: the thin wrapper has a minimal cost overhead and, while not a ton of benefit right now, when we are adding integration environments or bringing in a new exa tool in the future, it will keep the route handling and functional operations separated. most other tools have a wrapper so that keeps the pattern consistent. suggest we keep for now.

PR #137, opened 2025-09-02
THU-33 DX and Agents Context Improvement

robbyrob42:
draft pr


src/components/oauth-callback.tsx
```diff
@@ -1,56 +1,66 @@
+import Loading from '@/loading'
```
cjroth: Was this change intentional?

.claude/user-stories/THU-10.md
```diff
@@ -0,0 +1,15 @@
+# Proxy "fetch content" tool to protect user privacy
```
cjroth: This is interesting! Initially I experimented with checking in prompts / stories, but I found that they added a lot of clutter to the codebase in the long run without adding a lot of value and it was hard to keep them in-sync with the issues in Linear / GitHub Issues. Tbh I would rather not check these in after doing that experiment.

.claude/agents/analysis-tasks/task-1-project-discovery-architecture.md
```diff
@@ -0,0 +1,71 @@
...
+Analyze the codebase structure and create a high-level architectural summary.
+
+## **Prerequisites**
+
+- Access to project root directory
+- Ability to read configuration files and source code
+
+## **Input Requirements**
+
+- Project root path: `/Users/robertjacques/RootsystemProjects/MozillaThunderbolt/thunderbolt`
```
cjroth: We should remove the user-specific path / PII here and use something generic or replace it with a command like `$(pwd)`

.claude/agents/analysis-tasks/task-1-project-discovery-architecture.md
```diff
@@ -0,0 +1,71 @@
...
+- Main business domain (if known): `Mozilla Thunderbird extension/integration`
+
+## **Task Instructions**
+
+```
+Analyze this codebase structure at /Users/robertjacques/RootsystemProjects/MozillaThunderbolt/thunderbolt. Please examine and document:
```
cjroth: And here

.claude/agents/analysis-tasks/task-2-code-conventions-patterns.md
```diff
@@ -0,0 +1,68 @@
...
+Document coding conventions, naming patterns, and architectural patterns used throughout the codebase.
+
+## **Prerequisites**
+
+- Completed Task 1 (Architecture Overview)
+- Access to source code files
+
+## **Input Requirements**
+
+- Project root path: `/Users/robertjacques/RootsystemProjects/MozillaThunderbolt/thunderbolt`
```
cjroth: Actually I would just do relative paths

cjroth (commented):
I think this is a really cool idea. I want to experiment with it for awhile before I'm convinced that it's definitely a good idea. I think the biggest concern is just that it needs to be kept updated which essentially means after every PR gets merged into main.

I'd put all of it into a direct `agents` folder instead of being nested under `.claude` to keep it more provider agnostic. I'd also want to add a README explaining how it works / how to use it / how to generate it, a script or `make` command to run it, and a GitHub CI action that runs it automatically on the main branch.

robbyrob42:
I've got some new ideas and this PR is going to continue to age, i'm going to keep the branch but close this pr for now, i'll reopen a draft PR when I get a chance to put those changes through their paces. thx!

PR #141, opened 2025-09-04
Add PostHog feature flags for preview features

raivieiraadriano92:
# Summary
Implements PostHog feature flags to control visibility of Tasks and Automations preview features. Features now require both a remote flag AND local user setting to be enabled.

### Changes
- **New hook**: `usePreviewFeature()` - combines PostHog flags with local settings
- **Route protection**: Tasks/Automations routes only render when enabled
- **UI updates**: Menu items hidden when features are disabled  
- **Tool integration**: AI tools conditionally loaded based on feature state
- **Settings**: Preview toggles only shown when PostHog flags are active

### How it works
```typescript
const enabled = posthog.isFeatureEnabled('tasks') && userSetting === true
```

### Benefits
- Controlled feature rollouts via PostHog dashboard
- Kill switch for instant feature disabling
- Clean UX - hidden features don't clutter interface
- Maintains existing user opt-in requirements


src/settings/preferences.tsx
```diff
@@ -622,47 +622,51 @@ export default function PreferencesSettingsPage() {
...
+                control={previewFeaturesForm.control}
+                name="experimentalFeatureAutomations"
+                render={({ field }) => (
+                  <div className="flex-row flex items-center gap-4">
+                    <div className="flex-1">
+                      <label className="text-sm font-medium">Automations</label>
+                    </div>
+                    <Switch
+                      checked={field.value}
+                      onCheckedChange={async (value) => {
```
cjroth: This should be defined above as a function like `const handleSomeUserEvent =`

src/hooks/use-preview-feature.ts
```diff
@@ -0,0 +1,29 @@
+import posthog from 'posthog-js'
```
cjroth: I don't think it's necessary to have a custom getPreviewFeatures function to check Posthog in place of the getBooleanSetting feature tbh, because there's no way that a user would have enabled the setting in the first place since we're already feature-gating the toggles in the settings page based on PostHog.

I think this is sort of like a secondary check that would only be necessary if somehow users enabled the feature in their database despite not getting access to it in Posthog. But I don't think we need to worry about that since at that point the feature is already visible in the codebase anyways.

I could be missing something so lemme know if you're thinking of something that I'm overlooking!

cjroth (changes requested):

src/hooks/use-preview-feature.ts
raivieiraadriano92: So - this second check would cover the scenario where:

- we enable the feature in PostHog
- user enable the feature in his local settings
- we want to globally disable the feature.

If we don't validate both PostHog and Local Setting we can't disable that feature once the user has enabled locally. I mean, the toggle will not appear to the user, however the settings will remain enabled in the local database.

src/settings/preferences.tsx
raivieiraadriano92: done

src/hooks/use-preview-feature.ts
cjroth: Hmmm ok interesting. I feel like it would be really weird for us to take away a feature after we already added it - might piss of users? But I also kind of see what you mean - like if we accidentally released something prematurely or realize it's not ready after it is live.

cjroth: So basically this is a question of: "should disabling a preview feature in posthog disable it for users who already enabled it?"

raivieiraadriano92: As we talked earlier - we won't remove preview features after the user enabled them.

PR #142, opened 2025-09-05
THU-7: Revamp seed data approach

raivieiraadriano92:
## Problem
The current seed data algorithm automatically creates seed data whenever tables are empty, causing unexpected behavior where deleted sample data gets recreated and system models can't be updated.

## Solution
- **Migration-based seeding**: Seed data is now only created during initial app setup through SQL migrations
- **System model protection**: Users cannot delete, edit, or disable system models through the UI
- **Conflict handling**: Uses `INSERT OR IGNORE` to handle existing data gracefully during upgrades
- **Comprehensive tests**: Added 8 test cases covering migration behavior and system model protection

## Key Changes
- Populated empty seed migration files with actual seed data
- Removed automatic seeding logic from app initialization
- Added system model protection functions and UI restrictions
- Updated all seed migrations to use `INSERT OR IGNORE` for safe upgrades

## Benefits
- ✅ Seed data only created during initial setup
- ✅ System models can be updated through new migrations
- ✅ Deleted sample data won't be automatically recreated
- ✅ Existing users can upgrade without conflicts


src/drizzle/0005_seed-settings.sql
```diff
@@ -0,0 +1,8 @@
+-- Seed default settings
+-- These settings are inserted with ON CONFLICT DO NOTHING to avoid overwriting existing values
+INSERT OR IGNORE INTO settings (key, value) VALUES
+('cloud_url', 'https://api.thunderbolt.ai'),
```
cjroth, line 4: This is the wrong url

src/drizzle/0004_seed-models.sql
```diff
@@ -0,0 +1,9 @@
+-- Seed system models
+-- Using hardwired IDs to ensure consistency across installs
+INSERT OR IGNORE INTO models (id, name, provider, model, is_system, enabled, is_confidential, context_window, tool_usage) VALUES
+('0198ecc5-cc2b-735b-b478-785b85d3c731', 'Qwen 3', 'flower', 'qwen/qwen3-235b', 1, 1, 1, 256000, 1);
```
cjroth, line 4: Just checking - did we generate this UUIDs completely randomly, or did AI generate them? If AI, we should probably regenerate in case they came from training data

src/drizzle/0005_seed-settings.sql
```diff
@@ -0,0 +1,8 @@
...
+('anonymous_id', '0198ecc5-cc2b-735b-b478-000000000001'),
```
cjroth, line 5: Hmm anonymous Id will need to be generated uniquely and securely for each user

src/drizzle/0008_seed-prompts.sql
```diff
@@ -0,0 +1,7 @@
+-- Seed default prompts
+-- Using hardwired IDs to ensure consistency across installs
+-- These prompts reference the system model seeded in 0004_seed-models.sql
```
cjroth, line 3: I think the original prompts were lost - these don't look right

src/lib/system-model-protection.ts
```diff
@@ -0,0 +1,9 @@
+import { getModelById } from './dal'
```
cjroth, line 1: This is... interesting.

src/settings/models/detail.tsx
```diff
@@ -77,6 +78,7 @@ export default function ModelDetailPage() {
 
   const updateModelMutation = useMutation({
     mutationFn: async (model: Partial<Model> & { id: string }) => {
+      await checkSystemModelProtection(model.id, 'edit')
```
cjroth, line 81: This is kind of an odd way of doing it. I don't think it's a bad idea to add a check before deleting the models at the db level, but I think it's odd to build a custom function just for that purpose. I would just move these update and delete functions into the dal.ts where all the other database logic is and then add an if (isSytem) throw Error in there.

cjroth (changes requested):
Ok, now that I see what the changes look like here, I think there are pros and cons that I didn't think about yet. My concern is that this will be difficult to maintain / update if we want to change the system models or system prompts.

I think it is a bit painful that we will need to write custom SQL migrations every time we want to tweak the system prompts or models and that new users opening the app for the first time will technically have their database go through all past migrations before finally reaching the desired states.

Maybe we could do some research on best practices and how other apps do this before committing to this approach?

Regarding the quality of the migrations here: we need to be more careful when self-reviewing the code. We need to be careful not to let these kind of unintended AI-generated code changes in, such as it changing the prompts. Please be more rigorous when reviewing code from AI.

src/drizzle/0004_seed-models.sql
cjroth: Actually I think these were hardwired - as long as they match the original ones

cjroth:
This was an awesome first pass at this! I'm closing this in favor of #183 now that we've had some time to think through different approaches. Super appreciate the initial work you did here.

PR #144, opened 2025-09-08
THU-50: Revamp Settings -> Models page

raivieiraadriano92:
### Changes
- Removed automatic tool support check (no more test prompt requirement).
- Added **“Supports tools”** checkbox when adding/editing a model.  
  - Default state is based on existing logic, but users can override it.
- Removed **Flower** provider type (now part of Thunderbolt).
- Added **Anthropic** as a provider option (using AI SDK provider).
- Updated model list:
  - Show first letter of model name as icon.
  - Show lock icon for encrypted models.
- Removed status validation on adding a model

### Preview
image

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
       "name": "thunderbolt",
       "dependencies": {
         "@ai-sdk-tool/parser": "2.0.12",
+        "@ai-sdk/anthropic": "^2.0.15",
         "@ai-sdk/openai": "^2.0.19",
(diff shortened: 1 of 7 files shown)


cjroth (changes requested):
Looks great and seems to work well!

Couple minor tweaks:
- Could we move the supports tools checkbox to be the very last item in the form
- Could we change the text from "Supports tools" to "Enable tool use"
- Could we add a tooltip when you hover over the lock icon that just says "Encrypted"

cjroth (approved):

PR #150, opened 2025-09-11
THU-49: Fix conversation lag in long chats

raivieiraadriano92:
Wrapped message component with React.memo to avoid re-renders


cjroth (approved):
This definitely improves the lag in long conversations during streaming.

cjroth (approved):

PR #154, opened 2025-09-12
THU-64: Fix infinite loading loop when incomplete tool call states exist

cjroth:
https://linear.app/mozilla-thunderbolt/issue/THU-64/fix-infinite-loading-loop-bug-when-there-are-incomplete-tool-calls

This PR fixes the infinite loading loop described in THU-64, but currently it has a bug that causes the LLM to respond to the first message twice:
Screenshot 2025-09-11 at 5 14 14 PM


cjroth:
Closing in favor of other approaches / PRs for now

PR #155, opened 2025-09-12
THU-76: Enable PostHog error tracking

raivieiraadriano92:
## Changes
- **CI Workflow**: Combined source map upload into main CI workflow, only runs on main branch pushes (not PRs)
- **Source Maps**: Vite config already has `sourcemap: true` enabled for production builds
- **Error Tracking**: PostHog will now capture exceptions with proper source map support for better debugging

## Benefits
- Reuses existing bun installation in CI (no duplicate setup)
- Source maps only uploaded for main branch (saves CI time on PRs)
- Better error tracking with readable stack traces in PostHog
- Simplified workflow management (removed separate sourcemap workflow)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,19 @@ jobs:
           # TODO: Investigate Bun module mocking behavior in CI environment
           bun test --test-name-pattern="^(?!.*Google Utils).*$"
 
+      # Upload source maps to PostHog only for main branch
+      - name: Build application with source maps
+        run: bun run build
+        env:
+          NODE_ENV: production
+
+      - name: Inject & upload source maps to PostHog
+        uses: PostHog/upload-source-maps@v0.4.6
+        with:
+          directory: dist
+          env-id: ${{ secrets.POSTHOG_CLI_ENV_ID }}
+          cli-token: ${{ secrets.POSTHOG_CLI_TOKEN }}
+
   rust:
     needs: detect-changes
     if: needs.detect-changes.outputs.rust == 'true'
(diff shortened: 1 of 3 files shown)


.github/workflows/posthog-sourcemaps.yml
```diff
@@ -0,0 +1,69 @@
+name: Upload Source Maps to PostHog
+
+on:
+  push:
+    branches:
+      - raivieiraadriano92/thu-76-enable-posthog-error-tracking
```
cjroth: I think it's safe to not specify this branch here since these config files wouldn't exist in main until we merge anyways - but better safe than sorry

cjroth (changes requested):
From PostHog's docs, it looks like they published an official PostHog GH Action that we can use to simplify this: https://posthog.com/docs/error-tracking/upload-source-maps/github-actions

For the build failures - I double checked and the env vars should be correct and have correct permissions - is it possible that we need to set the project name maybe?

.github/workflows/posthog-sourcemaps.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  build_and_upload:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Bun
```
cjroth: Would it prevent us from having to run these bun setup things again if we put this in ci.yml?

raivieiraadriano92: yes, I just moved the upload job to ci.yml

cjroth (approved):

cjroth:
This is great! I can imagine a scenario where we may want to run it on every commit in the future so that it works with preview builds - but for now, let's merge as is and we can do that if it feels necessary later.

PR #156, opened 2025-09-15
THU-21: Add a weather forecast component

raivieiraadriano92:
## Description

This PR updates the weather forecast flow and completes the first iteration of the Weather Forecast Component described in the ticket.

### Key changes:

- API response format
- - The backend weather API now returns a JSON object instead of a formatted text block.
- get_weather_forecast tool output
- - Updated to return the JSON response directly.
- - This allows us to render the <WeatherForecast /> component immediately after the tool call, without waiting for the model to parse/stream a large text response.
- Prompt refinement for assistant text
- - The assistant prompt was refined so that, in addition to rendering the component, it also generates a short friendly summary with helpful suggestions (no day-by-day breakdown).
- - This avoids showing only a bare component and creates a more natural chat experience.

### Why

- Returning JSON reduces latency and token usage since the model no longer parses verbose forecast text.
- Rendering the component right after the tool call makes the UI feel instant and responsive.
- Adding a concise summary and tips improves user experience by making the forecast feel more conversational.

---

## Render Preview

https://github.com/user-attachments/assets/e06dba54-8a13-4146-b141-bed5c80090a4

---

## Related Issue

Closes [THU-21](https://linear.app/mozilla-thunderbolt/issue/THU-21/add-a-weather-forecast-component)


src/integrations/thunderbolt-pro/tools.ts
```diff
@@ -185,16 +188,50 @@ export const configs: ToolConfig[] = [
...
+- Do NOT include a day-by-day breakdown in text when this tool is used.
+- You may still write a short overview (2–4 sentences) and friendly suggestions (bulleted).`,
+    verb: 'Display 7-day forecast for {location}',
     parameters: weatherSchema,
-    execute: getWeatherForecast,
+    execute: async (params: WeatherParams) => {
+      const forecastData = await getWeatherForecast(params)
+      return forecastData
+
+      // return {
```
cjroth: Should we clean this up?

src/lib/weather-forecast.ts
```diff
@@ -0,0 +1,204 @@
...
+    const date = dayjs(dateString)
+    const hour = date.hour()
+    return hour >= 6 && hour < 18
+  } catch {
+    return true
+  }
+}
+
+export const getWeatherMetadata = (code: number, dateString?: string): WeatherMetadata => {
+  const isDay = dateString ? isDayTime(dateString) : true
```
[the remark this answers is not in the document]
cjroth: Is this a real concern?

cjroth (approved):
Looks good - I think we should merge this.

src/components/chat/weather-forecast.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Skeleton } from '../ui/skeleton'
+
+type WeatherForecastProps = WeatherForecastData
+
+export const WeatherForecast = ({ location, days = [] }: WeatherForecastProps) => {
+  const [temperatureUnit, setTemperatureUnit] = useState<'c' | 'f'>('c')
+
+  const [selectedDayIndex, setSelectedDayIndex] = useState(0)
+
+  const selectedDayMetadata = useMemo(
```
cjroth, line 16: I think we should use useQuery here since it is built for this purpose

src/components/chat/weather-forecast.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useMemo, useState } from 'react'
+import { ToggleGroup, ToggleGroupItem } from '../ui/toggle-group'
+import { cn } from '@/lib/utils'
+import dayjs from 'dayjs'
+import { convertTemperature, getWeatherMetadata, type WeatherForecastData } from '@/lib/weather-forecast'
+import { Skeleton } from '../ui/skeleton'
+
+type WeatherForecastProps = WeatherForecastData
+
+export const WeatherForecast = ({ location, days = [] }: WeatherForecastProps) => {
```
cjroth, line 11: I think we could split this component into two components
- WeatherForecast (current component) which handles fetching and state and is actually rendered in the app
- DisplayWeatherForecast - just the tsx/css rendering only - this way we can display it in storybook without having to make any api calls

cjroth (changes requested):

src/integrations/thunderbolt-pro/tools.ts
raivieiraadriano92: done

src/components/chat/weather-forecast.tsx
raivieiraadriano92: As we talked previously - the api request happens on the tool call.
This way the model has access to make its own reasoning and we can render a custom component in the chat ui

src/components/chat/weather-forecast.tsx
raivieiraadriano92: As we talked previously , the getWeatherMetadata is just a map to get static data based on the weather code.

PR #158, opened 2025-09-23
THU-77: Data reset fails in Tauri Mac

raivieiraadriano92:
### Description
Refactor reset logic in `fs.ts` to branch on **database type** rather than whether the app is running in Tauri.  

Previously, reset assumed `isTauri` always meant `libsql`, which broke once sqlocal started being used inside Tauri. This change makes the logic explicit:
- `libsql-tauri` → use Tauri FS to reset the app data dir.
- `sqlocal` → wipe OPFS entries directly.

### Changes
- Updated `resetAppDir` to check `databaseType` instead of `isTauri`.
- Updated `resetAppDirOpfs` to:
  - Remove all OPFS entries at the root.
  - Only recreate the app-data directory in the browser (skip in Tauri, since
    sqlocal manages the DB file itself).
- Keeps reset consistent across environments and avoids the
  `TypeError: Name is invalid` error in Tauri.

### Result
- Reset now works reliably in both browser and Tauri with sqlocal.
- Libsql reset continues to work via Tauri FS.
- Cleaner separation of logic by backend, not runtime.

### DIFF
--- a/src/lib/fs.ts
+++ b/src/lib/fs.ts
@@ -72,11 +72,16 @@ const resetAppDirOpfs = async (): Promise<void> => {
     throw new Error('OPFS is not available')
   }
 
-  const appDataDirPath = await createAppDirOpfs()
-  const root = await navigator.storage.getDirectory()
+  const appDataDirPath = await createAppDir()
+  const root: any = await navigator.storage.getDirectory()
 
-  await root.removeEntry(appDataDirPath, { recursive: true })
-  await root.getDirectoryHandle(appDataDirPath, { create: true })
+  for await (const [name] of root.entries()) {
+    await root.removeEntry(name, { recursive: true })
+  }
+
+  if (!isTauri()) {
+    await root.getDirectoryHandle(appDataDirPath, { create: true })
+  }
 }
 
 /**
@@ -95,9 +100,11 @@ export const resetAppDir = async (): Promise<void> => {
     return
   }
 
-  if (isTauri()) {
+  if (databaseType === 'libsql-tauri') {
     await resetAppDirTauri()
-  } else {
+  } else if (databaseType === 'sqlocal') {
     await resetAppDirOpfs()
+  } else {
+    throw new Error(`Unsupported database type: ${databaseType}`)
   }
 }


src/lib/fs.ts
```diff
@@ -39,11 +43,22 @@ const createAppDirTauri = async (): Promise<string> => {
  * Creates app data directory in web environment using virtual path for OPFS
  */
 const createAppDirOpfs = async (): Promise<string> => {
-  const virtualPath = 'app-data'
+  const virtualPath = OPFS_APP_DIR_NAME
```
cjroth: I think we should just refer to OPFS_APP_DIR_NAME. Assigning one variable to another here :)

src/lib/fs.ts
```diff
@@ -39,11 +43,22 @@ const createAppDirTauri = async (): Promise<string> => {
...
   console.log('Web environment - using virtual app data path:', virtualPath)
   return virtualPath
 }
 
+/**
+ * Removes every entry from the OPFS root directory to ensure a clean state.
+ */
+const clearOpfsRoot = async (root: OpfsDirectoryHandle): Promise<void> => {
+  const iterator = (root as unknown as { entries: () => DirectoryEntriesIterator }).entries()
```
cjroth: Is there any way to avoid `as unknown as ...` and use the proper types?

src/lib/fs.ts
```diff
@@ -97,7 +112,10 @@ export const resetAppDir = async (): Promise<void> => {
 
   if (isTauri()) {
```
cjroth: This if statement doesn't make sense to me. I think instead of isTauri() we should do more like if (databasetype === 'sqlocal') { a } else { b }

src/settings/preferences.tsx
```diff
@@ -61,6 +62,7 @@ type PreferencesAction =
   | { type: 'SET_LOCATIONS'; payload: LocationData[] }
```
cjroth, line 61: Problem that I just noticed which is not caused by this PR, but we should fix it since we're touching up against it here: I just noticed that these reducer actions are not designed well. I've noticed that AI tends to create a bunch of "setter" actions like this. But this is actually not the point of useReducer. Each action should actually map to an event that happens and it is ok for it to change multiple parts of the state. For example, we want to have a "SEARCH_STARTED", "RESET_STARTED", etc etc.

src/settings/preferences.tsx
```diff
@@ -456,15 +461,25 @@ export default function PreferencesSettingsPage() {
...
-    try {
-      await resetAppDir()
-      trackEvent('settings_database_reset')
-      // Refresh the page to reinitialize the app
-      window.location.reload()
-    } catch (error) {
-      console.error('Failed to reset database:', error)
-      dispatch({ type: 'SET_IS_RESETTING', payload: false })
-    }
+    /**
```
cjroth: I'd rather see a harmless console error about `DatabaseSingleton not initialize` than complicate the code with setTimeouts tbh. I've noticed AI loves to try to "fix" race condition bugs this way but IMO it just makes the code quality worse and introduces more complexity and makes it harder to create unit tests. If there's a way to fix this properly, I'd be interested, but I don't think the error causes any problem so maybe we can just leave it alone for now.

src/lib/fs.ts
```diff
@@ -95,9 +99,9 @@ export const resetAppDir = async (): Promise<void> => {
...
   }
 
-  if (isTauri()) {
-    await resetAppDirTauri()
-  } else {
-    await resetAppDirOpfs()
+  if (databaseType === 'libsql-tauri') {
+    await resetLibsqlWithTauriFs()
+  } else if (databaseType === 'sqlocal') {
+    await resetSqlocalWithOpfs()
```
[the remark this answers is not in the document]
cjroth: I kind of agree with this one - I think we should add an else that throws an error to let us know that the db type isn't supported. In the event that we add another db type in the future and forget to update this, it would just silently fail here.

src/lib/fs.ts
```diff
@@ -67,16 +67,20 @@ const resetAppDirTauri = async (): Promise<void> => {
 /**
  * Resets app data directory in web environment using OPFS
  */
-const resetAppDirOpfs = async (): Promise<void> => {
+const resetSqlocalWithOpfs = async (): Promise<void> => {
```
cjroth: I don't think we need to change this function at all do we? I think the bug would be solved entirely if we just fixed the branching logic where this gets called. Maybe I'm misunderstanding the bug?

cjroth (changes requested):

PR #169, opened 2025-10-04
THU-90: Refactor chat creation flow with /chat/new route

raivieiraadriano92:
## Description
Previously, when a user clicked “New Chat”, the app would immediately create a new chat record in the database, retrieve its id, and then navigate directly to /chats/:id.

## New Approach
With this update, we refactored the flow to use a dedicated /chats/new route, improving both UX and architecture.
Now, when the user opens a new chat:

- They are first taken to /chats/new (a temporary in-memory state).
- Once they start typing or send the first message, the chat record is created in the database.
- The app then navigates to /chats/:id using useNavigate with the relative option to prevent reloads.

The `relative: 'path'` option tells React Router to resolve the new path relative to the current route instead of reloading the full page. This ensures state preservation, faster navigation, and no redundant re-renders between /chats/new and /chats/:id.

## Preview

https://github.com/user-attachments/assets/652130c0-0c41-4c08-8e1f-ea9e7619712b

---

> [!NOTE]
> Refactors chat creation to route to /chats/new and create the thread on first message, updating DAL APIs, routing, and tests accordingly.
> 
> - **Frontend/Routing**:
>   - Default index now redirects to `/chats/new` (removes dependency on `initData.initialThreadId`).
>   - `ChatDetailPage` handles `chats/new` by generating a UUID client-side and navigating to `chats/:id` on first message; updates query keys and navigation.
>   - `ChatUI` "New Chat" navigates to `/chats/new` (no pre-create).
>   - Sidebar "New Chat" and "Clear all chats" navigate to `/chats/new` and stop pre-creating threads.
>   - Minor: adjust `AppRoutes` param usage.
> - **DAL**:
>   - Add `createChatThread(id)` and `getChatThread(id)`; replace `getChatThreadById`.
>   - Change `getOrCreateChatThread(id)` to get-or-create by provided ID; remove internal UUID generation and empty-thread reuse logic.
> - **Tests**:
>   - Rewrite chat thread tests to cover `createChatThread`, `getChatThread`, and new `getOrCreateChatThread(id)` semantics; remove old empty-thread reuse tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 42b63a049cfec84f991f32882e96a046a51d240d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/lib/dal.ts
+++ b/src/lib/dal.ts
@@ -1,5 +1,4 @@
-import { and, asc, desc, eq, like, notExists, sql } from 'drizzle-orm'
-import { v7 as uuidv7 } from 'uuid'
+import { and, asc, desc, eq, like, sql } from 'drizzle-orm'
 import { DatabaseSingleton } from '../db/singleton'
 import {
   accountsTable,
--- a/src/lib/dal.ts
+++ b/src/lib/dal.ts
@@ -278,6 +278,23 @@ export const getChatThreadById = async (id: string) => {
 }
 
 /**
+ * Gets a specific chat thread by ID or create a new one with the provided ID
+ */
+export const getOrCreateChatThreadById = async (id: string) => {
+  const db = DatabaseSingleton.instance.db
+
+  const thread = await getChatThreadById(id)
+
+  if (thread?.id) {
+    return thread
+  }
+
+  await db.insert(chatThreadsTable).values({ id, title: 'New Chat' })
+
+  return await getChatThreadById(id)
+}
+
+/**
  * Gets an existing empty chat thread or creates a new one
  */
 export const getOrCreateChatThread = async (isEncrypted: boolean = false): Promise<string> => {
--- a/src/lib/dal.ts
+++ b/src/lib/dal.ts
@@ -272,61 +272,32 @@ export const getAllChatThreads = async () => {
 /**
  * Gets a specific chat thread by ID
  */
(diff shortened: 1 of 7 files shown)


src/lib/dal.ts
```diff
@@ -277,6 +277,23 @@ export const getChatThreadById = async (id: string) => {
   return await db.select().from(chatThreadsTable).where(eq(chatThreadsTable.id, id)).get()
 }
 
+/**
```
cjroth, line 279: Minor code style thing (just my preference here):
- I'd break this into `createChatThread` and `getChatThread` functions
  - I think `byId` is unnecessarily verbose - generally getters are by id, so it is kind of assumed
  - I think splitting it this way makes the logic more reusable, testable, and atomic
- I'd have a getOrCreateChatThread function that does the if/then routing between those two functions
- I'd add unit tests in dal.test.ts

Just want to make sure we're setting good patterns for everything as the codebase grows :)

cjroth (approved):
The seems to work great! I'm very happy to see this PR.

One minor thing also is I don't think we should add the flower submodule - I think we removed it from main already?

src/lib/dal.ts
raivieiraadriano92: done

PR #129, opened 2025-08-27
THU-40: Implement granular PostHog event tracking

raivieiraadriano92:
## Description

This PR introduces comprehensive PostHog analytics integration with a structured event tracking system. The implementation includes:

### Key Features
- **Typed Event System**: All events use a typed `EventType` union for compile-time safety and IDE autocomplete
- **Structured Naming Convention**: Events follow the `<feature>_<action>` pattern (e.g., `chat_send_prompt`, `task_add`)
- **Comprehensive Coverage**: Tracks 25+ user interactions across all major features
- **Privacy-First**: Respects user data collection preferences and includes no PII

### Event Categories Implemented
- **Chat & Messaging** (`chat_*`): Message sending, replies, chat management
- **Model Management** (`model_*`): Model selection and configuration
- **Settings** (`settings_*`): Theme, name, location, and database management
- **Task Management** (`task_*`): Task CRUD operations and search
- **Automation** (`automation_*`): Automation creation, editing, and execution
- **UI & Navigation** (`ui_*`): Sidebar interactions and keyboard shortcuts

### Technical Implementation
- Centralized `trackEvent()` function with error handling
- Type-safe event names preventing typos and runtime errors
- Consistent property structure for analytics insights
- Integration with existing user preference system

### Documentation
- Added comprehensive Event Tracking section to README.md
- Includes implementation examples and best practices
- Documents privacy controls and data collection policies

This analytics system will provide valuable insights into user behavior and feature usage while maintaining user privacy and code quality standards.

---

## Render Preview

image
image
image
image
image
image

---

## Related Issue

Closes [THU-40](https://linear.app/mozilla-thunderbolt/issue/THU-40/implement-granular-posthog-event-tracking)

PR #174, opened 2025-10-08
THU-55: Create Link Preview Component

raivieiraadriano92:
> [!NOTE]
> Adds a link preview UI and tool with prompt guidance, extends fetch_content to return summaries, and refactors cloud URL access.
> 
> - **UI**:
>   - **Link Preview**: New `LinkPreview` + `LinkPreviewContainer` with skeleton; renders via `display-tool-handler` on `display-link_preview`.
>   - **Storybook**: Adds `link-preview.stories.tsx`.
>   - **Icons**: `tool-icon` now uses `useCloudUrl` hook for proxied favicons.
> - **Hooks**:
>   - New `use-cloud-url` for retrieving cloud URL.
> - **Tools (Thunderbolt Pro)**:
>   - New `linkPreviewSchema` and `configs` entry `display-link_preview` (render-only tool).
>   - `search` tool description updated with heuristic to render 1–3 previews.
>   - `FetchContentData` now exported and includes `summary`.
> - **Backend**:
>   - Exa `fetch-content` now requests `{ text: true, summary: true }`.
> - **Prompt**:
>   - Style: “Markdown only (no HTML/XML)”.
>   - Adds display tools policy and usage heuristics for link previews.
> - **Tests**:
>   - Update `fetchContent` tests to expect `summary`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d0cc634d393bb8f51799ee93bcb6c2dbe0f9b9e8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Preview

https://github.com/user-attachments/assets/b83a1f3d-dd86-4c5c-8ef5-2e189500ddea


src/ai/prompt.ts
```diff
@@ -49,6 +49,7 @@ export const createPrompt = ({ preferredName, location }: PromptParams) => {
     // —— Style guide ——
     `Respond in Markdown (no XML) that is pleasant, concise, and helpful. Use subheaders, bullet points, and bold / italics to help structure the response. Use emojis where appropriate.`,
     `Never invent information unless the user explicitly requests creative fiction.`,
+    `You MUST call the display-link_preview tool after completing all search or deep search operations, once you have identified one or more relevant URLs. Use it to render link previews for the most relevant webpages you’ve found.`,
```
cjroth: Atm it is calling this display tool even when generating the daily brief. I think we need to tweak this to clarify that this should only be used when the goal is to show the link to the user as the final output and not if the final output is something else, like a document or report. I'm not sure exactly what the best wording is.

src/components/chat/link-preview.tsx
```diff
@@ -0,0 +1,105 @@
...
+const useFetchLinkPreviewContent = (url: string) => {
+  const [linkPreview, setLinkPreview] = useState<LinkPreviewProps>()
+  const [isLoading, setIsLoading] = useState(true)
+
+  useEffect(() => {
+    if (url) {
+      fetchContent({ url })
+        .then((content) =>
+          setLinkPreview({
+            description: markdownToText(content?.text ?? ''),
```
cjroth: Would it make sense to use the `summary` property returned by Exa here instead of `text`?

src/hooks/use-cloud-url.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useState } from 'react'
+import { getCloudUrl } from '@/lib/config'
+
+export const useCloudUrl = () => {
```
cjroth, line 4: Nice, very clean

src/lib/utils.ts
```diff
@@ -213,3 +213,40 @@ export const splitPartType = (type: string): [string, string] => {
   }
   return [type.slice(0, dashIndex), type.slice(dashIndex + 1)]
 }
+
+export function markdownToText(markdown: string): string {
```
cjroth: This one should have some unit tests to verify that it works as expected - but I would prefer if we could avoid using it at all (see above comment)

cjroth (changes requested):
Code quality is good here. I think we still need to test it more - I'm seeing some bugs:
- It is showing me link previews in the daily brief tool calls, which it shouldn't.
- It is getting a CORS error on some preview images. Some preview images don't load (the browser's error placeholder image is showing up) and other times the preview image is just not shown at all)
- Most important: if I give the prompt "what are the top 3 robot vacuums", it provides a single preview of whatever website it went to to find 3 robot vacuums instead of actually linking me to those 3 robot vacuums. I think we might need to play around with the prompt here or think about how to achieve that better.

src/lib/utils.ts
raivieiraadriano92: So, at first I  wanted to avoid increasing the response payload with an extra text field - this because the preview is showing only 2 lines. But I agree that this method seems a little weird and not 100% reliable.

cjroth:
@raivieiraadriano92 I merged your changes into #129 so I'm going to close this one. It was realllly hard to get it to use the link preview correctly, so it made sense to merge this one with the prompt engineering stuff I'm working on in that PR. Nice work here!

PR #175, opened 2025-10-08
THU-93: Add localization settings

ital0:
### **Summary**
Adds localization settings to allow users to configure their preferred units (temperature, wind speed, precipitation, time format, distance) with automatic detection based on OS locale.

### **Key Changes**
- **Backend**: New `/units` API endpoint serving comprehensive units data
- **Frontend**: New "Localization" section in preferences with 5 unit dropdowns
- **Auto-detection**: Detects metric/imperial system based on user's locale
- **UX**: Loading states, auto-closing dropdowns, no UI blinking
- **Testing**: Unit tests for all new functionality

### **Files Added**
- `backend/src/data/units.json` - Units data
- `src/lib/unit-detection.ts` - Locale detection logic
- `src/hooks/use-units.ts` - Data fetching hook
- `src/hooks/use-localization-form.ts` - Form logic hook
- `src/hooks/use-localization-dropdowns.ts` - Dropdown state hook
- `src/test-utils/` - Shared test utilities

### **Files Modified**
- `backend/src/api/routes.ts` - Added `/units` endpoint
- `src/settings/preferences.tsx` - New localization UI
- `src/types.ts` - New type definitions
- `src/lib/dal.ts` - Updated defaults
- `src/lib/analytics.tsx` - Added localization tracking

### **Technical Details**
- Uses React Query for caching (24h)
- Centralized default units constants
- Removed legacy enum types
- Imperial fallback for US/Liberia/Myanmar, metric for others
- Custom hooks for better separation of concerns

---

> [!NOTE]
> Adds localization settings (distance, temperature, date/time, currency) with backend units endpoints/data, integrates preferences into weather/tools/AI, and updates analytics and DAL with broad test coverage.
> 
> - **Backend**:
>   - **New Endpoints**: `GET /v1/units?country=...` and `GET /v1/units-options` in `backend/src/api/routes.ts`.
>   - **Localization Data/Utils**: Add `units-by-country.json`, `units-options.json`, `country-mapping.json` and `resolveCountryCode` utility; comprehensive tests.
> - **Pro Weather/Tools**:
>   - Pass user `distanceUnit`/`temperatureUnit` to Open‑Meteo; map to API params; update `/pro/weather/*` and `/pro/locations/search` request/validation.
>   - Remove `SimpleContext` usage from routes/tests.
> - **Frontend**:
>   - Preferences: new **Localization** section (5 dropdowns) with hooks (`use-units-options`, `use-country-units`, `use-localization-form`, `use-localization-dropdowns`) and cached fetching.
>   - Auto-apply country defaults on location change; new schemas (`src/schemas/api.ts`) and types (`PreferencesSettings`, currency/date types).
>   - AI: `createPrompt` and `ai/fetch` include localization context.
> - **DAL & Analytics**:
>   - Add bulk `getSettings`/`setSettings`; refactor consumers to use them.
>   - Track `settings_localization_update`; update TELEMETRY docs.
> - **Tests/Deps**:
>   - Extensive unit tests across backend/frontend; add Testing Library deps and query test utils.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d4736a25eb512cc96eff2b670958a10c9c923a6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/lib/dal.ts
```diff
@@ -139,6 +140,11 @@ export const getPreferencesSettings = async () => {
   const preferredName = await getSetting('preferred_name', '')
   const dataCollection = await getBooleanSetting('data_collection', true)
   const experimentalFeatureTasks = await getBooleanSetting('experimental_feature_tasks', false)
+  const temperatureUnit = await getSetting('temperature_unit', DEFAULT_IMPERIAL_UNITS.temperature)
```
cjroth: A minor nice-to-have refactor we could do here is to add a setSettings and getSettings (plural) that lets use fetch and set multiple settings in a single query.

cjroth (changes requested):
* Code quality here is very high - this is really good
* Fast turnaround - I’m really impressed with how fast you got this PR submitted 
* I think we should still go with the original idea of auto-filling the unit preferences after the user selects a location
    * I think we should use this: https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/supplemental/unitPreferenceData.json and possibly other json files from this repo that are useful
    * We can check that exact json object into the backend and basically have an endpoint that is like /units?location=<location> which returns something along the lines of
        * { distance: “imperial”, temperature: “c”, weight: “kg” } (vaguely - feel free to tweak specifics as needed
    * I’m picturing this working as a useQuery that is triggered when the location changes, basically
* I think current units are a bit too specific to weather, but these will actually be used throughout the app in all scenarios, so instead of “wind speed”, we could have “distance” for example which will achieve the same outcome but in a more general way. I think we should go with
    * Distance - I think “m” vs “km” is more about the specific scenario than the location (same with yard vs ft) so I think we should maybe just change this to “metric” and “imperial”. Maybe also change name from “distance” to “units” to be more general?
    * Temperature - I don’t think we need K
    * Precipitation - I think this is too specific to weather - we could just go based on whether we choose imperial or metric for “units” / distance above
    * Time Format - let’s simplify to just 12 vs 24 time
    * Date Format - might be nice to have this
    * Currency - might be nice to have
* We’ll want to update ai/prompt.ts to add the these localization preferences to the system prompt so that it knows the user’s preferences
* We should update the display weather component to default to C vs F based on the new setting value
* We should figure out how to make the getWeatherForecast api endpoint/tool to respond in the correct units (maybe a param we pass to it?)

ital0:
image

src/settings/preferences.tsx
```diff
@@ -596,6 +622,279 @@ export default function PreferencesSettingsPage() {
...
+        <Form {...localizationForm}>
+          <form className="flex flex-col gap-4" onSubmit={(e) => e.preventDefault()}>
+            <FormField
+              control={localizationForm.control}
+              name="distanceUnit"
+              render={({ field }) => (
+                <FormItem className="flex flex-row items-center gap-4">
+                  <div className="flex-1">
+                    <FormLabel>Distance</FormLabel>
+                    <FormDescription>Choose your preferred distance unit.</FormDescription>
```
cjroth: I think we can remove these "Choose..." descriptions actually and just have the labels to make the UI a little bit cleaner

src/settings/preferences.tsx
```diff
@@ -596,6 +622,279 @@ export default function PreferencesSettingsPage() {
...
+                          <ChevronsUpDown className="ml-2 h-4 w-4 shrink-0 opacity-50" />
+                        </Button>
+                      </FormControl>
+                    </PopoverTrigger>
+                    <PopoverContent className="p-0 w-auto">
+                      <Command>
+                        <CommandList>
+                          <CommandGroup>
+                            {unitsOptionsData?.timeFormat?.map((format) => (
+                              <CommandItem
```
cjroth, line 825: The initial state here should show `12h` or `24h` but the `h` is missing

Image

src/settings/preferences.tsx
```diff
@@ -596,6 +622,279 @@ export default function PreferencesSettingsPage() {
...
+                          role="combobox"
+                          disabled={unitsOptionsLoading}
+                          className={cn('w-auto justify-between', !field.value && 'text-muted-foreground')}
+                        >
+                          {unitsOptionsLoading ? 'Loading...' : field.value || 'Select...'}
+                          <ChevronsUpDown className="ml-2 h-4 w-4 shrink-0 opacity-50" />
+                        </Button>
+                      </FormControl>
+                    </PopoverTrigger>
+                    <PopoverContent className="p-0 w-auto">
```
cjroth, line 771: Can we show just the example format here like `2025-12-01` instead of `YYYY-MM-DD (2025-12-01)`?

Image

src/lib/dal.ts
```diff
@@ -133,20 +134,35 @@ export const getAllSettings = async () => {
  * Gets preferences settings with specific structure
  */
 export const getPreferencesSettings = async () => {
-  const locationName = await getSetting('location_name', '')
-  const locationLat = await getSetting('location_lat', '')
-  const locationLng = await getSetting('location_lng', '')
-  const preferredName = await getSetting('preferred_name', '')
+  // Get all string settings in a single query
```
cjroth: What if we could update the `getSettings` function to accept an optional object of default values as the 2nd param and then use getSettings directly in `preferences.tsx` so that we no longer need this function?

src/lib/dal.ts
```diff
@@ -218,6 +234,51 @@ export const updateSetting = async (key: string, value: string | null): Promise<
   })
 }
 
+/**
+ * Get multiple settings in a single query
+ * @param keys - Array of setting keys to fetch
+ * @returns Object with key-value pairs for the requested settings
+ */
+export const getSettings = async (keys: string[]): Promise<Record<string, string | null>> => {
```
cjroth: Nice, much better than having to call each one individually

backend/src/data/localization/units-by-country.json
```diff
@@ -0,0 +1,2832 @@
...
+      }
+    },
+    "ZZ": {
+      "units": "metric",
+      "temperature": "C",
+      "timeFormat": "24",
+      "dateFormatExample": "DD/MM/YYYY",
+      "currency": null
+    }
+  }
```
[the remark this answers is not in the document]
ital0: on it

cjroth (changes requested):
This is a really good first PR - your code is super clean and well organized. I appreciate the quality a lot.

src/hooks/use-country-units.ts
```diff
@@ -0,0 +1,37 @@
...
+      const response = await ky
+        .get(`${cloudUrl}/units`, {
+          searchParams: { country: countryName },
+        })
+        .json()
+
+      const validatedData = countryUnitsResponseSchema.parse(response)
+      return validatedData
+    },
+    enabled: false,
```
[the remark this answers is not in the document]
ital0: That was intentional since we don't want to keep fetching country units unless user changes their Location settings.

backend/src/data/localization/units-by-country.json
```diff
@@ -0,0 +1,2840 @@
+{
+    "AC": {
+      "units": "metric",
```
raivieiraadriano92: Just a minor observation - shouldn't we use unit in the singular form instead of plural?

src/schemas/api.ts
```diff
@@ -0,0 +1,40 @@
...
+export const unitsOptionsResponseSchema = z.object({
+  units: z.array(z.string()),
+  temperature: z.array(temperatureUnitSchema),
+  timeFormat: z.array(z.string()),
+  dateFormats: z.array(dateFormatSchema),
+  currencies: z.array(currencySchema),
+})
+
+export const countryUnitsResponseSchema = z.object({
+  units: z.string(),
```
raivieiraadriano92: Same thing here - maybe we should use singular instead of plural

raivieiraadriano92 (approved):
LGTM

src/lib/dal.ts
```diff
@@ -218,6 +197,56 @@ export const updateSetting = async (key: string, value: string | null): Promise<
   })
 }
 
+/**
+ * Get multiple settings in a single query with auto-detection of types
+ * @param config - Object with setting keys and their default values
+ * @returns Object with key-value pairs for the requested settings, properly typed
+ */
+export const getSettings = async <T extends Record<string, string | number | boolean>>(config: T): Promise<T> => {
```
cjroth: I'm going to use this one in my PR once this one is merged. Btw, I think maybe this is more complex than needed - what if we kept it consistent with getSetting (singular) where it just does a single SQL query and does not try to auto-detect the type in this way? I think it's simpler to just assume that if you're going to use this function, you should plan on knowing how to parse each value.

cjroth (changes requested):

src/lib/dal.ts
ital0: @cjroth that makes sense. Actually, the intention behind trying to detect the type was precisely to avoid any kind of error in case the developer forgets or doesn’t know how to handle it.
But I’m totally in favor of keeping things simple.

ital0: @cjroth ok, what I did here was create a default handling for all data types except boolean.
I don’t think it’s a good pattern to compare booleans as if they were strings (like x === 'true').

src/ai/prompt.ts
```diff
@@ -33,6 +40,14 @@ export const createPrompt = ({ preferredName, location }: PromptParams) => {
       ? `You must use units that are appropriate for the user's country based on the task at hand. If tools give you results in the wrong units, you must convert them. For example, if the user's location is in the United States, use miles and Fahrenheit and miles per hour. If the user's location is in Canada, use kilometers, Celsius and kilometers per hour.`
```
cjroth, line 40: I think this instruction and the one you added are sort of contradicting each other - we should probably remove this tbh.

src/integrations/thunderbolt-pro/tools.ts
```diff
@@ -1,10 +1,31 @@
...
+  try {
+    const settings = await getSettings({
+      temperature_unit: 'F',
+      distance_unit: 'imperial',
+    })
+    return {
+      distanceUnit: settings.distance_unit,
+      temperatureUnit: settings.temperature_unit,
+    }
+  } catch (error) {
```
cjroth, line 21: I'd rather have the error get thrown so we can fix it than to have it get quietly redirected to a console.warn here, since I wouldn't expect this to fail in normal circumstances

src/lib/dal.ts
cjroth: Tbh I think this setting stuff will change in my PR, so I'm not that concerned either way

backend/src/pro/weather.ts
```diff
@@ -17,6 +34,72 @@ export class OpenMeteoWeather {
...
 
+  /**
+   * Convert user preferences to Open-Meteo API parameters
+   */
+  private getUserLocalizationParams(userPreferences?: WeatherPreferences): LocalizationParams {
+    const params: LocalizationParams = {}
+    
+    if (userPreferences?.temperatureUnit) {
+      switch (userPreferences.temperatureUnit.toUpperCase()) {
+        case 'F':
```
cjroth, line 45: I think we should make the possible values of temperatureUnit just "C" or "F" - I can't picture how these different variations would get into the settings table, and if they can, then that seems like an inconsistency that we need to fix when updating the setting

src/settings/preferences.tsx
```diff
@@ -118,6 +125,20 @@ export default function PreferencesSettingsPage() {
   const [state, dispatch] = useReducer(preferencesReducer, initialState)
   const { open, searchQuery, locations, isSearching, isResetting } = state
 
+  // Localization dropdown states
```
cjroth, line 128: Thinking out loud here. I think the state management of this preferences screen has gotten a bit inconsistent - it was already getting messy before your PR, so not your fault, but it's becoming more apparent to me. I think we've got 3 different patterns going on here: a hook defined in an outside file, some that use useState, and others that are using useForm.

I think my PR will improve this somewhat. But just saying it out loud that we should think through this a bit in the future

cjroth (changes requested):
When I check the weather display tool in the chat, the celcius toggle is selected even though I have F set in the settings and the numbers appear to be in F. Then when I click F, it is trying to convert C to F despite the numbers already being F, resulting in very high numbers.

Image

Image

src/settings/preferences.tsx
cjroth: (no change needed here)

cjroth:
Merged this into #191

PR #177, opened 2025-10-08
refactor: enhance loading logic in App component with framer-motion

raivieiraadriano92:


cjroth:
Just commenting for visibility - we've identified the most likely cause of the flickering as the auto-resizing text area. Removing the auto-resizing seems to resolve it. We will investigate ways to fix the flicker based on this finding. We will remove the animation.

PR #193, opened 2025-10-16
THU-103 + THU-126: Qwen3 -> gpt-oss

cjroth:
> [!NOTE]
> Removes Flower across backend/frontend, makes gpt-oss-120b the default/supported model with reset-to-default and improved selection logic, migrates backend deps/commands to Bun, and refactors tests/utilities.
> 
> - **Models**:
>   - Default system model changed from `qwen/qwen3-235b` to `gpt-oss-120b`; `supportedModels` includes `gpt-oss-120b`.
>   - DAL: ordered/filtered model queries, fallback to system when selected is disabled, `resetModelToDefault` added.
>   - UI: model list shows modification indicator and supports reset; sidebar swaps Integrations/Models entries.
> - **Flower Removal**:
>   - Backend: delete `flower` routes/auth, env keys, and tests; PostHog/cluster/instrumentation tidied.
>   - Frontend: remove `src/flower/*`, Flower middleware/tooling, fetch override, and related settings.
>   - Capabilities/Settings: drop Flower URLs and `disable_flower_encryption`.
> - **AI pipeline**:
>   - Replace Flower-specific middleware with `extractReasoningMiddleware` in fetch/streaming/tests; remove Flower toolset.
> - **Dev/Build**:
>   - Backend now uses Bun (install/test/format/dev); Makefile and README updated; Python/uv tasks removed.
> - **Testing**:
>   - Add DB `setup/reset/teardown` utils; refactor DAL tests to use them.
> - **Misc**:
>   - Minor OpenAI route/tests updates; data and formatting cleanups; capability and lockfile pruning.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 52043c43b0aa41fd741b0b09b6b93c4e325c5d49. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/settings/models/index.tsx
```diff
@@ -1133,7 +1153,15 @@ export default function ModelsPage() {
                             </Tooltip>
                           </TooltipProvider>
                         )}
-                        {model.name}
+                        <ModificationIndicator
```
raivieiraadriano92, line 1155: just a thought - what about use a badge with a label besides the name instead of just a line below the name?

src/dal/models.ts
```diff
@@ -69,7 +77,8 @@ export const getSelectedModel = async (): Promise<Model> => {
   if (selectedModelId) {
     const model = await getModel(selectedModelId)
 
-    if (model?.id) {
+    // Check if model exists and is enabled
+    if (model?.id && model.enabled) {
```
raivieiraadriano92, line 81: I see we are validating if the model is enabled always after calling getModel - would be a good idea move this to the `where` condition in `getModel`? Or even create a new method like `getEnabledModel` or something like that?

raivieiraadriano92 (approved):
Just added 2 minor thoughts/suggestions - other than that it LGTM

src/settings/models/index.tsx
cjroth: Will follow up with this in Matrix :)

src/dal/models.ts
cjroth: I think we should consider this, but for now it's only 2 places so I'm going to leave that for another PR

PR #196, opened 2025-10-17
THU-97: Onboarding Experience

ital0:
Multi-step wizard to guide the user about privacy and provide a brief overview of the application.

For the implementation of the components, I used an approach based on a main wrapper that contains the various steps, supporting plug-and-play addition of new steps or easy reordering.

I encapsulated the provider integration logic within a hook, allowing easy coupling of new providers if necessary.

For the unit tests, since it was necessary to use the testing library to test React components, new development dependencies were added so that Bun could support the executed tests. To achieve this, happydom was added [according to Bun’s own documentation](https://bun.com/guides/test/testing-library).

---

> [!NOTE]
> Adds a 5-step onboarding wizard (privacy → OAuth → name → location → celebration), introduces a reusable OAuth connect hook/button, integrates onboarding into the app, refactors Preferences/Integrations to shared hooks, and sets up a testing environment.
> 
> - **Onboarding**:
>   - Add `OnboardingDialog` with steps: privacy (`onboarding-privacy-step`), auth (`onboarding-auth-step`), name, location, and celebration, plus `StepIndicators` and shared UI (`IconCircle`).
>   - New state management via `use-onboarding-state`; wired into `App` and gated by `user_has_completed_onboarding` default setting.
> - **OAuth & Integrations**:
>   - Introduce `use-oauth-connect` hook and `ConnectProviderButton` (Google/Microsoft), with session-based `returnContext` and preferred-name saving.
>   - Update `oauth-callback` to route by context; refactor `settings/integrations` to use the hook and new button; add provider logo components.
> - **Preferences & Localization**:
>   - Add `use-country-units`, `use-location-search`, and `extractCountryFromLocation`; update `settings/preferences` to use them and prompt to apply country units.
> - **UI**:
>   - Add `AppLogo` and replace icon in `sidebar-header`.
> - **Testing & Stories**:
>   - Configure Bun test env (`bunfig.toml`, `happydom.ts`, `testing-library.ts`, `matchers.d.ts`).
>   - Add extensive unit tests for onboarding, hooks, tools, and webview; add Storybook stories for onboarding components.
> - **Misc**:
>   - Minor snapshot header casing change; small auth postMessage guard; tools/tests adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 198e7132a16a45720d01c8fccb63e88162c4ed44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/hooks/use-country-units.ts
```diff
@@ -1,40 +1,68 @@
-import { useQuery } from '@tanstack/react-query'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
 import ky from 'ky'
 import type { CountryUnitsData } from '@/types'
 import { countryUnitsResponseSchema } from '@/schemas/api'
 import { getCloudUrl } from '@/lib/config'
+import { extractCountryFromLocation } from '@/lib/country-utils'
 import { useSettings } from './use-settings'
 
+const STALE_TIME = 24 * 60 * 60 * 1000 // 24 hours
```
cjroth: We're using regular camel case convention even for consts like this throughout the codebase - let's keep it consistent with those

src/hooks/use-location-search.ts
```diff
@@ -0,0 +1,120 @@
...
+      return { ...state, searchQuery: '', locations: [] }
+    default:
+      return state
+  }
+}
+
+/**
+ * Custom hook for location search functionality
+ */
+export const useLocationSearch = () => {
```
cjroth, line 56: I feel like we might be able to re-use some of the logic from the preferences page here - or maybe rewrite the location search state logic in a way that enables that?

cjroth (changes requested):
Here's my feedback:

## Functionality
- We should set the localization settings based on the location the user sets (but don't show the modal like we do in settings - just do this silently)
- Finish the connect Google integration
- Location/name inputs should automatically focus when you get to that screen so that the user doesn't have to click into the input
- Location step should be skippable
- I think we should add one final screen after the location screen that is sort of like a celebration/welcome graphic letting the user know they are all set up and everything is ready to go
- I think this should use a regular modal, not the responsive one (that one is more of a card that slides up from the bottom which I don't think make sense here)
- On mobile, I think we should make this a full-screen overlay covering the page instead of showing it like a dialog in order to utilize the small screen space

## Design
- I think the background blur could be reduced slightly so the user can see get a little bit more of a teaser of the app UI
- Instead of 3 vertically stacked buttons in most screens, I would like to see a footer with a left-aligned back button, a right-aligned skip button ([ghost](https://ui.shadcn.com/docs/components/button#ghost)), and then a right-aligned continue button (primary)
- I'd kind of like each step to be the same width and height (eg we could use a flex box and have content vertically centered) so that the modal doesn't change size every time the user clicks next
- On the screen for connecting Google, I think we should find a way to show the google logo
- In general, try to spend some time polishing the UI a bit and making it feel really clean, uniform, and ready for production

## Code
- Can we name the components based on their functionality instead of numerically so that it's easier to re-order them and find the files quickly?
- If possible, I'd like to see storybook stories for each screen to make it easy to review/test/preview this
- We should add unit tests for the hooks/state logic

src/hooks/use-onboarding-navigation.ts
```diff
@@ -0,0 +1,52 @@
+import { useState } from 'react'
+
+const TOTAL_STEPS = 4
```
raivieiraadriano92: same as @cjroth commented above - let's keep te regular camel case convention

src/hooks/use-units-options.test.ts
```diff
@@ -1,4 +1,19 @@
-import { describe, expect, it } from 'bun:test'
+import { renderHook, waitFor } from '@testing-library/react'
+import { vi, describe, it, beforeEach, afterEach, expect } from 'vitest'
+import '@testing-library/jest-dom'
+import { useUnitsOptions } from './use-units-options'
+import { createQueryTestWrapper } from '@/test-utils/react-query'
+import { setupTestDatabase, resetTestDatabase } from '@/dal/test-utils'
+import { updateSetting } from '@/dal/settings'
+
+// Mock only ky, use real database for getSettings
```
cjroth, line 9: I wouldn't mind using an injection pattern from ky like we do fetch in other places. I notice a lot of libraries have started to make fetch injectable. Separately, I've been on the fence about whether I like ky or prefer just using fetch. No strong feelings about this, just thinking out loud

src/hooks/use-units-options.test.ts
```diff
@@ -1,4 +1,19 @@
-import { describe, expect, it } from 'bun:test'
+import { renderHook, waitFor } from '@testing-library/react'
+import { vi, describe, it, beforeEach, afterEach, expect } from 'vitest'
```
cjroth: Should use `bun:test` here instead of vitest

src/hooks/use-onboarding-state.ts
```diff
@@ -0,0 +1,326 @@
+import { useReducer, useEffect } from 'react'
```
cjroth, line 1: I think this file is missing unit tests

src/lib/auth.ts
```diff
@@ -101,16 +101,33 @@ export const startOAuthFlow = async (
...
+    const features = 'width=600,height=700,scrollbars=yes,resizable=yes'
+    popup = window.open(authUrl, 'thunderbolt-oauth', features)
+
+    // Fallback attempt without explicit features in case some environments block feature strings
+    if (!popup) popup = window.open(authUrl, '_blank')
+
+    if (!popup) throw new Error('Failed to open authentication window. Please allow pop-ups and try again.')
+    try {
+      popup.focus()
+    } catch {
```
cjroth: Was this actually causing an error? Looks overly defensive to me. I think we should at least console.warn here - some future person debugging this might get burned

src/lib/auth.ts
```diff
@@ -101,16 +101,33 @@ export const startOAuthFlow = async (
...
     const handler = (event: MessageEvent) => {
-      if (event.data.type === 'oauth-callback') {
+      if (event.data?.type === 'oauth-callback') {
         window.removeEventListener('message', handler)
-        if (popup && !popup.closed) popup.close()
+        if (popup && !popup.closed) {
+          try {
+            popup.close()
+          } catch {
+            // Ignored: some browsers disallow programmatic close
```
cjroth: Same question here. Need to at least console.warn here as well

src/settings/preferences.tsx
```diff
@@ -105,9 +69,10 @@ const preferencesReducer = (state: PreferencesState, action: PreferencesAction):
 
 export default function PreferencesSettingsPage() {
   const [state, dispatch] = useReducer(preferencesReducer, initialState)
-  const { open, searchQuery, locations, isSearching, isResetting, localizationDialogOpen, pendingCountryUnits } = state
+  const { isResetting, localizationDialogOpen, pendingCountryUnits } = state
+  const locationSearch = useLocationSearch()
```
cjroth, line 73: I love this reusable code here. Nice

src/stories/onboarding/wrappers/OnboardingDialogWrapper.tsx
```diff
@@ -0,0 +1,28 @@
+import { Dialog, DialogContent } from '@/components/ui/dialog'
+import { OnboardingPrivacyStepWrapper } from './OnboardingPrivacyStepWrapper'
+import { StepIndicators } from '@/components/onboarding/step-indicators'
+
+const TOTAL_STEPS = 5
```
cjroth: Nitpick but most other places in the codebase use camel case even for constants like this

src/stories/onboarding/wrappers/OnboardingDialogWrapper.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const OnboardingDialogWrapper = () => {
+  return (
+    <Dialog open={true} onOpenChange={() => {}}>
```
cjroth: Hmm can onOpenChange not just be skipped entirely?

src/lib/auth.ts
```diff
@@ -101,16 +101,33 @@ export const startOAuthFlow = async (
     const { openUrl } = await import('@tauri-apps/plugin-opener')
     await openUrl(authUrl)
   } else {
-    popup = window.open(authUrl, '_blank', 'noopener,noreferrer,width=600,height=700')
-    if (!popup) throw new Error('Failed to open authentication window')
-    popup.focus()
+    // Open a popup window. Avoid using 'noreferrer'/'noopener' because we rely on window.opener
```
cjroth: Curious what caused the need for this change here since the auth popup worked well in the settings > integrations page

src/hooks/use-onboarding-state.ts
```diff
@@ -0,0 +1,326 @@
...
+  | { type: 'SET_LOCATION_VALUE'; payload: string }
+  | { type: 'SET_LOCATION_VALID'; payload: boolean }
+  | { type: 'SET_SUBMITTING_LOCATION'; payload: boolean }
+  | { type: 'SUBMIT_LOCATION'; payload: { locationName: string; locationLat: number; locationLng: number } }
+  | { type: 'NEXT_STEP' }
+  | { type: 'PREV_STEP' }
+  | { type: 'SKIP_STEP' }
+
+const initialState: OnboardingState = {
+  currentStep: 1,
```
cjroth, line 48: Sort of minor thing but personally I would prefer to name the steps in case we want to re-order them - eg `setCurrentStep(Steps.AddLocation)` or `setCurrentStep('add-location')` could be more robust and clear.

I'm ok with merging as-as if this is not a trivial change to make - just a suggestion / code-style preference

cjroth (changes requested):
This is great! Nice work!

- Code is super clean, well-organized, and easy to understand!
- Great test coverage and really happy to see happy-dom working (haha)
- I tested it and it was really well without any bugs!

Main requests from comments below:
- Curious to understand the cause of the auth.ts change
- Want to see the onboarding state logic tested
- A few UI requests that I will add in a video

src/lib/auth.ts
ital0: @cjroth just remembered why I added these blocks.

I was trying to understand why I couldn't have more control on the opened popup. I wasn't being able to detect if closed the window or canceled the flow, and that would completetly break what I was trying to do.

cjroth (changes requested):
I just made some style/copy changes and fixed some issues with the Connect Google button state getting stuck (I refactored the button into a new component that is shared with the integrations page).

I'm seeing some new issues now after doing this:
- IMO the unit tests should be checking business logic, not specific CSS classes or text on each screen. I feel like these CSS / copy tests don't really prevent business logic bugs and just create more hurdles when changing small things like copy
- We should be using bun:test not vitest
- There's some cursor bug bot comments - please review those and see if they are real

ital0:
> I just made some style/copy changes and fixed some issues with the Connect Google button state getting stuck (I refactored the button into a new component that is shared with the integrations page).
> 
> I'm seeing some new issues now after doing this:
> 
> * IMO the unit tests should be checking business logic, not specific CSS classes or text on each screen. I feel like these CSS / copy tests don't really prevent business logic bugs and just create more hurdles when changing small things like copy
> * We should be using bun:test not vitest
> * There's some cursor bug bot comments - please review those and see if they are real

@cjroth I wouldn't remove CSS validations to keep consistency with the layouts. But yes, I'll check the businesses tests.

cjroth (approved):

PR #197, opened 2025-10-18
THU-47: Enforce encryption state per thread

raivieiraadriano92:
> [!NOTE]
> Binds chat threads to model confidentiality, disables incompatible model choices and message sends, and refactors DAL to support thread encryption with new helpers and tests.
> 
> - **Chat UI/UX**:
>   - Enforce per-thread encryption: block send when `chatThread.isEncrypted !== selectedModel.isConfidential`; pass `modelId` in message metadata.
>   - Fetch thread via `getChatThread` and pass `chatThread` into `PromptInput`; require `chatThreadId` prop.
>   - Disable incompatible models in selector with tooltip; show encrypted-conversation banner using new `TimelineMessage` (lock icon).
>   - Replace inline trigger timeline markup with reusable `TimelineMessage`.
> - **Prompt Input**:
>   - New `chatThread` prop to gate model selection; tooltips explain disabled states.
> - **DAL (Threads/Messages/Prompts)**:
>   - `createChatThread(id, modelId)` → store `isEncrypted` from model; add `updateChatThread`; `getOrCreateChatThread(id, modelId)` required.
>   - `getChatThread`/`getAllChatThreads` typed; return `null` when not found.
>   - `getLastMessage` now returns full message or `null`; `saveMessagesWithContextUpdate` uses `getChatThread` and `updateChatThread` for `contextSize`.
>   - `runAutomation` uses `createChatThread` and model lookup; add `getPrompt` helper.
>   - Misc DAL APIs add explicit return types; `getHttpMcpServers` returns rows directly.
> - **Chat Detail**:
>   - Use `updateChatThread` to set title; pass modelId to `getOrCreateChatThread`; include model metadata on send.
> - **Settings (MCP Servers)**:
>   - Use shared `McpServer` type; handle nullable URLs in UI.
> - **Tests**:
>   - Update for new signatures (`modelId` required); expect `null` for missing last message; add comprehensive `updateChatThread` tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9075c3f36f212020547dfc1d0eaf9ae1f2aa47ec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/chats/detail.tsx
+++ b/src/chats/detail.tsx
@@ -1,4 +1,4 @@
-import { getModel, getChatMessages, getOrCreateChatThread, saveMessagesWithContextUpdate } from '@/dal'
+import { getChatMessages, getOrCreateChatThread, saveMessagesWithContextUpdate } from '@/dal'
 import { generateTitle } from '@/lib/title-generator'
 import { convertDbChatMessageToUIMessage } from '@/lib/utils'
 import type { SaveMessagesFunction, ThunderboltUIMessage } from '@/types'
--- a/src/chats/detail.tsx
+++ b/src/chats/detail.tsx
@@ -48,29 +48,14 @@ export default function ChatDetailPage() {
 
   const addMessagesMutation = useMutation({
     mutationFn: async (messages: ThunderboltUIMessage[]) => {
-      const isNewChat = params.chatThreadId === 'new'
-      let isEncrypted = false
-
       if (!chatThreadId) {
         throw new Error('No chat thread ID')
       }
 
-      if (isNewChat) {
-        const modelId = messages[0]?.metadata?.modelId
-        if (!modelId) {
-          throw new Error('No model ID')
-        }
-
-        const model = await getModel(modelId)
-        if (!model) {
-          throw new Error('No model found')
-        }
-
-        isEncrypted = model.isConfidential === 1
-      }
+      const modelId = messages[0]?.metadata?.modelId
 
       // Fetch thread info to check if we need to generate a title
-      const thread = await getOrCreateChatThread(chatThreadId, isEncrypted)
+      const thread = await getOrCreateChatThread(chatThreadId, modelId)
 
       // Save messages and update context size using DAL
       const dbChatMessages = await saveMessagesWithContextUpdate(chatThreadId, messages)
@@ -80,7 +65,7 @@ export default function ChatDetailPage() {
         updateThreadTitle(messages, chatThreadId)
       }
 
-      if (isNewChat) {
+      if (params.chatThreadId === 'new') {
         navigate(`/chats/${chatThreadId}`, { relative: 'path' })
       }
 
--- a/src/chats/detail.tsx
+++ b/src/chats/detail.tsx
@@ -55,7 +55,7 @@ export default function ChatDetailPage() {
       const modelId = messages[0]?.metadata?.modelId
 
       // Fetch thread info to check if we need to generate a title
-      const thread = await getOrCreateChatThread(chatThreadId, modelId)
+      const thread = await getOrCreateChatThread(chatThreadId, modelId ?? '')
 
       // Save messages and update context size using DAL
       const dbChatMessages = await saveMessagesWithContextUpdate(chatThreadId, messages)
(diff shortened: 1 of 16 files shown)


src/chats/detail.tsx
```diff
@@ -52,12 +49,29 @@ export default function ChatDetailPage() {
 
   const addMessagesMutation = useMutation({
     mutationFn: async (messages: ThunderboltUIMessage[]) => {
+      const isNewChat = params.chatThreadId === 'new'
+      let isEncrypted = false
```
cjroth: I'd rather use `model.isConfidential === 1` directly than do `let isEncrypted` - in my experience `let` can introduce bugs / is harder to reason about

src/settings/mcp-servers.tsx
```diff
@@ -419,7 +411,7 @@ export default function McpServersPage() {
                               }}
                               className="text-lg font-medium cursor-pointer"
                             >
-                              {formatServerTitle(server.url, server.id)}
+                              {formatServerTitle(server.url ?? '', server.id)}
```
cjroth, line 414: Was this a typescript thing?

src/components/chat/chat-ui.tsx
```diff
@@ -170,6 +179,13 @@ export default function ChatUI({
     const textToSend = input.trim()
     if (isStreaming || !textToSend) return
 
+    // Validate encryption state
+    if (chatThread && chatThread.isEncrypted !== selectedModel?.isConfidential) {
+      throw new Error(
+        `This model is not available for ${chatThread.isEncrypted === 1 ? 'encrypted' : 'unencrypted'} conversations.`,
+      )
+    }
```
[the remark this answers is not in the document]
cjroth: Should look into this

src/components/chat/chat-ui.tsx
```diff
@@ -90,6 +94,11 @@ export default function ChatUI({
     onOverflow: () => setShowOverflowModal(true),
   })
 
+  const { data: chatThread = null } = useQuery({
```
cjroth, line 97: This isn't really a request for this PR specifically, but I feel like all of these chat components (ChatUI, ChatState, detail.tsx, chat.tsx) are super confusing - it's like 4 layers deep of React components that all handle different database getters and state logic. I think we should consider refactoring all of this to separate display logic and state logic somehow. I'm picturing maybe somehow having a context provider with a useReducer and some useQueries in it or something like that and having all of these components just be mainly display logic. Something to think about for the future.

cjroth: In fact... I decided to make a ticket for this haha https://linear.app/mozilla-thunderbolt/issue/THU-131/refactor-chat-ui-component-hierarchy

src/chats/detail.tsx
```diff
@@ -52,12 +49,29 @@ export default function ChatDetailPage() {
...
+
       if (!chatThreadId) {
         throw new Error('No chat thread ID')
       }
 
+      if (isNewChat) {
+        const modelId = messages[0]?.metadata?.modelId
```
cjroth: In general I feel like this code could somehow be more readable/cleaner. Just a lot of code branching going on here

cjroth (changes requested):
This works great and I think the UX is a good starting point until we get deeper into design details of how this should look and behave to the user. I love it.

src/settings/mcp-servers.tsx
raivieiraadriano92: Yes, more exactly here: `src/dal/mcp-servers.ts -> getHttpMcpServers -> url: server.url as string`

`url` type was being forced to string but the value itself still could be null which could cause some unexpected error.

src/components/chat/chat-ui.tsx
raivieiraadriano92: Totally agree - I think the React.Context would fit really well for this use case. I would love to work on this refactor :D

src/chats/detail.tsx
raivieiraadriano92: Once each thread should strictly be marked as encrypted or unencrypted upon creation - I thought it would make sense to move that logic to createChatThread

src/components/chat/chat-ui.tsx
raivieiraadriano92: It could be useful for this refactoring: https://ai-sdk.dev/cookbook/next/use-shared-chat-context

cjroth: Amazing! I'll mark this as resolved here and we can continue that convo in Linear :)

cjroth (approved):
Let's ship it

PR #199, opened 2025-10-20
THU-129: Trigger display components with XML tags instead of tool calls.

cjroth:
Test the concept of allowing the LLM to respond using XML-style tags for display components, instead of initiating tool calls for UI elements.

## Context

Previously, when the LLM attempted to display UI elements as part of its internal chain-of-thought and tool calls, it resulted in responses where display elements were mixed with thought processes and tool interactions. This led to cluttered and confusing outputs that included UI elements during information gathering, rather than as part of the final user response.

The purpose of this experiment is to determine if including display elements as part of the final LLM response—using XML-style tags—helps generate more clearly formatted, visually appealing responses without duplicated information. The goal is to improve the user output by ensuring display elements are separated from the LLM's internal reasoning and tool use steps.

---

> [!NOTE]
> Implements an XML widget architecture (weather, link-preview) with parsing/rendering, per-message caching, and new backend link-preview/proxy enhancements.
> 
> - **Frontend**:
>   - **Widget System**: Add registry (`src/widgets`), auto parsers/types (`src/ai/widget-parser.ts`, `src/ai/widget-types.ts`), and renderer to embed widgets via `<widget:... />` tags.
>   - **Widgets**: Implement `weather-forecast` and `link-preview` components with Storybook stories and skeletons.
>   - **Text Rendering**: `TextPart` now parses mixed text/widget content; `AssistantMessage` passes `messageId` and removes bespoke display handler; `ToolPart` memoization.
>   - **Prompt**: Overhaul system prompt, include `widgetPrompts`, efficiency/usage rules for tools and previews.
>   - **Caching**: New `useMessageCache` for lazy per-message data enrichment.
>   - **UX/Infra**: Favicon proxy URLs are encoded; Storybook loads `stories.tsx` paths.
> - **Backend**:
>   - **Link Preview API**: New `/pro/link-preview/*` route to fetch/parse OG/Twitter metadata with UA, timeout, and CORS; comprehensive tests.
>   - **Proxy**: Improve `/pro/proxy/*` decoding/validation, CORS/headers, error handling; add tests for encoding and streaming.
>   - **Exa**: Tweak `getContents` options (timeout/extras/text) and update tests; mount link-preview routes.
> - **Data Layer**:
>   - **DB**: Add `chat_messages.cache` JSON column + migration; update DAL (`getMessage`, `updateMessageCache`, typed returns).
>   - **Defaults/Settings**: Set `cloud_url` default from env.
> - **Integrations**:
>   - Split Pro client into typed `api.ts` + `schemas.ts`; update tools to re-export API and simplify configs; adjust tests to new shapes/timeouts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 341e94207fdf076423d46bcc2814a3e65e37689a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/lib/assistant-message.ts
```diff
@@ -30,16 +64,16 @@ export const groupToolParts = (parts: GroupableUIPart[]): GroupedUIPart[] => {
     currentGroup = []
   }
 
-  // Walk through the incoming parts and buffer every consecutive non-display tool call.
+  // Walk through the incoming parts and buffer every consecutive tool call.
   parts.forEach((part) => {
-    const [partType, toolName] = splitPartType(part.type)
+    const [partType] = splitPartType(part.type)
```
cjroth, line 69: I removed the part that filters out `display-` tools here because we no longer use them

raivieiraadriano92 (approved):
It looks great - much faster

Just a thought that I had - now we fetch the widget data inside of the widget component - in the future wouldn't this cause outdated information?

For example, if the user request the weather forecast today and tomorrow he open the same chat thread, the information would be reloaded. I mean, previous data would be lost so the user can't keep a "history".

backend/src/pro/link-preview.ts
```diff
@@ -0,0 +1,134 @@
+import { getCorsOrigins, getSettings } from '@/config/settings'
```
cjroth, line 1: I added this endpoint to the backend to fetch preview images and titles because the exa fetch contents endpoint is slow and expensive. In the future we may want to add caching.

src/dal/chat-messages.ts
```diff
@@ -1,8 +1,16 @@
...
 import { DatabaseSingleton } from '../db/singleton'
 import { chatMessagesTable, chatThreadsTable } from '../db/tables'
-import type { ThunderboltUIMessage, UIMessageMetadata } from '../types'
 import { convertUIMessageToDbChatMessage } from '../lib/utils'
+import type { ThunderboltUIMessage, UIMessageMetadata } from '../types'
+
+/**
+ * Gets a single chat message by ID
+ */
+export const getMessage = async (messageId: string) => {
```
raivieiraadriano92: just a minor suggestion - there are some DAL methods which specify the returning type as `Promise<T | null>` which enforce the type safety.

What do you think about? For the `getMessage` it would be like `Promise<ChatMessage | null>`

I have done this for all remaining method in my encryption state PR.

src/dal/chat-messages.ts
```diff
@@ -88,3 +96,43 @@ export const saveMessagesWithContextUpdate = async (threadId: string, messages:
 
   return dbChatMessages
 }
+
+/**
+ * Updates a specific cache field for a message
+ * Uses JSON patch-like syntax for nested keys (e.g., "linkPreviews.https://example.com")
+ * Note: Only splits on the FIRST dot to avoid splitting URLs
+ */
+export const updateMessageCache = async (messageId: string, cachePath: string, value: unknown) => {
```
raivieiraadriano92: Same here but for these void cases it would enforce the returning `Promise<void>`.

What do you think?

src/db/tables.ts
```diff
@@ -28,6 +28,9 @@ export const chatMessagesTable = sqliteTable('chat_messages', {
     .references(() => chatThreadsTable.id, { onDelete: 'cascade', onUpdate: 'cascade' }),
   modelId: text('model_id').references(() => modelsTable.id),
   parentId: text('parent_id').references((): any => chatMessagesTable.id, { onDelete: 'cascade' }),
+  cache: text('cache', { mode: 'json' }).$type<{
+    linkPreviews?: Record<string, { title: string; description: string; image: string | null }>
```
raivieiraadriano92: Just confirming - shouldn't we have a key here for `weather` as well?

src/ai/visual-parser.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Format: <visual-name attr="value" attr2="value2" />
+ * Example: <weather-forecast location="Seattle" region="Washington" country="USA" />
+ */
+export const parseContentParts = (text: string): ContentPart[] => {
+  const parts: ContentPart[] = []
+
+  // Match any self-closing tag with kebab-case naming
+  // Captures: tagName and attributes
+  const visualTagRegex = /<([a-z][a-z0-9-]*)\s+((?:[^/]|\/(?!>))+)\/>/gi
```
ital0: and if you move this regex outside this function just to avoid its recreation everytime this function gets called?

ital0: Another suggestion would be to detect if texts contains `<` `>` tags. If not, you can have an early return

```typescript
if (!text.includes('<') || !text.includes('/>')) {
    return text.trim() ? [{ type: 'text', content: text.trim() }] : []
}
```

src/ai/visual-parser.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    lastIndex = matchIndex + fullTag.length
+  }
+
+  // Add remaining text after the last match (or all text if no matches)
+  if (lastIndex < text.length) {
+    let textAfter = text.slice(lastIndex).trim()
+
+    // Remove incomplete tags at the end (for streaming)
+    const incompleteTagMatch = textAfter.match(/<[^>]*$/)
```
ital0: noticed that this regex is detecing any incomplete tags (including html ones), not only the visual one that I believe this function is trying to catch. If this assumption is correct, you might want to update this to:

```typescript
const incompleteVisualTagMatch = textAfter.match(/<[a-z][a-z0-9-]*\s+[^>]*$/)
if (incompleteVisualTagMatch) {
  textAfter = textAfter.slice(0, incompleteVisualTagMatch.index).trim()
}
```

src/ai/visual-types.ts
```diff
@@ -0,0 +1,30 @@
...
+  args: z.object({
+    url: z.string().describe('The URL to preview'),
+  }),
+})
+
+/**
+ * Union of all possible visual types
+ * Add new visual types here as they're created
+ */
+const _VisualSchema = z.discriminatedUnion('visual', [WeatherForecastVisualSchema, LinkPreviewVisualSchema])
```
ital0: question: are you using `_` in the beggining because it's a union type?

src/dal/chat-messages.ts
ital0: in that case, I would add return types to all methods in this file.

src/hooks/use-message-cache.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      if (!message) {
+        throw new Error(`Message ${messageId} not found`)
+      }
+
+      // Navigate the cache path to check for cached value
+      // Only split on first dot to handle URLs properly
+      const firstDotIndex = cacheKey.indexOf('.')
+
+      let current: any = message.cache
```
ital0: would it make sense to have these actions in a helper function? So would allow other components to reuse, if needed.

src/integrations/thunderbolt-pro/api.ts
```diff
@@ -0,0 +1,267 @@
...
+import { WeatherForecastDataSchema, type WeatherForecastData } from '@/lib/weather-forecast'
+import ky from 'ky'
+import { z } from 'zod'
+
+const requestTimeout = 10000
+
+/**
+ * Schemas for the pro API
+ */
+export const searchSchema = z
```
ital0: just for consistency and maintanability, would it be possible to move these schemas to an external file?

ital0 (changes requested):

src/dal/chat-messages.ts
cjroth: Agreed, added

src/dal/chat-messages.ts
cjroth: Agreed

src/ai/visual-parser.ts
cjroth: Hmm good catch. The syntax you suggested would also still match html, but I do think it's better.

I thought about LLM-style sentinels like `<|link-preview /|>` but these are intended more for LLM servers (eg LiteLLM) to parse directly from the model, so I don't think we should use that. I think in realty when I ask the LLM to show me HTML, it puts it in a code block (eg using backticks) so it's ok to strip out HTML here.

src/ai/visual-types.ts
cjroth: Screenshot 2025-10-21 at 5 42 47 PM

src/hooks/use-message-cache.ts
cjroth: Good point - and actually - I think this was over-engineers in the first place - I'm going to simply it to just simple key-value.

.storybook/main.ts
```diff
@@ -1,7 +1,7 @@
 import type { StorybookConfig } from '@storybook/react-vite'
 
 const config: StorybookConfig = {
-  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
+  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)', '../src/**/stories.@(js|jsx|mjs|ts|tsx)'],
```
cjroth, line 4: Making it so that we can also create just stories.tsx files (when we want the folder name to indicate what it's for)

src/ai/widget-types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { widgetSchemas } from '@/widgets'
+import { z } from 'zod'
+
+/**
+ * Union of all possible widget types - auto-generated from widget registry
+ */
+const _WidgetSchema = z.discriminatedUnion(
+  'widget',
+  widgetSchemas as unknown as readonly [z.ZodObject<any>, z.ZodObject<any>, ...z.ZodObject<any>[]],
```
cjroth: Uggg need to fix this

raivieiraadriano92 (commented):
LGTM

src/components/chat/text-part.tsx
```diff
@@ -1,18 +1,42 @@
...
+import { WidgetRenderer } from './widget-renderer'
 
 interface TextPartProps {
   part: TextUIPart
+  messageId: string
 }
 
-export const TextPart = ({ part }: TextPartProps) => {
+// Animation classes for subtle slide-in effect
+const animationClasses = 'animate-in slide-in-from-bottom-2 fade-in duration-300 ease-out'
```
cjroth: This doesn't need to be up here

src/components/chat/widget-renderer.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const widgetConfig = widgetRegistry.find((w) => w.name === widget.widget)
+
+  if (!widgetConfig) {
+    return null
+  }
+
+  // Type safety is ensured by the widget registry - widget.args matches the component's props
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return createElement(widgetConfig.module.Component as any, {
+    ...(widget.args as any),
```
cjroth: Fix these 2 anys

src/widgets/link-preview/widget.tsx
```diff
@@ -0,0 +1,117 @@
...
+  if (!data) {
+    return <LinkPreview title={url} description="Failed to load preview" url={url} image={null} />
+  }
+
+  const imageUrl = data.image && cloudUrl.value ? `${cloudUrl.value}/pro/proxy/${encodeURIComponent(data.image)}` : null
+
+  return <LinkPreview title={data.title} description={data.description} url={url} image={imageUrl} />
+}
+
+export const LinkPreview = ({ description, image, title, url }: LinkPreviewProps) => {
```
cjroth: should get split into display.tsx

PR #210, opened 2025-10-27
fix(THU-135): fix backend CORS regex

ital0:
Requests were blocked by CORS on Android mobile WebViews, causing "Access-Control-Allow-Origin" errors.

## Root Cause

- Backend server only listened on localhost (not accessible from mobile devices) - I had to force the server in my local to 0.0.0.0.
- CORS regex didn't include mobile Tauri origins (http://tauri.localhost)

## Solution

- Configure server to listen on 0.0.0.0 to accept mobile device connections (for local development only)
- Update CORS regex to include mobile Tauri origins: tauri://localhost, http://tauri.localhost, null, and file:// schemes

---

> [!NOTE]
> Expand default CORS origin regex and its env fallback to include Tauri, file, and null origins.
> 
> - **Backend config (`backend/src/config/settings.ts`)**:
>   - Expand default `corsOriginRegex` to allow `tauri://localhost`, `http://tauri.localhost`, `http://localhost:\d+`, `null`, and `file://.*`.
>   - Align env fallback for `CORS_ORIGIN_REGEX` with the new default pattern.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2070f2e33d23ea06982152324a9633257a35dd8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/config/settings.ts
+++ b/backend/src/config/settings.ts
@@ -27,7 +27,9 @@ const settingsSchema = z.object({
 
   // CORS settings
   corsOrigins: z.string().default('http://localhost:1420'),
-  corsOriginRegex: z.string().default('^(tauri:\/\/localhost|http:\/\/localhost:\d+)$'),
+  corsOriginRegex: z
+    .string()
+    .default('^(tauri://localhost|http://tauri\\.localhost|http://localhost:\\d+|null|file://.*)$'),
   corsAllowCredentials: z.boolean().default(true),
   corsAllowMethods: z.string().default('GET,POST,PUT,DELETE,PATCH,OPTIONS'),
   corsAllowHeaders: z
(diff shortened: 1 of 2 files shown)


backend/src/config/settings.ts
```diff
@@ -27,7 +27,9 @@ const settingsSchema = z.object({
 
   // CORS settings
   corsOrigins: z.string().default('http://localhost:1420'),
-  corsOriginRegex: z.string().default('^(tauri:\/\/localhost|http:\/\/localhost:\d+)$'),
+  corsOriginRegex: z
```
cjroth, line 30: I think it might be better to just accept any string here (or any valid url) - this way we don't need to update this regex + deploy if we want to test a new pattern in prod for a hotfix like this in the future

backend/src/index.ts
```diff
@@ -41,6 +41,7 @@ const createApp = async (fetchFn: typeof fetch = globalThis.fetch) => {
 
   return (
     configuredApp
+      .use(createLoggerMiddleware(settings))
```
cjroth: Curious why this one

ital0: @cjroth just making sure that the logs won't be overwritten by any other implementation

backend/src/config/settings.ts
ital0: @cjroth I was thinking about some kind of vulnerability or security issue.

backend/src/index.ts
ital0: but I can definitely rollback if you think it will cause anything.

cjroth: I'd say safer to not change it unless we need to

backend/src/config/settings.ts
cjroth: Hmm can't think of anything, bc that would mean someone would need access to the env, at which point things would already be bad. Thoughts?

ital0: I couldn’t think of anything either, but I preferred to err on the side of caution and make sure that, theoretically, nothing would slip through.
Are you sure you’d rather have it open?

cjroth: no strong opinion - just want to make our lives easier if we end up needing to debug the regex pattern. up to you

ital0: well I just added Tauri-specific patterns. It shouldn't impact on anything else. My opinion is that we should go with this pattern.

cjroth: sounds good, lets do it

PR #215, opened 2025-10-29
feat(THU-138): improving app loading error handling

ital0:
Enhanced error handling during app initialization to distinguish between database errors and other failures, ensuring PostHog fetch failures don't block app loading.

- Structured Error System: Added InitError types with specific error codes for granular error handling
- Non-blocking PostHog: PostHog initialization failures no longer prevent app from loading
- Conditional UI: "Clear Local Database" button only appears for database-related errors
- Enhanced UX: Added "Contact Support" button with error details, improved warning dialog

---

> [!NOTE]
> Refactors app initialization into a hook with typed errors, adds an AppErrorScreen, and makes PostHog initialization optional with error tracking.
> 
> - **App Initialization**:
>   - Introduces `useAppInitialization` to orchestrate app dir, DB init/migrate, defaults, tray, and PostHog; exposes `initData`, `initError`, `isInitializing`, `retry`, `clearDatabase`.
>   - Updates `src/app.tsx` to use the hook; replaces inline init; renders `AppErrorScreen` on failure and keeps loading state.
> - **Error Handling**:
>   - Adds structured error types (`src/types/handle-errors.ts`) and `createHandleError` util; integrates `trackError` for analytics.
>   - New `AppErrorScreen` with conditional "Clear Local Database" for DB errors and "Contact Support" email.
> - **Analytics**:
>   - Changes `initPosthog` to return `HandleResult` and be non-blocking; adds URL sanitization and error tracking via `captureException`.
> - **UI**:
>   - Adds `src/components/app-error-screen.tsx` and Storybook stories.
> - **Tests**:
>   - Adds/updates tests for initialization hook, error utils, and PostHog (`sanitizeUrl`, `before_send`, `trackError`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1585bf7fcb38bcbc0ae2054fb8218fe5099c87ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/lib/posthog.tsx
```diff
@@ -31,57 +31,67 @@ export const sanitizeUrl = (url: string): string => {
...
+          }
+
+          return event
+        },
+      }) as PostHog
+    }
+
+    return { success: true, client: posthogClient }
+  } catch (error) {
+    // TODO: track these errors in some analytics tool.
```
cjroth: Ironically we are using Posthog as our frontend error tracking tool 😆

src/app.tsx
```diff
@@ -98,91 +103,223 @@ function AppRoutes(_: { initData: InitData }) {
...
-                onClick={handleClearDatabase}
-                className="bg-destructive text-white hover:bg-destructive/90"
-              >
-                Clear Database
-              </AlertDialogAction>
-            </AlertDialogFooter>
-          </AlertDialogContent>
-        </AlertDialog>
+        <div className="flex flex-col gap-3">
+          {isDatabaseError && (
```
cjroth: I think it would be nice to pull this error handling component into a separate file + add to storybook so we can keep this app.tsx file more focused

src/app.tsx
```diff
@@ -98,91 +103,223 @@ function AppRoutes(_: { initData: InitData }) {
...
-  const databaseType = await getDatabaseType()
-  const dbPath = await getDatabasePath(databaseType, appDirPath)
+const init = async (): Promise<InitResult<InitData>> => {
+  try {
+    // Step 1: App directory creation
+    let appDirPath: string
+    try {
+      appDirPath = await createAppDir()
+    } catch (error) {
+      // TODO: track these errors in some analytics tool.
```
cjroth: I think for all of these other than Posthog, we can track them in Posthog - could you do a little research into the best way to do that? I think we could either track as regular events or maybe using their error tracking feature - not sure which is best suited

src/app.tsx
```diff
@@ -98,91 +103,223 @@ function AppRoutes(_: { initData: InitData }) {
   )
 }
 
-const init = async (): Promise<InitData> => {
-  const appDirPath = await createAppDir()
-  const databaseType = await getDatabaseType()
-  const dbPath = await getDatabasePath(databaseType, appDirPath)
+const init = async (): Promise<InitResult<InitData>> => {
```
cjroth: I feel like we should pull this init method into a separate file and also add unit tests for it - it's getting more a bit heavier

src/lib/posthog.tsx
```diff
@@ -31,57 +31,67 @@ export const sanitizeUrl = (url: string): string => {
   return url
 }
 
+type PostHogInitResult =
```
cjroth: Should we just use InitResult type (the one you added below) instead?

src/types/init-errors.ts
```diff
@@ -0,0 +1,25 @@
+export type InitErrorCode =
```
cjroth: I think this makes sense - I actually think maybe we should generalize this to be more than just "init" errors and maybe we use this for all errors that we specifically want to intercept and handle at the UI level like this. I still think it makes sense to handle the try/catch and UI the way you did in app.tsx, but we might follow the same pattern in other places as well. So basically I think I'm just saying, maybe we name these like "HandledErrorCode" instead of "InitErrorCode" or something along those lines (whatever a good general-purpose name would be). Thoughts?

src/app.tsx
```diff
@@ -98,91 +103,223 @@ function AppRoutes(_: { initData: InitData }) {
...
+                  </AlertDialogAction>
+                </AlertDialogFooter>
+              </AlertDialogContent>
+            </AlertDialog>
+          )}
+
+          <Button
+            variant="outline"
+            onClick={() =>
+              window.open(
```
cjroth: It would be super helpful if it could include the stack trace in here

cjroth (commented):
I tested this locally and it works perfectly! Just have some thoughts about code organization and reusability

src/lib/posthog.tsx
ital0: what should be done in this case? Since there won't be client available.

src/types/init-errors.ts
ital0: it makes sense :) it will follow its purpose, which is basically understand what failed and some detailing.

src/lib/posthog.tsx
cjroth: Tbh nothing can easily be done in this situation - just logging this is fine for now

src/app.tsx
ital0: initially I thought in adding it only for non production environments, but I'm moving forward with this to all envs.

src/app.tsx
ital0: encapsulating all this logic inside a hook. It is better to have React tree handling this instead of single functions and a single one orchestrating each call.

cjroth: Hmm yeah I could agree with that. Since it being called in a useEffect and has some aspect of state to it, I think that make sense

cjroth (approved):
Much better than before, let's merge :)

PR #218, opened 2025-11-03
THU-137: Prevent migration file misery

raivieiraadriano92:
> [!NOTE]
> Normalizes SQL quoting in migrations and tweaks the bundler to emit single-quoted, cleaner TS, regenerating the bundled migrations.
> 
> - **DB migrations**:
>   - Normalize identifier/column quoting to backticks in `0005` and `0006` `INSERT` statements.
>   - Update `mcp_servers.type` default to backticked `http` in `0000`.
>   - Regenerate `src/drizzle/_migrations.ts` to reflect sanitized SQL and quoting changes.
> - **Bundler (`src/db/bundle-migrations.ts`)**:
>   - Adjust generated TS: unquoted object keys, single-quoted strings, trailing commas; preserve sanitized SQL.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c248fef6fd22a7eece51cef5e3ddf32da99aa745. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cjroth (changes requested):
These changes look good to me - but I'm wondering what the original cause of the formatting differences is. I suspect the drizzle command (`bun db generate`) could be the problem, or maybe code editor differences. Can we make sure we know how to prevent this from happening in the future?

PR #211, opened 2025-10-27
THU-131: Refactor chat UI component hierarchy

raivieiraadriano92:
…ailPage

---

> [!NOTE]
> Replaces chat state with a Zustand store and new hydration hooks, overhauls Chat UI into modular components with reasoning/tool grouping and duration tracking, and updates DAL/AI fetch to support it.
> 
> - **State Management & Architecture**:
>   - Replace `Chat`/`ChatState` with a centralized Zustand store (`src/chats/chat-store.ts`) and hydration (`use-hydrate-chat-store.ts`); remove legacy `src/chats/chat-state.tsx` and `src/chats/chat.tsx`.
>   - Add hooks for automation and scrolling: `use-chat-automation.tsx`, `use-chat-scroll-handler.ts`.
>   - Introduce throttled partial-assistant save handler `save-partial-assistant-messages-handler.ts`.
> - **Chat UI Modularization**:
>   - Split UI into `ChatMessages`, `ChatPromptInput`, `SuggestionButtons`, `EncryptionMessage`, `ErrorMessage`; refactor `chat-ui.tsx` to use store/hooks.
>   - Extract `ModelSelector` and simplify `PromptInput` integration.
> - **Reasoning & Tools Rendering**:
>   - Add reasoning-tool grouping via new `reasoning_group` type and components: `ReasoningGroup`, `ReasoningDisplay`, `ReasoningGroupTitle`, `ReasoningItem`.
>   - Track and persist part durations; update assistant message flow and tests.
> - **AI Fetch & Streaming**:
>   - `ai/fetch.ts`: switch request payload to `{ id, messages }`, save by `id`, consolidate settings, enable reasoning extraction, enrich message metadata.
> - **DAL**:
>   - Add `updateMessage` and export; use it to persist updated parts/durations.
> - **Content View**:
>   - Update object view to support reasoning text and formatted tool outputs (`context.tsx`, `object-sidebar-content.tsx`).
> - **Styling & Icons**:
>   - Add shimmer CSS and apply to `ToolIcon`; minor `Expandable` trigger sizing.
> - **Automations**:
>   - Memoize model change handler in `automation-form-modal.tsx`.
> - **Dependencies**:
>   - Add `zustand` to `package.json` and lockfile.
> 
[...]

PR #219, opened 2025-11-03
THU-105: Create email/calendar connection display component

ital0:
The widget appears inline in chat when a task requires integration; users can select a provider or choose not to connect. After OAuth, it briefly shows a connected state, returns to the original chat, and the AI continues the original action using the newly connected integration. It handles navigation, page refreshes, and edge cases, and hides itself once an integration is successfully connected.

---

> [!NOTE]
> Introduces a connect-integration chat widget (Google/Microsoft) with OAuth completion auto-retry, hidden-state caching, and persists message metadata; updates chat flow, parser, hooks, settings, and DB schema.
> 
> - **Widgets**:
>   - **New `connect-integration`**: `schema`, `instructions`, `widget`, `constants`, registry wiring, hidden-state hook.
>   - Widget renderer filters hidden `connect-integration` via message cache.
> - **Chat Flow**:
>   - Auto-retry after OAuth via `useHandleIntegrationCompletion`; hides retry trigger messages (`oauthRetry`).
>   - OAuth callback supports path return context; safer `oauth_return_context` handling.
> - **Parsing & Tests**:
>   - `create-parser` allows empty-string attrs; add parser tests for `connect-integration`.
> - **Hooks**:
>   - Add `use-integration-status`, `use-connect-integration-widget-state`, `use-handle-integration-completion` (all with tests).
>   - `use-message-cache` gains `enabled` flag.
> - **Persistence/DB**:
>   - Add `chat_messages.metadata` column + migration; DAL saves/updates metadata; converters map metadata.
> - **Settings/UI**:
>   - New `integrations_do_not_ask_again` default setting.
>   - Extract provider icons; update Integrations page.
> - **Other**:
>   - Webview init timing fix in sidebar webview.
>   - Analytics: expose `resetPosthog` and test improvements.
>   - Bump `posthog-js`; pin `vite` version.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 224f5375f8b13a24309883044a2e2e0e0e8ca321. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cjroth (changes requested):
Really good work - works very well overall.

Changes requested:
* Design tweaks
    * Make outer border full width (same as chat response width like weather widget)
    * Hovering over Google / Microsoft should use cursor: pointer
    * Make Google + Microsoft icons a little bigger
    * Buttons on bottom should be bottom-aligned or remove space below them (whatever feels more appropriate)
* Instead of “do not connect”, let’s have a “not now” and “do not ask again” settings - similar to when apps request permission on iOS/android - (using settings hook to store this)
* Bug: when navigating to a previous conversation, the “choose different provider” button isn’t working (just flickers instead of navigating to previous state)
* Change title "Choose your email provider to connect Email \n to access your email" to just "Choose your email provider \n to access your email" (eg remove duplicated text)

src/chats/chat-state.tsx
```diff
@@ -3,16 +3,19 @@ import ChatUI from '@/components/chat/chat-ui'
 import { useSettings } from '@/hooks/use-settings'
```
cjroth: Just a heads up - we'll need to re-implement this part based on changes in https://github.com/thunderbird/thunderbolt/pull/211

cjroth (changes requested):

src/lib/create-parser.ts
```diff
@@ -22,14 +22,25 @@ export const createParser = <T extends z.ZodObject<any>>(
   const argsKeys = Object.keys(argsSchema.shape)
 
   return (attrs: Record<string, string>): z.infer<T> | null => {
-    // Quick check: ensure all required args are present
-    const hasAllArgs = argsKeys.every((key) => attrs[key] !== undefined && attrs[key] !== '')
-    if (!hasAllArgs) {
+    // Check which fields are optional by inspecting the schema
```
cjroth: We should test this thoroughly with different models - I was having problems with optional arguments before where the model was getting confused about when to provide optional args and when not to. In my experience, the LLM has the best luck when all fields were always required (even we we just ignored them sometimes).

In my testing, this seems to be fine atm for the connect-integration widget, but I am weary of adding a new pattern that will confuse LLMs in the future as the number of widgets grows.

src/widgets/connect-integration/instructions.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * AI Instructions for the connect-integration widget
+ */
+export const instructions = `## Connect Integration
+
+<widget:connect-integration provider="google" service="calendar" reason="to view your upcoming events" />
+
+### CRITICAL: Always Check Available Tools First
```
cjroth: Wouldn't the tools here not be available if the user hasn't connected something yet?

cjroth (changes requested):

src/widgets/connect-integration/instructions.ts
ital0: they might be failing...?

cjroth:
I was thinking about this logic where we mutate the integration logic a bit - I think this current approach is a bit complex. Would this make sense to simplify it?

```
### Current Flow (Complex):
1. Widget message exists at position N
2. User clicks OAuth → completes
3. System calls `regenerate()` → **destroys widget message**
4. If credentials failed, manually restore widget with `restoreWidgetMessage()`

### Your Simplified Flow (Better):
1. Widget message exists at position N
2. User clicks OAuth → completes  
3. Widget **updates its display** to show "✓ Connected to Google"
4. System appends a hidden user message: `{ role: 'user', content: '[System: OAuth completed, retry request]' }`
5. The effect above **automatically triggers** because last message is now `role === 'user'`
6. Assistant generates new response with valid credentials
7. Widget stays in original position (immutable messages ✓)

### Benefits:

✅ **Messages are immutable** - no editing/repositioning  
✅ **Widget is stateful** - updates display based on OAuth state  
✅ **Reuses existing automation trigger** - same mechanism as lines 195-212  
✅ **Much simpler** - no `restoreWidgetMessage`, no complex refs, no sessionStorage coordination

### Implementation:

You'd just need to:
1. Make the widget read OAuth state and update its display (keep it in original position)
2. After OAuth, append a user message to trigger the assistant
3. Remove all the `oauthRetryRef`, `restoreWidgetMessage`, and related complexity

File: src/widgets/connect-integration/instructions.ts (hunk 3 above)
Comment (cjroth) [reply src/widgets/connect-integration/instructions.ts]: Walk me through the use case, I'm not following what you mean

Comment (ital0) [reply src/widgets/connect-integration/instructions.ts]: @cjroth the thought here was for the case where some tool has a bug in the code (or it's failing during its initialization for any other problem), and thus it wouldn’t be possible for the model to use it.

Comment (ital0) [reply src/widgets/connect-integration/instructions.ts]: actually, I’ve never witnessed a failure of this kind, but I just imagined this small edge case

Comment (cjroth) [reply src/widgets/connect-integration/instructions.ts]: Hmm. This could happen with MCP servers. But if I'm understanding what you mean correctly, the way our current logic works, this would not result in the tool being missing from the model context - it would just result in errors when the model calls it. In other words, we would have to add logic to proactively check if tools are in errors states for this to work - which may be true for MCP in the future but I don't think we'll use MCP for Google/Microsoft.

I guess what you're thinking about here is, have the model *not* show this widget if for example Google tools were not available? But in that case, would we want to still show it for the Microsoft part?

I think it's safe to assume these tools will always be available honestly. In a situation where Google is not available at all, I think it would make sense to disable / hide Google in the widget itself so that other providers like Microsoft are still visible

Comment (ital0) [reply src/widgets/connect-integration/instructions.ts]: makes sense. So I take this assumption and move forward updating the instruction.

**Comment by ital0** [comment]
> I was thinking about this logic where we mutate the integration logic a bit - I think this current approach is a bit complex. Would this make sense to simplify it?
> 
> ```
> ### Current Flow (Complex):
> 1. Widget message exists at position N
> 2. User clicks OAuth → completes
> 3. System calls `regenerate()` → **destroys widget message**
> 4. If credentials failed, manually restore widget with `restoreWidgetMessage()`
> 
> ### Your Simplified Flow (Better):
> 1. Widget message exists at position N
> 2. User clicks OAuth → completes  
> 3. Widget **updates its display** to show "✓ Connected to Google"
> 4. System appends a hidden user message: `{ role: 'user', content: '[System: OAuth completed, retry request]' }`
> 5. The effect above **automatically triggers** because last message is now `role === 'user'`
> 6. Assistant generates new response with valid credentials
> 7. Widget stays in original position (immutable messages ✓)
> 
> ### Benefits:
> 
> ✅ **Messages are immutable** - no editing/repositioning  
> ✅ **Widget is stateful** - updates display based on OAuth state  
> ✅ **Reuses existing automation trigger** - same mechanism as lines 195-212  
> ✅ **Much simpler** - no `restoreWidgetMessage`, no complex refs, no sessionStorage coordination
> 
> ### Implementation:
> 
> You'd just need to:
> 1. Make the widget read OAuth state and update its display (keep it in original position)
> 2. After OAuth, append a user message to trigger the assistant
> 3. Remove all the `oauthRetryRef`, `restoreWidgetMessage`, and related complexity
> ```

@cjroth really appreciate this suggestion. I've been trying to simplify my implementation because I noticed it was too complex, but I couldn't figure out how to do it. I tried different approach, but this one decreased the implementation by ~80%!

**Comment by cjroth** [comment]
> @cjroth really appreciate this suggestion. I've been trying to simplify my implementation because I noticed it was too complex, but I couldn't figure out how to do it. I tried different approach, but this one decreased the implementation by ~80%!

Awesome! Yeah this chat state stuff is really tricky to keep simple. Glad to hear it worked :)

File: src/widgets/connect-integration/widget.tsx [hunk 4]
```diff
@@ -0,0 +1,464 @@
...
+
+    useEffect(() => {
+      const locationState = location.state as { oauth?: { code?: string; state?: string; error?: string } } | null
+      const oauth = locationState?.oauth
+
+      if (oauth) {
+        handleOAuthCallback(oauth)
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [location.state, messageId])
```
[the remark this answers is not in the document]
ital0: not a bug

src/chats/chat-state.tsx
```diff
@@ -152,16 +140,88 @@ export default function ChatState({ id, models, initialMessages, saveMessages }:
...
+        await chatHelpers.regenerate()
+      } catch (err) {
+        console.error('Failed to process OAuth retry:', err)
+        oauthRetryHandledRef.current.delete(widgetMessageId)
+      }
+    }
+
+    window.addEventListener(oauthRetryEvent, handleOAuthRetry as unknown as (event: Event) => void)
+    return () => window.removeEventListener(oauthRetryEvent, handleOAuthRetry as unknown as (event: Event) => void)
+  }, [status, chatMessages, chatHelpers, id, saveMessages, queryClient])
```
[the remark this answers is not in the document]
ital0: since I'm going to refactor this file that won't exists after #211, I won't check this bug report.

PR #248, opened 2025-11-19
THU-140: Fix Google/Microsoft login on mobile - Universal Link approach

ital0:
Implements verified HTTPS deep links (aka App Links (Android) or Universal Links (iOS)) (https://thunderbolt.io/oauth/callback) for seamless mobile OAuth with Google/Microsoft. Mobile devices now use system browser with PKCE flow. Includes deep link listener hook, comprehensive tests (33 passing), and refactored utilities for better code reuse.

---

> [!NOTE]
> Adds verified HTTPS deep links for mobile OAuth (with PKCE) and a deep-link listener, refactors settings to batch `updateSettings`, and updates Google/Microsoft auth plus new Microsoft tools.
> 
> - **Mobile Deep Linking & Tauri**:
>   - Add `@tauri-apps/plugin-deep-link`, initialize in `src-tauri/src/lib.rs`, and configure in `src-tauri/tauri.conf.json` with `https://thunderbolt.io/oauth/callback`.
>   - iOS entitlements (`associated-domains`) and Android `public/.well-known/assetlinks.json` updated.
>   - Capabilities grant `deep-link:default` in `capabilities/*.json`.
> - **OAuth Flow (PKCE + Redirects)**:
>   - New `getOAuthRedirectUri` and PKCE utils (`generateCodeVerifier/Challenge`).
>   - Google/Microsoft auth use platform-aware redirect + PKCE; desktop webview vs mobile system browser.
>   - New `useDeepLinkListener` to parse OAuth callbacks and navigate; integrated in `src/app.tsx`.
> - **Settings API Refactor**:
>   - Replace `updateSetting` with batch `updateSettings` across app, hooks, tests, and widgets.
> - **Integrations**:
>   - Add Microsoft Graph tools (`listMessages`, `getMessage`) and types.
>   - Update onboarding, integrations page, and connect-integration widget to new OAuth/return-context flow.
> - **Tests**:
>   - Comprehensive tests for deep-link listener, PKCE, OAuth redirect, settings, integrations, and tools.
> - **Dependencies**:
>   - Add `@tauri-apps/plugin-deep-link`; update lockfiles.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 21ca04280a3a11e3f2be9768c063b28331b39785. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/dal/settings.ts
+++ b/src/dal/settings.ts
@@ -231,41 +231,71 @@ export const createSetting = async (key: string, value: string | null): Promise<
 }
 
 /**
- * Update or create a setting in the settings table
- * @param key - The setting key
- * @param value - The value (can be string, null, boolean, number, or JSON-serializable object)
- * @param options - Optional configuration
- * @param options.recomputeHash - If true, updates the defaultHash to match the new value (makes it the new baseline for modification tracking)
- * @param options.updateHashOnly - If true, only updates the defaultHash using the provided value without changing the actual stored value (useful for updating the baseline without overwriting user customizations)
+ * Prepare a setting row for batch insert/update
  */
-export const updateSetting = async (
-  key: string,
-  value: any,
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const prepareSettingRow = (key: string, value: any, recomputeHash: boolean) => {
+  const stringValue = serializeValue(value)
+  return {
+    key,
+    value: stringValue,
+    ...(recomputeHash && { defaultHash: hashValues([key, stringValue]) }),
+  }
+}
+
+/**
+ * Update or create multiple settings at once in a single database operation
+ *
+ * @param settings - Object mapping setting keys to values
+ * @param options - Optional configuration (applies to all settings)
+ * @param options.recomputeHash - If true, updates the defaultHash to match the new value for all settings
+ * @param options.updateHashOnly - If true, only updates the defaultHash without changing actual values
+ *
+ * @example
+ * ```ts
+ * await updateSettings({
+ *   oauth_state: 'abc123',
+ *   oauth_provider: 'google',
+ *   oauth_verifier: 'xyz789',
+ * })
+ * ```
+ */
+export const updateSettings = async (
+  settings: Record<string, any>, // eslint-disable-line @typescript-eslint/no-explicit-any
   options: { recomputeHash?: boolean; updateHashOnly?: boolean } = {},
 ): Promise<void> => {
-  const db = DatabaseSingleton.instance.db
-  const stringValue = serializeValue(value)
+  const entries = Object.entries(settings)
+  if (entries.length === 0) return
 
+  const db = DatabaseSingleton.instance.db
+
+  // Handle updateHashOnly separately as it requires UPDATE statements
+  // Note: SQLite doesn't support batch UPDATE with different values per row,
+  // so we use parallel updates within a transaction for atomicity
   if (options.updateHashOnly) {
-    // Only update the hash, don't change the actual value
-    const newHash = hashValues([key, stringValue])
-    await db.update(settingsTable).set({ defaultHash: newHash }).where(eq(settingsTable.key, key))
+    await db.transaction(async (tx) => {
+      await Promise.all(
+        entries.map(([key, value]) => {
+          const stringValue = serializeValue(value)
+          const newHash = hashValues([key, stringValue])
+          return tx.update(settingsTable).set({ defaultHash: newHash }).where(eq(settingsTable.key, key))
+        }),
+      )
+    })
     return
   }
 
-  const updateFields: { value: string | null; defaultHash?: string | null } = { value: stringValue }
-
-  // If recomputeHash is true, update the defaultHash to match the new value
-  if (options.recomputeHash) {
-    updateFields.defaultHash = hashValues([key, stringValue])
-  }
+  // Single batch upsert for value updates
+  const values = entries.map(([key, value]) => prepareSettingRow(key, value, options.recomputeHash ?? false))
 
   await db
     .insert(settingsTable)
-    .values({ key, ...updateFields })
+    .values(values)
     .onConflictDoUpdate({
       target: settingsTable.key,
-      set: updateFields,
+      set: options.recomputeHash
+        ? { value: sql`excluded.value`, defaultHash: sql`excluded.default_hash` }
+        : { value: sql`excluded.value` },
     })
 }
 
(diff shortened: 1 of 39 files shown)


src/hooks/use-deep-link-listener.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    setupListener()
+
+    return () => {
+      if (unlisten) {
+        unlisten()
+      }
+    }
+  }, [handler, navigate, location.pathname, checkIsTauri, getCurrentUrls, listenToOpenUrl, getSettingsData])
```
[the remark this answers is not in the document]
ital0: Race condition that probably hardly will happen.

src/hooks/use-oauth-connect.ts
```diff
@@ -84,16 +85,48 @@ export const useOAuthConnect = (options: UseOAuthConnectOptions = {}): UseOAuthC
 
     try {
       if (isTauri()) {
-        const result = await startFlow(provider)
+        // Check if we're on mobile (iOS or Android)
+        const { platform } = await import('@tauri-apps/plugin-os')
```
cjroth: I think we could use one of these here https://github.com/thunderbird/thunderbolt/blob/2c3943ebf97ec87e9b450479f89be6f276be9f89/src/lib/platform.ts#L35

src/hooks/use-oauth-connect.ts
```diff
@@ -84,16 +85,48 @@ export const useOAuthConnect = (options: UseOAuthConnectOptions = {}): UseOAuthC
...
+        const currentPlatform = await platform()
+        const isMobile = currentPlatform === 'ios' || currentPlatform === 'android'
 
-        if (!result) return
+        if (isMobile) {
+          // For mobile: Open OAuth in system browser with App Link / Universal Link redirect
+          // The deep link listener will handle the callback
+          const { buildAuthUrl } = await import('@/lib/auth')
```
cjroth: I'm weary of lazy imports because we haven't figured out how something like this would act in different edge cases like offline / code splitting / etc - it's probably fine, but I've honestly just been avoiding dynamic imports as a whole for now with the goal of coming back to them intentionally in the future to think through code splitting and optimization

src/hooks/use-oauth-connect.ts
```diff
@@ -84,16 +85,48 @@ export const useOAuthConnect = (options: UseOAuthConnectOptions = {}): UseOAuthC
...
+          const { v4: uuidv4 } = await import('uuid')
 
-        const { tokens, userInfo } = result
+          const state = uuidv4()
+          const codeVerifier = generateCodeVerifier()
+          const codeChallenge = await generateCodeChallenge(codeVerifier)
 
-        await saveOAuthCredentials(provider, tokens, userInfo, { setPreferredName })
+          // Store OAuth state for callback validation
+          await updateSetting('oauth_state', state)
```
cjroth: I would love to change `updateSetting` to be `updateSettings` plural where we could update multiple values in a single sqlite query and keep the code a little cleaner. Not vital, but would be nice!

cjroth: (Then we'd no longer need updateSetting, since updateSettings would cover the single use case as well)

src/hooks/use-oauth-connect.ts
```diff
@@ -84,16 +85,48 @@ export const useOAuthConnect = (options: UseOAuthConnectOptions = {}): UseOAuthC
...
+          await updateSetting('oauth_provider', provider)
+          await updateSetting('oauth_verifier', codeVerifier)
+          await updateSetting('oauth_return_context', returnContext)
 
-        onSuccess?.()
+          const authUrl = await buildAuthUrl(provider, state, codeChallenge)
+
+          // Open in system browser (not webview)
+          const { openUrl } = await import('@tauri-apps/plugin-opener')
```
cjroth: This kind of Tauri-specific lazy import kinda makes sense though to me

cjroth (changes requested):
I'm really happy that we were able to achieve this with Universal links and a single client ID - this is super simple, clean, and will be maintainable. I'm also really happy that we have good unit tests for this, since this kind of thing feels like the type of thing that breaks easily, since it touches a lot of places.

Do we feel confident that this will work when merged? Is there a good way for me to test it first?

src/hooks/use-oauth-connect.ts
ital0: it makes sense in this case, specifically

src/hooks/use-oauth-connect.ts
ital0: I was sure that, just like `getSettings`, we had an `updateSettings`.
I’ll try to add it.

ital0:
@cjroth yes, I think we can move forward with the merge after I fix these small details.

You can test locally using an iOS simulator!
As for Android, we’d need to share the same keystore, because apparently Android validates the assetlinks.json for the app link to work. Besides that, I’m pushing a small fix for this file to correct the SHA256, which for now only includes the one I use locally and the production one!

src/hooks/use-oauth-connect.ts
ital0: Thanks for this feedback. 
I actually did some research about it and makes sense.

I believe the size increase we’ll get in terms of bundle size is worth it if it means making sure everything will work well and none of these imports will fail (removing all these lazy imports)

src/dal/settings.ts
```diff
@@ -231,41 +231,71 @@ export const createSetting = async (key: string, value: string | null): Promise<
...
+  const entries = Object.entries(settings)
+  if (entries.length === 0) return
+
   const db = DatabaseSingleton.instance.db
-  const stringValue = serializeValue(value)
 
+  // Handle updateHashOnly separately as it requires UPDATE statements
+  // Note: SQLite doesn't support batch UPDATE with different values per row,
+  // so we use parallel updates within a transaction for atomicity
   if (options.updateHashOnly) {
```
cjroth, line 275: Would something like this work?

--

```
await db.transaction(async (tx) => {
  const keys = Object.keys(hashMap);

  await tx
    .update(settingsTable)
    .set({
      defaultHash: sql`CASE
        ${keys
          .map(
            (k) => sql`WHEN ${settingsTable.key} = ${k} THEN ${hashMap[k]}`
          )
          .join(' ')}
        END`,
    })
    .where(inArray(settingsTable.key, keys));
});
````

cjroth: (Not a big deal - I think the transaction is good enough if not)

cjroth (approved):
Looks good :shipit: 

Other than minor question, let's merge 😎

PR #250, opened 2025-11-19
THU-173: <br> is showing up in results

ital0:
Initially I tried to fix this just by changing the instructions or create some other functions to normalize, but that wasn’t solving. Best solution was to use the `rehype-raw` library to fix it.

---

> [!NOTE]
> Converts HTML line breaks to real line breaks via custom ReactMarkdown components, fixing rendering in tables, lists, headers, and paragraphs.
> 
> - **Chat markdown rendering**:
>   - **Utility**: Add `src/components/chat/markdown-utils.tsx` with `markdownComponents` to process text and render HTML line breaks as React line breaks in `p`, `td`, `th`, and `li`.
>   - **Integration**: Use `components={markdownComponents}` in `src/components/chat/memoized-markdown.tsx`.
>   - **Tests**: Add `src/components/chat/markdown-utils.test.tsx` covering basic text, tables (including bullet lists), list items, headers, consecutive breaks, and mixed formatting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c6c9d3c5aa2c0acccc3aad293aacbedbeb926243. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -61,7 +61,6 @@
         "react-markdown": "^10.1.0",
         "react-resizable-panels": "^3.0.4",
         "react-router": "^7.8.1",
-        "rehype-raw": "^7.0.0",
         "remark-gfm": "^4.0.1",
         "tailwind-merge": "^3.3.1",
         "tailwindcss-animate": "^1.0.7",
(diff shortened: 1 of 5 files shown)


cjroth (commented):
Thoughts:

rehype-raw has 3 dependencies and it's last commit was about a year ago:

```
  "dependencies": {
    "@types/hast": "^3.0.0",
    "hast-util-raw": "^9.0.0",
    "vfile": "^6.0.0"
  },
```

AND it's touching private user data in the conversation.

Concerns I have:
- Maintained-ness
- Security
- Bundle size

And this is only a 1-line fix that brings those plugins together:
https://github.com/rehypejs/rehype-raw/blob/main/lib/index.js

hast and vfile are used only for types, so not truly needed

then `hast-until-raw` has a lot of deps:

```
  "dependencies": {
    "@types/hast": "^3.0.0",
    "@types/unist": "^3.0.0",
    "@ungap/structured-clone": "^1.0.0",
    "hast-util-from-parse5": "^8.0.0",
    "hast-util-to-parse5": "^8.0.0",
    "html-void-elements": "^3.0.0",
    "mdast-util-to-hast": "^13.0.0",
    "parse5": "^7.0.0",
    "unist-util-position": "^5.0.0",
    "unist-util-visit": "^5.0.0",
    "vfile": "^6.0.0",
    "web-namespaces": "^2.0.0",
    "zwitch": "^2.0.0"
  },
```


IMO it's not worth the risk / bundle size - I think we could create our own or implement a regex parser like we use for detecting <widget> tags instead

cjroth:
Also - if you go with the regex route here, note that I also added this which has some similarity (but probably different solution): https://linear.app/mozilla-thunderbolt/issue/THU-179/prevent-widget-tag-flicker

ital0:
@cjroth your analysis was very accurate. I agree 100%.

Actually my only filter here was the bundle size, since it would increase by around 3kb just with the library. But after reanalyzing it now with ALL its dependencies, it will actually add around 50kb to the bundle size.

Besides that, we still have the application’s privacy factor, which is basically the idea behind everything.

I’ll take this into consideration in the next tasks and I’ll refactor this solution.

cjroth:
Sounds good! Yeah with bundle size 3kb is not a big deal (assuming it's doing good tree shaking) but still worried about the security / maintenance part

raivieiraadriano92 (approved):
LGTM

cjroth (approved):
Looks great! Love the refactor here.

PR #260, opened 2025-11-23
THU-182: Enabling preview features (eg Tasks) is not updating the state

raivieiraadriano92:
> [!NOTE]
> Fixes settings not updating by invalidating all relevant settings queries; adds injectable fetch to inference/PostHog clients and strengthens PostHog privacy tests.
> 
> - **Settings (frontend)**:
>   - Add `shouldInvalidateSettingsSubset` to target invalidation of all affected `['settings', ...subset]` queries.
>   - Update `useSettings` mutations to use predicate-based `invalidateQueries`.
> - **Inference/Analytics (backend)**:
>   - Allow optional `fetch` injection in `getFireworksClient`, `getThunderboltClient`, and `getInferenceClient` (bypasses cache for testing).
>   - Update `getPostHogClient` to accept optional `fetch`, set `privacy_mode`, and add `clearPostHogClient`; `shutdownPostHog` now supports timeout.
> - **Tests**:
>   - Add/expand PostHog privacy integration tests to ensure no conversation content is sent while metadata is retained.
>   - Add unit tests for `shouldInvalidateSettingsSubset` behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c85a480b07df9d7119ce0c1c4a3fd9b90850f388. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/hooks/use-settings.ts
+++ b/src/hooks/use-settings.ts
@@ -130,11 +130,11 @@ type UseSettingsSchemaResult<T extends SettingSchema, CamelCase extends boolean 
  * This helper checks:
  *   1. Whether the query belongs to the settings subsystem
  *      (queryKey[0] === 'settings')
- *   2. Whether the updated key exists in that query’s subset
+ *   2. Whether the updated key exists in that query's subset
  *
  * If both conditions match, the query should be invalidated.
  */
-function shouldInvalidateSettingsSubset(query: { queryKey: readonly unknown[] }, key: string) {
+export function shouldInvalidateSettingsSubset(query: { queryKey: readonly unknown[] }, key: string) {
   const keys = query.queryKey
 
   // must be a settings query
--- a/src/hooks/use-settings.ts
+++ b/src/hooks/use-settings.ts
@@ -104,6 +104,50 @@ type UseSettingsSchemaResult<T extends SettingSchema, CamelCase extends boolean 
     }
 
 /**
+ * Determines whether a given query should be invalidated after updating
+ * a specific setting key.
+ *
+ * Why this is needed:
+ * -------------------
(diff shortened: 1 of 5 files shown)


src/hooks/use-settings.ts
```diff
@@ -103,6 +103,50 @@ type UseSettingsSchemaResult<T extends SettingSchema, CamelCase extends boolean
...
+ * predicate function during `invalidateQueries` to detect all affected subsets.
+ *
+ * This helper checks:
+ *   1. Whether the query belongs to the settings subsystem
+ *      (queryKey[0] === 'settings')
+ *   2. Whether the updated key exists in that query’s subset
+ *
+ * If both conditions match, the query should be invalidated.
+ */
+function shouldInvalidateSettingsSubset(query: { queryKey: readonly unknown[] }, key: string) {
```
ital0: I liked this! :) 
But do you think it should be called every time that `updateSettings` is used?

cjroth (changes requested):
Seems like a very clean solution, love it! Should we add any unit tests?

src/hooks/use-settings.ts
raivieiraadriano92: I think so - at least I can't think another way to do it.
I mean, basically if we don't invalidate the query the current problem would persist.

Would you have another approach?

raivieiraadriano92:
> Seems like a very clean solution, love it! Should we add any unit tests?

For sure, I forgot that - thanks for reminding me

src/hooks/use-settings.ts
ital0: That's exactly what I had in mind

PR #281, opened 2025-12-02
THU-158: Chat responses should continue streaming after user navigates away from the current thread and then returns to it

raivieiraadriano92:


src/components/chat/chat-messages.tsx
```diff
@@ -14,12 +14,16 @@ type ChatMessagesProps = {
...
-      triggerData: state.triggerData,
-    })),
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
+
+      return {
+        chatInstance: chatItem!.chatInstance!,
+        chatThread: chatItem!.chatThread!,
+        chatThreadId: chatItem!.id!,
+        triggerData: chatItem!.triggerData!,
```
ital0: can we make sure that `triggerData` will be always available? Just confirming because there are other references in this code that are checking its validity.

src/chats/chat-store.ts
```diff
@@ -73,22 +99,31 @@ export const useChatStore = create<ChatStore>()((set, get) => ({
...
+    updatedChats.set(selectedChatId, { ...chat, selectedModel: model })
+
+    set({ chats: updatedChats })
 
     updateSettings({ selected_model: model.id })
 
     trackEvent('model_select', { model: model.id })
   },
-
-  reset: () => {
```
ital0, line 91: is there any other method that would clean up the chats? I think that may cause memory leak

src/components/chat/chat-ui.tsx
```diff
@@ -12,7 +12,15 @@ import { useChat } from '@ai-sdk/react'
 import { useChatAutomation } from '@/chats/use-chat-automation'
 
 export default function ChatUI() {
-  const { chatInstance } = useChatStore(useShallow((state) => ({ chatInstance: state.chatInstance! })))
+  const { chatInstance } = useChatStore(
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
```
ital0: can we guarantee that `selectedChatId` will be always available or is it safer to use null coalesce? Does it make sense to use null coalesce?

src/chats/save-partial-assistant-messages-handler.ts
```diff
@@ -20,10 +20,14 @@ export const SavePartialAssistantMessagesHandler = ({
   useChat = useChat_default,
 }: SavePartialAssistantMessagesHandlerProps) => {
   const { chatInstance, chatThreadId } = useChatStore(
-    useShallow((state) => ({
-      chatInstance: state.chatInstance!,
-      chatThreadId: state.id!,
-    })),
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
```
ital0: can we guarantee that `selectedChatId` will be always available or is it safer to use null coalesce? Does it make sense to use null coalesce?

src/chats/use-chat-automation.tsx
```diff
@@ -9,9 +9,13 @@ type UseChatAutomationProps = {
 
 export const useChatAutomation = ({ useChat = useChat_default }: UseChatAutomationProps = {}) => {
   const { chatInstance } = useChatStore(
-    useShallow((state) => ({
-      chatInstance: state.chatInstance!,
-    })),
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
```
ital0: can we guarantee that `selectedChatId` will be always available or is it safer to use null coalesce? Does it make sense to use null coalesce?

src/chats/use-chat-scroll-handler.ts
```diff
@@ -14,9 +14,13 @@ export const useChatScrollHandler = ({
   useChat = useChat_default,
 }: UseChatScrollHandlerProps = {}) => {
   const { chatInstance } = useChatStore(
-    useShallow((state) => ({
-      chatInstance: state.chatInstance!,
-    })),
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
```
ital0: can we guarantee that `selectedChatId` will be always available or is it safer to use null coalesce? Does it make sense to use null coalesce?

src/hooks/use-handle-integration-completion.ts
```diff
@@ -125,10 +125,14 @@ export const useHandleIntegrationCompletion = ({ saveMessages }: UseHandleIntegr
   const oauthRetryHandledRef = useRef<Set<string>>(new Set())
 
   const { chatInstance, chatThreadId } = useChatStore(
-    useShallow((state) => ({
-      chatInstance: state.chatInstance,
-      chatThreadId: state.id,
-    })),
+    useShallow((state) => {
+      const chatItem = state.chats.get(state.selectedChatId!)
```
ital0: can we guarantee that `selectedChatId` will be always available or is it safer to use null coalesce? Does it make sense to use null coalesce?

PR #293, opened 2025-12-07
THU-158: Chat responses should continue streaming after user navigates away from the current thread and then returns to it

raivieiraadriano92:
This PR introduces support for multiple chat sessions running in the background, allowing users to seamlessly switch between active conversations without interrupting ongoing streams.

It also completes all sub-tasks under THU-158, including:

- THU-206: Some of the conversation steps are stuck in loading state indefinitely even after refreshing page

- THU-209: The last step is always stuck in a loading state while the response renders the text to the user - once the actual text response starts to render, all of the tool calls and reasoning should show as complete (no longer be loading)

- THU-210: Could we show a loading spinner in the side menu of any conversation that is currently streaming where the lock icon is to help the user understand which chats are still loading in the background?

- THU-211: The step title animation (where it appears to rotate up from the bottom on each new step) is re-animating every time I navigate to a chat thread that is streaming in the background, even though that step is already running. This is causing it to feel like it's starting over whenever I navigate to the thread even though it was already running.

---

> [!NOTE]
> Introduce session-scoped chat state enabling concurrent/background streaming, update hydration and components to use sessions, refine reasoning timing/UI, and show streaming status in the sidebar.
> 
> - **Core architecture**:
>   - Replace single chat state with session-scoped store (`sessions`, `currentSessionId`), plus actions: `createSession`, `setCurrentSessionId`, `updateSession`, and session-aware `setSelectedModel(id, modelId)`.
>   - Extract `createChatInstance` and route all fetch/send through session context; `saveMessages` now accepts `{ id, messages }`.
>   - Add `useCurrentChatSession` hook and migrate hooks/components to it.
> - **Hydration/flows**:
>   - `use-hydrate-chat-store`: create/reuse session, set models/MCP clients, and navigate when first saving; uses new `createChatInstance`.
>   - `detail` page simplified (no `isNew`).
> - **UI updates**:
>   - Chat UI, prompt input, messages, scroll/automation handlers updated to session APIs.
>   - Sidebar chat list shows spinner for threads with `status: 'streaming'`.
>   - Reasoning UI: switch to `isGroupReasoning`, adjust loading/animation and completion title.
> - **Telemetry/timing**:
>   - `message-metadata`: end reasoning timing on `text-start` (more accurate than `reasoning-end`).
> - **Tests/utilities**:
>   - Add `test-utils/chat-store-mocks` and refactor tests to new session model and APIs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3e1ec72c670c189f726c0708a1f1eaa86f102483. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/components/ui/header.tsx
```diff
@@ -39,9 +44,9 @@ export const Header = () => {
   const modelSelector = showModelSelector && (
     <ModelSelector
       models={models}
-      selectedModel={selectedModel}
-      chatThread={chatThread}
-      onModelChange={setSelectedModel}
+      selectedModel={selectedModel ?? null}
+      chatThread={chatThread ?? null}
+      onModelChange={(modelId) => setSelectedModel(chatThreadId ?? '', modelId ?? null)}
```
ital0: I understood what this ternary is for… But wouldn’t it be simpler to just run the function under this condition:

```
chatThreadId && setSelectedModel(chatThreadId, modelId ?? null)
```

Unless you need to call this function even when `chatThreadId` is falsy

src/chats/chat-instance.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (chatThread && chatThread.isEncrypted !== selectedModel?.isConfidential) {
+      throw new Error(
+        `This model is not available for ${chatThread.isEncrypted === 1 ? 'encrypted' : 'unencrypted'} conversations.`,
+      )
+    }
+
+    trackEvent('chat_send_prompt', {
+      model: selectedModel,
+      // @ts-ignore
```
ital0: is there any reason for this ts-ignore?

src/chats/chat-store.test.ts
```diff
@@ -79,6 +82,63 @@ const createMockAutomationRun = (overrides?: Partial<AutomationRun>): Automation
   }
 }
 
+/**
+ * Helper to hydrate the store with a session (replaces old hydrate method)
+ */
+const hydrateStore = (state: {
```
ital0: I noticed there are several instances of this method and `resetStore`, `createMockChatInstance` and `createMockUseChat`. 

Can you extract them to a shared test utility? Maybe something in `src/test-utils`

src/chats/save-partial-assistant-messages-handler.test.tsx
```diff
@@ -64,12 +119,14 @@ describe('SavePartialAssistantMessagesHandler', () => {
 
   beforeEach(() => {
     // Reset store state before each test
-    useChatStore.getState().reset()
+    resetStore()
   })
 
   afterEach(async () => {
+    // Cleanup rendered components before resetting store to prevent errors during unmount
+    cleanup()
```
ital0, line 26: can you double-check if this method is not being executed in `src/testing-library.ts`? Turning this in a redudant call.

ital0 (changes requested):
Great approach. And such a big PR :) 

Just some minor adjustments and I think it's ready to go!

src/chats/save-partial-assistant-messages-handler.test.tsx
raivieiraadriano92: In this case we need that because it must cleanup before resetStore() since global afterEach in testing-library.ts runs after this hook.

src/chats/chat-store.test.ts
raivieiraadriano92: it makes total sense - just implemented `src/test-utils/chat-store-mocks.ts`

src/chats/chat-instance.ts
raivieiraadriano92: fixed

src/components/ui/header.tsx
raivieiraadriano92: fixed

ital0 (approved):
LGTM

src/ai/message-metadata.ts
```diff
@@ -45,6 +45,12 @@ export const createMessageMetadata = (modelId: string) => {
         return duration ? { reasoningTime: { [id]: duration } } : { modelId }
       }
 
+      // The AI SDK keeps the reasoning part stream open until the text part stream ends,
+      // even though reasoning part content stops producing output once text part begins.
+      // This means 'reasoning-end' fires late (after text completes), making it
+      // unreliable for timing. We use 'text-start' to capture the actual moment
+      // reasoning finishes and text generation begins.
+      case 'text-start':
```
cjroth, line 53: Have you tested this with different models (eg providers)?

I would not be surprised if this were caused by the provider and not AI SDK - I've found lots of strange idiosyncrasies like this between providers.

cjroth (commented):
I asked Claude Opus to review - here's what it came back with. Feel free to disagree with it - just wanted to share. I tend to agree that we should avoid non-null assertions (`!`).

> I'll review the current branch by examining the key changed files in detail. Let me start by reading the core files involved in this refactor.
> 
> [7 tools called]
> 
> Let me also check a few more files for completeness:
> 
> [3 tools called]
> 
> Now let me run the tests to verify everything is working:
> 
> [3 tools called]
> 
> Excellent! Tests pass and TypeScript compiles cleanly. Here's my code review:
> 
> ---
> 
> # Code Review: Multi-Session Chat Store Refactor
> 
> ## Overview
> 
> This is a well-executed architectural refactor that transforms the chat store from a single-session model to a multi-session architecture. The primary benefit is enabling background chat streaming indicators in the sidebar, but this also sets up better foundations for future multi-chat features.
> 
> ## 👍 What's Good
> 
> ### 1. Clean Architecture
> The session-based design is sensible. Separating `ChatSession` as a type and storing in a `Map` keyed by ID is a natural fit:
> 
> ```8:14:src/chats/chat-store.ts
> type ChatSession = {
>   chatInstance: Chat<ThunderboltUIMessage>
>   chatThread: ChatThread | null // @todo: make required
>   id: string
>   selectedModel: Model
>   triggerData: AutomationRun | null
> }
> ```
> 
> ### 2. Centralized Test Utilities
> Extracting mock creators into `chat-store-mocks.ts` reduces duplication across test files significantly:
> 
> ```154:193:src/test-utils/chat-store-mocks.ts
> export const hydrateStore = (state: {
>   chatInstance: Chat<ThunderboltUIMessage> | null
>   chatThread: ChatThread | null
>   id: string
>   mcpClients?: unknown[]
>   models?: Model[]
>   selectedModel: Model | null
>   triggerData: AutomationRun | null
> }) => {
>   // ...
> }
> ```
> 
> ### 3. Bug Fixes
> The `text-start` handling in `message-metadata.ts` is a good fix for the timing issue with reasoning duration - well commented:
> 
> ```48:52:src/ai/message-metadata.ts
>       // The AI SDK keeps the reasoning part stream open until the text part stream ends,
>       // even though reasoning part content stops producing output once text part begins.
>       // This means 'reasoning-end' fires late (after text completes), making it
>       // unreliable for timing. We use 'text-start' to capture the actual moment
>       // reasoning finishes and text generation begins.
> ```
> 
> ### 4. Naming Improvement
> Renaming `isThinking` → `isGroupReasoning` is clearer about what it represents.
> 
> ---
> 
> ## ⚠️ Issues to Address
> 
> ### 1. **Non-null assertions throughout components**
> 
> Almost every component now has patterns like:
> 
> ```typescript
> const session = state.sessions.get(state.currentSessionId!)!
> ```
> 
> This is fragile. If `currentSessionId` is null or the session doesn't exist, you get cryptic runtime errors. Consider:
> 
> ```typescript
> // Option A: Guard clause with early return
> const { chatInstance, chatThreadId } = useChatStore(
>   useShallow((state) => {
>     const session = state.currentSessionId 
>       ? state.sessions.get(state.currentSessionId) 
>       : undefined
> 
>     return {
>       chatInstance: session?.chatInstance ?? null,
>       chatThreadId: session?.id ?? null,
>     }
>   }),
> )
> 
> if (!chatInstance) return null  // or loading state
> ```
> 
> The `Header` component already does this more safely:
> 
> ```21:21:src/components/ui/header.tsx
>       const session = state.sessions.get(state.currentSessionId ?? '')
> ```
> 
> ### 2. **Potential memory leak with sessions Map**
> 
> Sessions are created but never removed. If a user opens many chats, the `sessions` Map will grow unbounded:
> 
> ```79:93:src/chats/use-hydrate-chat-store.ts
>     // If the session already exists, set the current session id and update the mcp clients and models
>     if (sessions.has(id)) {
>       setCurrentSessionId(id)
>       // ...
>       return
>     }
> ```
> 
> Consider adding:
> - A `deleteSession(id: string)` action
> - LRU eviction (keep last N sessions)
> - Or clear non-current sessions on navigation
> 
> ### 3. **AnimatePresence without keys in ChatListItem**
> 
> The `AnimatePresence` in `ChatListItem` wraps children without unique keys, which can cause animation issues:
> 
> ```62:71:src/layout/sidebar/chat-list-item.tsx
>             <AnimatePresence>
>               <motion.div initial={{ opacity: 0 }} animate={{ opacity: 1 }} exit={{ opacity: 0 }}>
>                 {status === 'streaming' && thread.id !== currentSessionId && (
>                   <Loader2 className={`h-4 w-4 animate-spin text-muted-foreground`} />
>                 )}
>               </motion.div>
>               <motion.div layout>
>                 <span className="truncate flex-1 min-w-0">{thread.title}</span>
>               </motion.div>
>             </AnimatePresence>
> ```
> 
> The spinner `motion.div` should have a `key` when conditionally rendered, and `AnimatePresence` needs a `mode` prop:
> 
> ```tsx
> <AnimatePresence mode="wait">
>   {status === 'streaming' && thread.id !== currentSessionId ? (
>     <motion.div key="spinner" ...>
>       <Loader2 ... />
>     </motion.div>
>   ) : null}
>   <motion.div key="title" layout>
>     <span ...>{thread.title}</span>
>   </motion.div>
> </AnimatePresence>
> ```
> 
> ### 4. **Redundant null check**
> 
> In `chat-instance.ts`:
> 
> ```76:80:src/chats/chat-instance.ts
>     if (!selectedModel) {
>       throw new Error('No selected model')
>     }
> 
>     if (chatThread && chatThread.isEncrypted !== selectedModel?.isConfidential) {
> ```
> 
> Line 80 uses optional chaining on `selectedModel` right after asserting it's not null on line 76. The `?.` is unnecessary.
> 
> ### 5. **Missing eslint-disable comment removal**
> 
> The diff shows `hydrateChatStore` in the useEffect deps:
> 
> ```18:20:src/chats/detail.tsx
>   useEffect(() => {
>     hydrateChatStore()
>   }, [id])
> ```
> 
> If `hydrateChatStore` is recreated on each render (it is - it comes from `useHydrateChatStore`), this could cause re-hydration loops. Verify this is intentional or memoize it.
> 
> ---
> 
> ## 💡 Suggestions
> 
> ### 1. Export `ChatSession` type
> Other parts of the codebase might benefit from typing against `ChatSession`:
> 
> ```typescript
> export type ChatSession = {
>   // ...
> }
> ```
> 
> ### 2. Consider a `useCurrentSession` hook
> To avoid the repetitive pattern:
> 
> ```typescript
> export const useCurrentSession = () => {
>   return useChatStore(
>     useShallow((state) => {
>       const session = state.currentSessionId 
>         ? state.sessions.get(state.currentSessionId)
>         : null
>       return session
>     }),
>   )
> }
> ```
> 
> ### 3. The `@todo: make required` comment
> 
> ```10:10:src/chats/chat-store.ts
>   chatThread: ChatThread | null // @todo: make required
> ```
> 
> Worth tracking this - if it should be required, the session creation flow would need adjustment.
> 
> ---
> 
> ## Summary
> 
> | Category | Assessment |
> |----------|------------|
> | **Architecture** | ✅ Clean, sensible design |
> | **Type Safety** | ⚠️ Heavy reliance on non-null assertions |
> | **Memory Management** | ⚠️ No session cleanup mechanism |
> | **Tests** | ✅ Good coverage, well-organized mocks |
> | **Bug Fixes** | ✅ Reasoning timing fix is solid |
> 
> **Recommendation**: Address the non-null assertion patterns and add session cleanup before merging. The animation keys are a nice-to-have fix.

src/ai/message-metadata.ts
raivieiraadriano92: Yes, I tried DeepSeek-R1 - this seems the way how the SDK handles.

PR #297, opened 2025-12-08
THU-204: Fix Google Drive Issues

ital0:
> [!NOTE]
> Enhances Google Drive and Microsoft OneDrive tools (URL file IDs, Sheets/Slides, structured failure metadata), adds model setting to disable parallel tool calls with runtime handling, and renames nudge messages; includes DB migration and tests/UI updates.
> 
> - **AI runtime**:
>   - Add `experimental_repairToolCall` to skip malformed tool calls; log tool-call details.
>   - Support provider options to disable parallel tool calls when `models.supportsParallelToolCalls=0` (uses `vendor`/`provider`).
>   - Rename `NUDGE_MESSAGES` to `nudgeMessages` across code/tests.
> - **Database & Defaults**:
>   - Add `models.supports_parallel_tool_calls` column with migration `0013_keen_tattoo.sql` and snapshot/journal updates.
>   - Include `supportsParallelToolCalls` in default models, hashing, and new model creation flows (UI sets default `1`).
> - **Google integrations**:
>   - `searchDrive`: enable Shared Drives (`supportsAllDrives`, `includeItemsFromAllDrives`).
>   - New `extractDriveFileId` to accept full Drive/Docs/Sheets/Slides URLs.
>   - `getDriveFileContent`: support Docs/Sheets (CSV)/Slides/text; return structured fields (`mime_type`, `extraction_failed`, `failure_reason`, `file_category`); handle access/not-found; updated tool description and tests.
> - **Microsoft integrations**:
>   - Add OneDrive tools: `microsoft_search_onedrive`, `microsoft_get_onedrive_file_content` (text-only content with structured failure metadata).
> - **Thunderbolt Pro**:
>   - Clarify `fetch_content` description to avoid Drive/OneDrive links.
> - **Migrations meta/lockfile**:
>   - Update migrations registry and snapshots; trivial lockfile change.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 49542a95338262650e7d301d4295fc4ef7d881ce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ital0:
I had to activate this API as well: https://console.cloud.google.com/apis/api/drive.googleapis.com/metrics?project=thunderbolt-sandbox

@cjroth

cjroth (approved):

src/integrations/microsoft/tools.ts
```diff
@@ -136,6 +213,124 @@ export const getMessage = async (params: GetMessageParams, httpClient: KyInstanc
...
+        .text()
+
+      return {
+        file_id: params.file_id,
+        file_name: fileName,
+        mime_type: mimeType,
+        content: textResponse,
+        truncated: false,
+      }
+    }
```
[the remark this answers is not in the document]
ital0: fixed, but I haven't add it back, only removed Google's truncation. 
In my mind, if user asked to summarize something, it doesn't make sense to summarize something that could be truncated. LLM will decide this limit accordingly.

src/ai/fetch.ts
```diff
@@ -198,6 +214,10 @@ export const aiFetchStreamingResponse = async ({
     const maxSteps = 20
     const maxAttempts = 2
 
+    // Mistral models have known issues with parallel tool calls - disable for reliability
+    // See: https://github.com/vllm-project/vllm/issues/9019
+    const providerOptions = isMistralModel ? { mistral: { parallelToolCalls: false } } : undefined
```
cjroth: I have a suggestion here. I think that we should create a new boolean column on the models table called `supports_parallel_tool_calls`, then update that for our default models and use that. This way we're not hard-coding logic that is specific to Mistral. There's a lot of models that don't support parallel tools actually - thinking about how users can add them. I'm not sure if this is a large enough issue to warrant a checkbox on the models page when users add models or not, but it could be.

bun.lock
```diff
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
```
cjroth, line 3: Was this intentional?

src/ai/step-logic.ts
```diff
@@ -65,11 +65,92 @@ export const shouldRetry = (
...
   preventive:
     'You have gathered information from multiple tool calls. Please synthesize the results and provide your response to the user now.',
   retry:
     'You called tools but did not provide a response. Please synthesize all the information you gathered and respond to me now. Do not call any more tools.',
+  toolRefusal:
+    'You have access to all the tools listed in the system prompt. Do not claim you cannot access files, integrations, or services—use the appropriate tool instead. If a user asks about Google Drive, OneDrive, email, or calendar, you have tools for those.',
 } as const
+
+/** Generic refusal patterns that apply regardless of which integrations are enabled */
+const genericRefusalPatterns = [
```
cjroth: I don't think this approach is going to hold up - we're relying on the model using precise terminology, which will be brittle. This would not work if the user is chatting in any language besides English, for example.

src/ai/fetch.ts
```diff
@@ -209,21 +229,39 @@ export const aiFetchStreamingResponse = async ({
...
           if (isFinalStep(steps.length, maxSteps)) {
             console.info(`Final step ${stepNumber} - telling model to wrap it up...`)
             return {
               activeTools: [],
-              messages: [...stepMessages, { role: 'user' as const, content: NUDGE_MESSAGES.finalStep }],
+              messages: [...stepMessages, { role: 'user' as const, content: nudgeMessages.finalStep }],
+            }
+          }
+
+          // Check if previous step contained a tool refusal WITHOUT actually trying tools
```
cjroth: Hmmm. I see what you're trying to accomplish here. Tbh, I'm concerned that we're adding a lot of complexity, including making this logic depend on whether or not google + microsoft are connected, and we're hardwiring google and microsoft references which will later become extensions (thus wanting this logic to stay agnostic of integrations).

Honestly I think that we should try a different approach - I'd rather see if we can update the tool instructions for the microsoft / google tools or the system prompt. Worst case, we could even add a column to the models table that lets us add custom system prompts per model (something that has been on my mind for awhile now).

If we're not able to get Mistral to call tools well despite this, I think it raises a question of whether we actually want to use Mistral. Currently it's still very much an experiment.

Also: are you having problems with mistral-large-3, mistral-medium-3.1, or both? I ask because I noticed that mistral-medium-3.1 actually seemed to do better with tool calling and I wonder if that could be the different in their 3.1 vs 3.0 architecture.

src/ai/fetch.ts
```diff
@@ -254,6 +292,25 @@ export const aiFetchStreamingResponse = async ({
         onError: (error) => {
           console.error('streamText error', error)
         },
+
+        // Handle malformed tool calls from models with weaker tool-calling capabilities
+        experimental_repairToolCall: async ({ toolCall, error }) => {
```
cjroth, line 270: Nice, this should be helpful in debugging stuff I think

src/ai/fetch.ts
```diff
@@ -300,7 +357,7 @@ export const aiFetchStreamingResponse = async ({
               currentMessages = [
                 ...currentMessages,
                 ...response.messages,
-                { role: 'user' as const, content: NUDGE_MESSAGES.retry },
+                { role: 'user' as const, content: nudgeMessages.retry },
```
cjroth, line 333: Hah, thank you :)

src/ai/prompt.ts
```diff
@@ -42,6 +44,12 @@ export const createPrompt = ({ modelName, preferredName, location, localization
     .filter(Boolean)
     .join('\n')
 
+  // Explicitly list available tools to prevent models from hallucinating limitations
```
cjroth: I'm not convinced of this because the list of tools already goes directly into the system prompt when it gets fed into the model. I guess I could imagine a scenario where listing the tools in a single line (as opposed to JSON format) might somehow make it more clear? But honestly, if Mistral is hallucinating tool calls that badly, it might be a sign that it's not a good enough model.

Separately, here's an idea that I think might be more robust. What if in the tool call repair step we could reply to the model itself with a prompt like "That tool does not exist." and maybe even THEN it would make sense to list the tools one more time. (Not sure if that logic would work in the tool call repair step - but something along those lines.)

bun.lock
```diff
@@ -1,6 +1,5 @@
...
   "workspaces": {
```
cjroth, line 3: Was this intentional? I think if you update to the latest bun it might add this back?

cjroth (changes requested):
Nice work! I think there are some really good ideas in here, and some other ideas that I think we should rethink a bit based on my past experiments with prompt engineering.

Lemme know if everything I said makes sense - I'm happy to jump on a call to talk through all of it.

src/ai/fetch.ts
ital0: good catch here! Actually this approach sounds much more dynamic.

bun.lock
ital0: nope, it wasn't. Actually I can't understand why it keeps getting removed. But since it's always `0`, do you think this could case any issues?

src/ai/step-logic.ts
ital0: Totally makes sense. 
I didn’t remember that any language can be used. I’ll try changing the approach.

bun.lock
cjroth: Not sure - I haven't searched what this property is for. I think it got added automatically after updating bun a few weeks ago.

ital0: Well, I researched a little but couldn't understand much lol
I'll revert just in case.

PR #302, opened 2025-12-11
THU-224: Model is not responding in some scenarios after user chooses "Don't ask again" option in Connect Google/Microsoft widget.

ital0:
When users chose "Don't ask again" on the Connect Google/Microsoft widget, the model would sometimes output the widget tag but nothing would render — leaving users with blank responses.

### What was done

- Added integration status to the model's context so it knows when prompts are disabled or integrations are turned off
- The model now asks users if they'd like to connect before showing the widget (when "Don't ask again" was previously selected)
- Added an `override` option that lets the widget display after the user explicitly agrees to connect
- Fixed a bug where retry messages were appearing instead of being hidden (metadata was being overwritten)
- Simplified the widget instructions for better model consistency across different providers

### How it affects users

- Users who selected "Don't ask again" will now see a friendly prompt asking if they'd like to connect, rather than an empty response
- Users with disabled integrations are directed to Settings instead of seeing a broken widget
- The retry flow after OAuth connection now works correctly (hidden messages stay hidden)

---

> [!NOTE]
> Adds integration status to the model context and an override flag to the connect-integration widget, updates instructions/tests, and preserves message metadata to avoid blank or hidden-retry issues.
> 
> - **AI/Backend**
>   - **Settings/Context**: Fetch new integration fields and compute `integrationStatus` (`READY`, `GOOGLE_DISABLED`, `MICROSOFT_DISABLED`, `PROMPTS_DISABLED`); include in system prompt.
>   - **Prompt**: `createPrompt` accepts `integrationStatus` and embeds it in context.
> - **Connect-Integration Widget**
>   - **Schema**: Add required `args.override` (`"true" | ""`).
>   - **Component**: Accept `override` prop; suppress widget when "Do not ask again" is set unless `override="true"`; keep hidden when already connected.
>   - **Instructions**: Rewritten with provider/service selection, status-driven rules, display-only widget rule, and examples.
> - **Parser/Tests**
>   - Update tests to require/validate `override` attribute in `<widget:connect-integration ... />`.
> - **Chats**
>   - **Bug fix**: Preserve existing `message.metadata` when sending (prevents hidden retry messages from reappearing).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 052d221f377a161b84e14f2011b2ce097380960d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/ai/fetch.ts
+++ b/src/ai/fetch.ts
@@ -141,6 +141,11 @@ export const aiFetchStreamingResponse = async ({
     date_format: 'MM/DD/YYYY',
     time_format: '12h',
     currency: 'USD',
+    integrations_do_not_ask_again: false,
+    integrations_google_credentials: '',
+    integrations_google_is_enabled: false,
+    integrations_microsoft_credentials: '',
+    integrations_microsoft_is_enabled: false,
   })
 
   const model = await db.query.modelsTable.findFirst({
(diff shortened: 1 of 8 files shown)


src/ai/fetch.ts
```diff
@@ -166,6 +171,22 @@ export const aiFetchStreamingResponse = async ({
     console.log('Model does not support tools, skipping tool setup')
   }
 
+  // Compute integration status for the model
+  const getIntegrationStatus = (): string => {
+    // Check for disabled integrations (connected but turned off)
+    if (settings.integrationsGoogleCredentials && !settings.integrationsGoogleIsEnabled) {
```
cjroth, line 179: One thing here is that whichever thing returns false first is what it will show for this string. I think we should find a way to return something a little more accurate

cjroth (approved):
Looks good overall. Just the one comment and then I'm good with merging. I haven't tested thoroughly so please test very carefully to make sure everything works as expected before you merge.

PR #307, opened 2025-12-18
THU-191: CR SQLite POC

raivieiraadriano92:


src/db/singleton.ts
```diff
@@ -45,16 +47,28 @@ export class DatabaseSingleton {
...
+      const { CRSQLiteDatabase } = await import('./crsqlite-database')
+      this.#database = new CRSQLiteDatabase()
     }
 
     await this.#database.initialize(path)
     DatabaseSingleton.#initialized = true
+    this.#databaseType = type
 
-    const dbTypeName = type === 'libsql-tauri' ? 'LibSQL for Tauri' : type === 'bun-sqlite' ? 'Bun SQLite' : 'wa-sqlite'
+    const dbTypeName =
```
cjroth, line 64: Maybe we can use a switch statement here to make this more readable

src/db/sync-service.ts
```diff
@@ -0,0 +1,518 @@
...
+  /** If true, the client needs to upgrade to a newer app version */
+  needsUpgrade?: boolean
+  /** The minimum migration version required for sync */
+  requiredVersion?: string
+}
+
+/**
+ * Encode Uint8Array to base64 string
+ */
+const encodeBase64 = (data: Uint8Array): string => {
```
cjroth, line 64: These should probably go in a more universal utils file since they are pretty broadly useful - not just for sync

src/db/sync-service.ts
```diff
@@ -0,0 +1,518 @@
...
+  onError?: (error: Error) => void
+  /** Called when tables have been updated from remote changes */
+  onTablesChanged?: (tables: string[]) => void
+  /** Called when a version mismatch is detected (client needs upgrade) */
+  onVersionMismatch?: (requiredVersion: string) => void
+  /** Called when chat sessions have been updated from remote changes */
+  onChatSessionsChanged?: (chatThreadIds: string[]) => void
+}
+
+export class SyncService {
```
cjroth, line 111: I like how this is constructed as a class to handle state. Nice

PR #319, opened 2026-01-08
Soft Delete Implementation

raivieiraadriano92:
### Overview

This PR implements soft deletes across all major tables in the application. Instead of permanently removing records, deletion now sets a `deletedAt` timestamp while scrubbing nullable columns for privacy.

#### 1. Database Schema Changes

**New columns added** (migration `0014`):
- `deleted_at` column added to: `chat_messages`, `chat_threads`, `mcp_servers`, `tasks`, `triggers`

**Columns made nullable** (migration `0015`):
- All non-PK/FK columns are now nullable to support data scrubbing during soft delete
- This allows clearing sensitive data while preserving the record structure

#### 2. Two-Tier Type System

Introduced a separation between raw database types and application types:

```typescript
// Row types - match nullable database schema
export type ChatMessageRow = InferSelectModel<typeof chatMessagesTable>

// Application types - enforce non-null for previously-required fields
export type ChatMessage = WithRequired<ChatMessageRow, 'content' | 'role'>
```

This provides:
- Type safety at the DAL boundary
- Clear distinction between database rows and validated application data
- Compile-time guarantees that required fields are present in application code

#### 3. DAL Layer Updates

All DAL methods updated to:
- **Filter out soft-deleted records** in all `get*` queries via `isNull(deletedAt)`
- **Perform soft deletes** instead of hard deletes - sets `deletedAt` timestamp
- **Scrub nullable data** for privacy using `clearNullableColumns()` utility
- **Cascade soft deletes** where appropriate (e.g., deleting a model cascades to its prompts and their triggers)
- **Preserve original timestamps** - re-deleting an already-deleted record doesn't update its `deletedAt`

#### 4. New Utility: `clearNullableColumns`

A reusable utility that generates an object with all nullable columns set to `null`:

```typescript
await db.update(usersTable)
  .set({ ...clearNullableColumns(usersTable), deletedAt: Date.now() })
  .where(eq(usersTable.id, userId))
```

Automatically skips:
- Primary key columns (`id`)
- Foreign key columns
- Special columns (`key`, `deletedAt`)

#### 5. Documentation

Added soft delete principle to `AGENTS.md`:
> **Prefer soft deletes over hard deletes** - set a `deletedAt` timestamp instead of removing records. Filter out soft-deleted records in queries.

### Tables Affected

| Table | Soft Delete | Data Scrubbing | Cascade |
|-------|-------------|----------------|---------|
| `chat_threads` | ✅ | ✅ | → messages |
| `chat_messages` | ✅ | ✅ | - |
| `models` | ✅ | ✅ | → prompts → triggers |
| `prompts` | ✅ | ✅ | → triggers |
| `triggers` | ✅ | ✅ | - |
| `mcp_servers` | ✅ | ✅ | - |
| `tasks` | ✅ | ✅ | - |

### Testing

- All 185 DAL tests pass
- Added new tests for:
  - Verifying soft-deleted records are excluded from queries
  - Verifying `deletedAt` timestamp is set correctly
  - Verifying cascade behavior
  - Preserving original `deletedAt` for already-deleted records

---

> [!NOTE]
> Introduces a soft-delete architecture across the app with privacy scrubbing and consistent DAL filtering.
> 
> - DB: Adds `deleted_at` to `chat_messages`, `chat_threads`, `mcp_servers`, `tasks`, `triggers`; makes many columns nullable; adds partial indexes; migration `0014` and metadata updated
> - DAL: All `get*` queries exclude soft-deleted via `isNull(deletedAt)`; replace hard deletes with updates that set `deletedAt` and `clearNullableColumns(...)`; cascades (e.g., `deleteModel` -> prompts -> triggers; `deleteChatThread` -> messages); preserve original `deletedAt`
> - Utils: New `clearNullableColumns(table)` for scrubbing nullable fields; extended `convertUIMessageToDbChatMessage` to include `deletedAt: null`
> - Types: Split raw row types from app types using `WithRequired<>` to enforce non-null fields post-fetch
> - UI/Routes: New `NotFound` component and catch-all routes; redirect to `/not-found` when visiting a soft-deleted chat (`isChatThreadDeleted`)
> - Misc: `ai/fetch` now uses `getModel` DAL; exports updated (including bulk trigger delete); defaults/tests adjusted to cover soft-delete behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 64091618b3867a167d238086b6d7d3503be63d8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/dal/chat-threads.ts
```diff
@@ -81,24 +89,24 @@ export const getContextSizeForThread = async (threadId: string): Promise<number
...
 }
 
 /**
- * Deletes a specific chat thread by ID
+ * Soft deletes a specific chat thread by ID (sets deletedAt timestamp)
  */
 export const deleteChatThread = async (id: string): Promise<void> => {
   const db = DatabaseSingleton.instance.db
-  await db.delete(chatThreadsTable).where(eq(chatThreadsTable.id, id))
+  await db.update(chatThreadsTable).set({ deletedAt: Date.now() }).where(eq(chatThreadsTable.id, id))
```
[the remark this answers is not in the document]
cjroth: This seems like a real concern

cjroth (changes requested):
Two things I think we should change:
1. When we soft delete a thread, it should also soft delete all messages in the thread.
2. When we soft delete a message, it should delete the contents of each message as well (or change to empty string if it can't be null).

cjroth (commented):
This is great!

One more thing. I worry that the `...scrubbedData` pattern to overwrite all columns could result in accidentally omitting future columns that get added if we forget to update here.

Not sure if I have a good elegant solution off the top of my head, but just a thought. Perhaps we could add some reusable util function that ensures all columns are included, or at least we could add some comments to the table definitions reminding us (or LLMs) that this would need to get updated here when new columns get added.

raivieiraadriano92:
> scrubbedData



> This is great!
> 
> One more thing. I worry that the `...scrubbedData` pattern to overwrite all columns could result in accidentally omitting future columns that get added if we forget to update here.
> 
> Not sure if I have a good elegant solution off the top of my head, but just a thought. Perhaps we could add some reusable util function that ensures all columns are included, or at least we could add some comments to the table definitions reminding us (or LLMs) that this would need to get updated here when new columns get added.

agreed - I've just implemented a new method called `clearNullableColumns` which will take the schema and return all nullable columns. In a near future with the sync feature all columns will be nullable so they will automatically be set  while soft deleting.

What do you think?

src/dal/prompts.ts
```diff
@@ -40,7 +40,7 @@ export const getTriggerPromptForThread = async (threadId: string): Promise<Autom
     })
     .from(chatThreadsTable)
     .leftJoin(promptsTable, eq(chatThreadsTable.triggeredBy, promptsTable.id))
-    .where(eq(chatThreadsTable.id, threadId))
+    .where(and(eq(chatThreadsTable.id, threadId), isNull(chatThreadsTable.deletedAt)))
```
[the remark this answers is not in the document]
cjroth: This looks like a real concern

src/dal/prompts.ts
```diff
@@ -85,6 +85,32 @@ export const deleteAutomation = async (id: string): Promise<void> => {
...
+  const promptIds = prompts.map((p) => p.id)
+
+  // Soft-delete all triggers for these prompts in a single query
+  await deleteTriggersForPrompts(promptIds)
+
+  // Soft-delete all prompts for this model
+  await db
+    .update(promptsTable)
+    .set({ deletedAt: Date.now() })
+    .where(and(eq(promptsTable.modelId, modelId), isNull(promptsTable.deletedAt)))
```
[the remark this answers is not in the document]
cjroth: Real concern?

src/dal/chat-threads.ts
```diff
@@ -81,24 +90,45 @@ export const getContextSizeForThread = async (threadId: string): Promise<number
...
-  await db.delete(chatThreadsTable).where(eq(chatThreadsTable.id, id))
+  const deletedAt = Date.now()
+  await db
+    .update(chatMessagesTable)
+    .set({ ...clearNullableColumns(chatMessagesTable), deletedAt })
+    .where(and(eq(chatMessagesTable.chatThreadId, id), isNull(chatMessagesTable.deletedAt)))
+  await db
+    .update(chatThreadsTable)
+    .set({ ...clearNullableColumns(chatThreadsTable), deletedAt })
+    .where(and(eq(chatThreadsTable.id, id), isNull(chatThreadsTable.deletedAt)))
```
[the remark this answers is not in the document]
cjroth: This is a very real issue. I think we need to make these columns nullable (and review all columns in all tables).

cjroth (changes requested):

src/dal/chat-threads.ts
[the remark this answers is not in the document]
raivieiraadriano92: the method `clearNullableColumns` was refactored to handle with all types of columns:

- foreign keys will not be changed
- nullable columns will be set to `null`
- text columns will be set to an empty string `''`
- number columns will be set to `0`

src/dal/prompts.ts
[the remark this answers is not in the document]
raivieiraadriano92: solved

src/dal/chat-messages.ts
```diff
@@ -1,43 +1,49 @@
...
- * Gets a single chat message by ID
+ * Gets a single chat message by ID (excluding soft-deleted)
  */
 export const getMessage = async (messageId: string): Promise<ChatMessage | undefined> => {
   const db = DatabaseSingleton.instance.db
-  return await db.select().from(chatMessagesTable).where(eq(chatMessagesTable.id, messageId)).get()
+  return (await db
+    .select()
+    .from(chatMessagesTable)
+    .where(and(eq(chatMessagesTable.id, messageId), isNull(chatMessagesTable.deletedAt)))
```
ital0, line 16: Noticed that this method is being used in several places... And something I wondered is: does it worth to create a index for this field `deletedAt`? Since database will be filtering by it.

src/dal/models.ts
```diff
@@ -141,11 +142,23 @@ export const resetModelToDefault = async (id: string, defaultModel: Model): Prom
...
 /**
- * Deletes a model by ID
+ * Soft deletes a model by ID (sets deletedAt timestamp)
+ * Also soft-deletes all prompts referencing this model (and their triggers)
+ * Scrubs all non-enum data for privacy
+ * Only updates records that haven't been deleted yet to preserve original deletion timestamps
  */
 export const deleteModel = async (id: string): Promise<void> => {
+  // Import locally to avoid circular dependency
+  const { deletePromptsForModel } = await import('./prompts')
```
ital0, line 148: I think we might run into the same issues again with dynamically importing libraries here. We still don’t know what the side effects of that are.
Would importing this file have a significant impact?

ital0 (changes requested):
In addition to the other points I mentioned in the review, I also noticed that the `getOrCreateChatThread` method may have been impacted by these new changes.

```
The getOrCreateChatThread function will fail with a primary key violation when called with an ID of a soft-deleted thread. Since getChatThread now filters out soft-deleted threads (returns null), but the row still exists in the database, calling createChatThread with that same ID will fail on insert.

Scenario:

User bookmarks a chat URL: /chat/abc-123
User deletes the chat (soft delete)
User navigates to the bookmarked URL
getOrCreateChatThread('abc-123', modelId) is called
getChatThread('abc-123') returns null (soft-deleted)
createChatThread('abc-123', ...) fails with PRIMARY KEY constraint violation
```

src/types.ts
```diff
@@ -30,14 +30,41 @@ export type ThunderboltUIMessage = UIMessage<UIMessageMetadata, UIDataTypes, UIT
...
 export type SaveMessagesFunction = ({ id, messages }: { id: string; messages: ThunderboltUIMessage[] }) => Promise<void>
 
-export type ChatMessage = InferSelectModel<typeof chatMessagesTable>
-export type ChatThread = InferSelectModel<typeof chatThreadsTable>
+/**
+ * Helper type to make specific keys required (non-null).
+ * Used to create application types from row types where certain fields
+ * should be guaranteed to be present for non-deleted records.
+ */
+type WithRequired<T, K extends keyof T> = T & { [P in K]-?: NonNullable<T[P]> }
```
cjroth, line 38: Nice!

src/lib/utils.ts
```diff
@@ -255,3 +257,37 @@ export const truncateText = (text: string, maxLength = 4000): string => {
   if (text.length <= maxLength) return text
   return text.substring(0, maxLength) + '...[truncated]'
 }
+
+/**
+ * Columns that should never be cleared during soft delete
+ */
+const PRESERVE_COLUMNS = new Set(['id', 'key', 'deletedAt'])
```
cjroth: For consistency, I'd make this camel case

cjroth: Actually - I'm not sure that I like treating columns specially based on their name. I know this is probably fine but it opens the possibility of someone getting surprised because they name a column "key" and then wonder why soft-delete doesn't work in the future.

I think using `tableConfig.foreignKeys` to get the actual foreign keys and if there's an equivalent to get primary key, then that's the right way - but I would avoid hardcoding values unless we think that treating certain column names specially is really the way we want to go.

AGENTS.md
```diff
@@ -3,6 +3,7 @@
 - **Bias towards tasteful simplicity** - favor elegant, readable, maintainable solutions that add minimal complexity. Avoid over-engineering, premature optimization, and defensive coding patterns that obscure intent.
 - **Always implement proper, architectural solutions** - no shortcuts, hacky fixes, or temporary workarounds. Research best practices when needed.
 - **Prefer optimistic code over defensive code** - let errors surface loudly during development rather than wrapping everything in if-checks and try/catch blocks. Handle errors architecturally at higher levels (e.g., error handling middleware).
+- **Prefer soft deletes over hard deletes** - set a `deletedAt` timestamp instead of removing records. Filter out soft-deleted records in queries.
```
cjroth: I think we could go a step further and say that we must always use soft delete

cjroth (changes requested):
Looks good overall - honestly it's a fairly simple and straightforward PR with a lot of changes needed across the dal functions and migrations which makes sense. Other than the minor things I mentioned, I think we're almost ready to merge.

src/dal/models.ts
raivieiraadriano92: For this specific case we need this dynamic import due the circular dependency - the `prompts.ts` is also importing the `models.ts`

About the other dynamic import in this file - they were remove, they weren't actually necessary.

src/dal/chat-messages.ts
raivieiraadriano92: I'm not sure - I think this implementation is already really simple and it's using the standard drizzle utility functions.
What would be your suggested implementation?

ital0: That would be at database level, not code level.

```sql
CREATE INDEX idx_chat_threads_deleted_at ON chat_threads (deleted_at);
CREATE INDEX idx_chat_messages_deleted_at ON chat_messages (deleted_at);
CREATE INDEX idx_models_deleted_at ON models (deleted_at);
CREATE INDEX idx_prompts_deleted_at ON prompts (deleted_at);
CREATE INDEX idx_triggers_deleted_at ON triggers (deleted_at);
CREATE INDEX idx_tasks_deleted_at ON tasks (deleted_at);
CREATE INDEX idx_mcp_servers_deleted_at ON mcp_servers (deleted_at);
```

raivieiraadriano92: done

raivieiraadriano92:
> In addition to the other points I mentioned in the review, I also noticed that the `getOrCreateChatThread` method may have been impacted by these new changes.
> 
> ```
> The getOrCreateChatThread function will fail with a primary key violation when called with an ID of a soft-deleted thread. Since getChatThread now filters out soft-deleted threads (returns null), but the row still exists in the database, calling createChatThread with that same ID will fail on insert.
> 
> Scenario:
> 
> User bookmarks a chat URL: /chat/abc-123
> User deletes the chat (soft delete)
> User navigates to the bookmarked URL
> getOrCreateChatThread('abc-123', modelId) is called
> getChatThread('abc-123') returns null (soft-deleted)
> createChatThread('abc-123', ...) fails with PRIMARY KEY constraint violation
> ```

good one, it was added an verification step where if the chat is deleted the user will be redirected to the `/chat/new` page and a toast will be shown

AGENTS.md
raivieiraadriano92: done

src/lib/utils.ts
raivieiraadriano92: good one - done

ital0 (approved):

cjroth (approved):
Looks good

PR #322, opened 2026-01-12
THU-221: Authentication not working on mobile

ital0:
### Problem

On iOS, users could complete the sign-in flow successfully but weren't actually logged in after closing the modal. This happened because iOS WKWebView doesn't persist cookies for the `tauri://localhost` origin used by Tauri apps.

### Solution

Implemented bearer token authentication for mobile platforms. Instead of relying on cookies (which work on desktop), the app now:

- Stores the session token locally after successful sign-in
- Sends it via `Authorization: Bearer` header on subsequent requests
- Restores the token on app startup for persistent sessions

This approach follows Better Auth's recommended pattern for mobile apps where cookie-based sessions aren't reliable.

### Impact

iOS and Android users can now sign in and stay authenticated across app sessions.

---

> [!NOTE]
> Introduces bearer-token–based auth for reliable mobile sessions and unifies auth across platforms.
> 
> - Backend: enables `bearer()` plugin in `backend/src/auth/auth.ts` alongside `emailOTP`
> - Client: routes all requests with `Authorization: Bearer <token>` via `fetchOptions.auth` in `src/contexts/auth-context.tsx`; captures `set-auth-token` response header to persist
> - New token store: adds `src/lib/auth-token.ts` to cache/persist tokens in settings; supports `get/set/load/clear`
> - App init: loads stored token during startup in `src/hooks/use-app-initialization.ts`
> - Sign-in: saves returned token after OTP verification in `src/components/use-sign-in-modal-state.ts`
> - Logout: clears persisted token in `src/components/logout-modal.tsx`
> - Tests: adds `src/lib/auth-token.test.ts` and updates `logout-modal.test.tsx` for DB setup/reset
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d385dbb4106b01bb4108c47b4b8e2e6cf1023aa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/use-sign-in-modal-state.ts
+++ b/src/components/use-sign-in-modal-state.ts
@@ -1,4 +1,6 @@
 import type { AuthClient } from '@/contexts'
+import { setAuthToken } from '@/lib/auth-token'
+import { isMobile } from '@/lib/platform'
 import { useReducer, type FormEvent } from 'react'
 
 type ModalStatus = 'idle' | 'sending' | 'sent' | 'verifying' | 'success' | 'error'
@@ -109,6 +111,11 @@ export const useSignInModalState = ({ authClient, onClose }: UseSignInModalState
       if (result.error) {
         dispatch({ type: 'VERIFY_ERROR', payload: result.error.message || 'Invalid code. Please try again.' })
         return
+      }
+
+      // On mobile, store the token for bearer auth (cookies don't persist)
+      if (isMobile() && result.data?.token) {
+        await setAuthToken(result.data.token)
       }
 
       // Sign-in successful - show success state
--- a/src/components/use-sign-in-modal-state.ts
+++ b/src/components/use-sign-in-modal-state.ts
@@ -1,6 +1,5 @@
 import type { AuthClient } from '@/contexts'
 import { setAuthToken } from '@/lib/auth-token'
-import { isMobile } from '@/lib/platform'
 import { useReducer, type FormEvent } from 'react'
 
 type ModalStatus = 'idle' | 'sending' | 'sent' | 'verifying' | 'success' | 'error'
@@ -113,8 +112,8 @@ export const useSignInModalState = ({ authClient, onClose }: UseSignInModalState
         return
       }
 
-      // On mobile, store the token for bearer auth (cookies don't persist)
-      if (isMobile() && result.data?.token) {
+      // Store the token for bearer auth
+      if (result.data?.token) {
         await setAuthToken(result.data.token)
       }
 
--- a/src/lib/auth-token.ts
+++ b/src/lib/auth-token.ts
@@ -8,6 +8,7 @@
  * while persisting to settings database for durability across app restarts.
  */
 
+import { deleteSetting, getSettings, updateSettings } from '@/dal/settings'
 import { isMobile } from './platform'
 
 const AUTH_TOKEN_SETTING_KEY = 'auth_bearer_token'
@@ -24,7 +25,6 @@ export const getAuthToken = (): string | null => {
 export const setAuthToken = async (token: string | null): Promise<void> => {
   cachedToken = token
 
-  const { updateSettings, deleteSetting } = await import('@/dal/settings')
   if (token) {
     await updateSettings({ [AUTH_TOKEN_SETTING_KEY]: token })
   } else {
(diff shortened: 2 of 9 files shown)


src/lib/auth-token.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!isMobile()) return
+
+  const settings = await getSettings({ [AUTH_TOKEN_SETTING_KEY]: String })
+  cachedToken = settings.authBearerToken
+}
+
+/** Clear the auth token (for sign-out) */
+export const clearAuthToken = async (): Promise<void> => {
+  await setAuthToken(null)
+}
```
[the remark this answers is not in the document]
ital0: fixed

src/components/use-sign-in-modal-state.ts
```diff
@@ -111,6 +113,11 @@ export const useSignInModalState = ({ authClient, onClose }: UseSignInModalState
         return
       }
 
+      // On mobile, store the token for bearer auth (cookies don't persist)
```
cjroth: If we go with storing token in our settings table - I think we should do it on all platforms - this way there's not a separate mechanism happening between mobile and web

src/lib/auth-token.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Provides sync access (required by Better Auth's fetchOptions.auth.token)
+ * while persisting to settings database for durability across app restarts.
+ */
+
+import { deleteSetting, getSettings, updateSettings } from '@/dal/settings'
+import { isMobile } from './platform'
+
+const AUTH_TOKEN_SETTING_KEY = 'auth_bearer_token'
+
+let cachedToken: string | null = null
```
cjroth, line 16: If it makes sense to cache this, we could use our `memoize` function like we do here: https://github.com/thunderbird/thunderbolt/blob/30ad226dda482571c12b042130cf7d78a916e5c4/src/lib/tool-metadata.ts#L23

This would avoid needing to re-implement the pattern and gives us a nice tested way to clear the cache for unit tests

cjroth: Also wondering if the extra millisecond or two that this would save is worth the complexity - sqlite lookups are super fast anyways

cjroth (changes requested):
**Updated:**

It seems like the more correct fix would be to figure out why WKWebView doesn't persist cookies for the tauri://localhost origin wouldn't it? I'm cautious about changing our approach unless there's a good reason why that actually wouldn't work on mobile.

But, if cookies are truly not supported by Webviews: I think this makes sense, and we're already treating the settings table as secure (stores API keys and such anyways).

I think this approach basically makes sense as long as there are no security issues - I haven't fully thought through that, but I can't think of any obvious problem with this approach.

The main feedback I'd have here is, if we're going to use the settings table / bearer approach for mobile - let's just use it universally for desktop/web as well then. This way we don't have to have code branching with edge cases where something works differently in different environments.

ital0:
> **Updated:**
> 
> It seems like the more correct fix would be to figure out why WKWebView doesn't persist cookies for the tauri://localhost origin wouldn't it? I'm cautious about changing our approach unless there's a good reason why that actually wouldn't work on mobile.
> 
> But, if cookies are truly not supported by Webviews: I think this makes sense, and we're already treating the settings table as secure (stores API keys and such anyways).
> 
> I think this approach basically makes sense as long as there are no security issues - I haven't fully thought through that, but I can't think of any obvious problem with this approach.
> 
> The main feedback I'd have here is, if we're going to use the settings table / bearer approach for mobile - let's just use it universally for desktop/web as well then. This way we don't have to have code branching with edge cases where something works differently in different environments.

@cjroth I agree!
Actually, my first approach was to investigate why `WKWebView` doesn’t support storing this type of cookie, since that would make it possible to solve the whole problem directly on the frontend. However, I ran into these threads:

- https://github.com/tauri-apps/tauri/issues/2604
- https://github.com/orgs/tauri-apps/discussions/5337
- https://github.com/tauri-apps/wry/issues/1587
- https://github.com/tauri-apps/wry/discussions/1198

In short, as pointed out in the ticket, the issue only affects iOS, since it uses Safari under the hood.
Safari enforces some stricter policies around custom schemes (as is the case with Tauri apps and localhost).

As for using Bearer Tokens across all platforms, it’s feasible and much more reliable, since we’d have a single implementation for everything.

Regarding security, ideally I should have used some kind of secure storage for mobile devices, but I chose to reduce complexity and rely on the security the OS provides by encrypting the app’s internal data. Still, this would definitely be the best approach.

cjroth:
Cool. So just documenting here that we'll proceed with using the settings table like any other setting value for now.<br><br>I created [THU-246](https://linear.app/mozilla-thunderbolt/issue/THU-246/research-sqlite-security-model) to describe more for the future task

cjroth (approved):
looks good!

PR #328, opened 2026-01-14
Raivieiraadriano92/thu 243 build cr sqlite sync poc frontend

raivieiraadriano92:
> [!NOTE]
> Implements an offline-first CRDT-backed database and real-time sync.
> 
> - Switch DB to `crsqlite` (wasm) with worker client (`crsqlite-worker*`, `CRSQLiteDatabase`) and `DatabaseSingleton` updates; add row normalization helpers
> - Add WebSocket sync service (`db/sync-service.ts`) with initial HTTP push/pull (`db/initial-sync.ts`), migration version checks, and settings-based endpoint; wire up `use-sync-service` hook
> - Extend migrations and schema: partial indexes, CRR initialization (`initializeCRRs`), and migration handling around `ALTER TABLE`
> - Update app init flow to capture/push local changes across migrations and perform initial sync; add `SITE_ID`/version tracking in `sync-utils`
> - Chat store: add `recreateSession(id, saveMessages)` and `use-save-messages`; hydrate uses new factory; tests for chat-store and sync serialization
> - UI: new `SyncStatusIndicator` and integrate into `Header`
> - Auth: temporary mock client in `auth-context` for CORS during sync development
> - Settings: fix `reconcileDefaults` to avoid overwriting user values with null defaults; tests added
> - Tooling: copy `crsqlite.wasm` in Vite, ignore in git; add `@vlcn.io/*` deps
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 144cff96fc6ebb61bbe8c0860fb62df4fa0c5072. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


package.json
```diff
@@ -31,7 +31,8 @@
     "check": "bun run type-check && bun run lint && bun run format-check",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
-    "prepare": "husky"
+    "prepare": "husky",
+    "postinstall": "cp node_modules/@vlcn.io/crsqlite-wasm/dist/crsqlite.wasm public/crsqlite.wasm"
```
cjroth: would it make sense to have vite do this as a build step? I don't think this is wrong, but i do think postinstall is sometimes blocked by IDEs / ai agents for security reasons

src/lib/reconcile-defaults.ts
```diff
@@ -28,15 +28,19 @@ export const reconcileDefaultsForTable = async <T extends { defaultHash: string
     const keyValue = (defaultItem as any)[keyField]
     const existing = await db.select().from(table).where(eq(table[keyField], keyValue)).get()
 
-    if (!existing) {
+    // Drizzle sqlite-proxy may return an object with undefined values instead of undefined
```
cjroth: I experienced inconsistencies between libsql and other versions of sqlite like this previously. I felt like the best way to fix it was in the sqlite proxy adapter so that we can guarantee consistency in the application layer code.

Would that be possible or make sense in this case?

src/hooks/use-sync-service.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { DatabaseSingleton } from '@/db/singleton'
+import { useSettings } from '@/hooks/use-settings'
+import { useQueryClient } from '@tanstack/react-query'
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+/**
+ * Mapping from database table names to React Query keys that should be invalidated
+ * when that table changes from a sync operation
+ */
+const TABLE_TO_QUERY_KEYS: Record<string, string[][]> = {
```
cjroth: Should use camel case for const name for consistency here

ital0 (dismissed):
--

src/hooks/use-app-initialization.ts
```diff
@@ -18,6 +19,139 @@ import ky from 'ky'
 import type { PostHog } from 'posthog-js'
 import { useCallback, useEffect, useState } from 'react'
 
+// Default cloud URL - used for initial sync before settings are loaded
+const DEFAULT_CLOUD_URL = import.meta.env.VITE_THUNDERBOLT_CLOUD_URL || 'http://localhost:8000/v1'
```
cjroth, line 24: Do you think it would be possible to re-order the initial initialization logic to fetch the settings from the db, then run this code, then run migrations afterwards in order to ensure that we're using the value from the settings table?

I think this would prevent a custom cloud url in the settings from working at all

src/lib/platform.ts
```diff
@@ -93,31 +93,51 @@ const fetchCapabilities = memoize(async (): Promise<Capabilities> => {
...
+ * - Consistent sync support across web browsers and desktop apps
+ * - Same codebase, no platform-specific database logic
+ * - CRDT automatically resolves conflicts when syncing between devices
  */
 export const getDatabaseType = async (): Promise<DatabaseType> => {
-  if (!isTauri()) return 'wa-sqlite'
-
-  const { libsql } = await getCapabilities()
-  return libsql ? 'libsql-tauri' : 'wa-sqlite'
+  // Always use cr-sqlite for CRDT-based multi-device sync on all platforms
```
cjroth, line 114: I think this makes sense - and using the same db across all platforms actually makes things more consistent.

That said, I think ideally we could still use some sort of disk-based sqlite database like libsql in development so that we can open the database in a sqlite viewer.

Can you think of (a) a way to open this in a real db viewer so we can inspect the database or (b) a way to store still use libsql locally in development with syncing disabled?

src/db/singleton.ts
```diff
@@ -1,12 +1,14 @@
+import type { CRSQLiteDatabase } from './crsqlite-database'
```
cjroth, line 1: This PR might not be the place to do it, but I'd be in favor of rearchitecting our `Database.instance.db` stuff to use React context, hooks, and injection. I guess this class would stay the same but we'd access the instance differently and stop calling it a "singleton".

package.json
cjroth: Separately, we probably don't want to check it into git if we're also copying it here or at build time

cjroth: (though checking into git is not wrong either - I don't have a strong opinion here - the main downside of checking into git is that we'd have to remember to copy it whenever we update the version if they release new versions)

cjroth (changes requested):
Should remove pnpm-lock.yaml.

This is amazing work. I'm really excited for it.

src/chats/use-hydrate-chat-store.ts
```diff
@@ -5,88 +5,30 @@ import {
...
-      navigate(`/chats/${id}`, { relative: 'path' })
-    }
-  }
+  const saveMessages = createSaveMessages()
 
   const hydrateChatStore = async () => {
     const { createSession, sessions, setCurrentSessionId, setMcpClients, setModels } = useChatStore.getState()
 
-    // Check if this ID belongs to a deleted chat - redirect to 404 if so
-    const isDeleted = await isChatThreadDeleted(id)
```
ital0, line 84: I see this has been removed. What happens if the user tries to navigate to a deleted chat?

src/db/initial-sync.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  setServerVersion(BigInt(response.serverVersion))
+}
+
+/**
+ * Perform initial sync (push + pull) via HTTP
+ * Used during app initialization before WebSocket is established
+ */
+export const performInitialSync = async (httpClient: KyInstance): Promise<void> => {
```
ital0, line 122: Does it worth to add a retry policy for this? Just thinking about the case of any failure during the initial sync.

src/chats/use-hydrate-chat-store.ts
raivieiraadriano92: good catch - it was a merging conflict

src/contexts/auth-context.tsx
```diff
@@ -58,7 +101,9 @@ export const AuthProvider = ({ children, authClient: overrideClient }: AuthProvi
       return null
     }
 
-    const client = createAuthClientInstance(cloudUrl.value)
+    // TODO: Replace with real auth client once CORS is resolved
+    // const client = createAuthClientInstance(cloudUrl.value)
+    const client = createMockAuthClient()
```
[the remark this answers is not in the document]
raivieiraadriano92: it's fine for now - it will be replaced before merging into `main`

src/db/singleton.ts
raivieiraadriano92: Just created a new ticket to handle this one separately: https://linear.app/mozilla-thunderbolt/issue/THU-249/rearchitect-databaseinstancedb

src/hooks/use-sync-service.tsx
raivieiraadriano92: done

package.json
raivieiraadriano92: good catch - I didn't think about it before but it makes more sense to copy this at build time

src/lib/platform.ts
raivieiraadriano92: we could add a if statement for use libsql in Tauri Development Builds - np

another option, maybe, could be implement a dev-only feature to export the DB .sqlite so we can open it in the viewer - this would work in all platforms

what do you think?

src/hooks/use-app-initialization.ts
raivieiraadriano92: good catch - it's fixed

src/db/sync-service.ts
```diff
@@ -0,0 +1,564 @@
...
+  /** Called when a version mismatch is detected (client needs upgrade) */
+  onVersionMismatch?: (requiredVersion: string) => void
+  /** Called when chat sessions have been updated from remote changes */
+  onChatSessionsChanged?: (chatThreadIds: string[]) => void
+}
+
+/**
+ * Encode Uint8Array to base64 string
+ */
+const encodeBase64 = (data: Uint8Array): string => {
```
ital0: minor: I noticed that the base64 functions (encode and decode) are duplicated here and in `src/db/initial-sync.ts`

src/db/sync-service.ts
```diff
@@ -0,0 +1,564 @@
...
+  for (let i = 0; i < binary.length; i++) {
+    bytes[i] = binary.charCodeAt(i)
+  }
+  return bytes
+}
+
+/**
+ * Serialize a CRSQLChange for network transport
+ */
+const serializeChange = (change: CRSQLChange): SerializedChange => ({
```
ital0: same, we have this function duplicated here and in `initial-sync.ts`

ital0 (changes requested):
good job here!

src/db/sync-service.ts
raivieiraadriano92: good catch - it's fixed

src/db/sync-service.ts
raivieiraadriano92: it's fixed

PR #330, opened 2026-01-15
THU-234: Switching theme mode causes crash

darkbanjo:
> [!NOTE]
> Addresses a crash when rapidly toggling themes by hardening selection handling and persisted state.
> 
> - Add guard in `ThemeToggle` `onValueChange` to ignore falsy values before calling `setTheme` and tracking events
> - In `theme-provider`, introduce `isValidTheme` and initialize state from a validated `savedTheme` instead of casting; continue persisting to `localStorage`
> - Add `ThemeToggle` tests covering rendering, selection behavior, and prevention of deselection-triggered updates
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44f2436daaaaa131a90a570fd9cb445e1c0a6969. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


raivieiraadriano92 (approved):
LGTM

cjroth (approved):
I'm curious what is causing the value to get set to the invalid value (null/empty string or whatever it was) - in theory, I'd rather see the root cause fixed by preventing that. But if the cause is more complicated than that, I think this is fine.

PR #334, opened 2026-01-15
THU-245: Changing a location alert is broken

ital0:
> [!NOTE]
> **Fix: preserve user-set values during defaults reconciliation**
> 
> - In `reconcile-defaults.ts`, skip updating when `existing.value` is non-null and `defaultItem.value` is null (even if hashes match), protecting localization and similar settings; still sets `defaultHash` when missing and updates when both values are null.
> - Tests in `reconcile-defaults.test.ts` cover preserving recomputeHash-set values vs. null defaults and the both-null case.
> - In `preferences.tsx`, declining localization changes now only closes the dialog (no hash-only updates); dialog buttons refined (cancel without handler, action `autoFocus`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2a633a83b5c286c16e04e6ab945521fdb2e86292. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/lib/reconcile-defaults.test.ts
+++ b/src/lib/reconcile-defaults.test.ts
@@ -308,4 +308,79 @@ describe('reconcileDefaultsForTable', () => {
     // Value should be preserved
     expect(setting?.value).toBe('some_value')
   })
+
+  test('preserves user values set via recomputeHash when code default is null', async () => {
+    const db = DatabaseSingleton.instance.db
+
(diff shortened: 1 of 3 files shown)


raivieiraadriano92 (approved):
LGTM

cjroth (approved):
:shipit:

PR #341, opened 2026-01-21
THU-242: Waitlist flow - Backend

ital0:
> [!NOTE]
> Implements a waitlist system and integrates it with passwordless auth, plus centralizes email handling and normalization.
> 
> - **DB:** Adds `waitlist` table with unique `email`, `status` (pending/approved), `batch_id`, timestamps, and indexes
> - **API:** New `POST /v1/waitlist/join` (email normalized, idempotent; sends joined/reminder emails; prevents enumeration) and mounts routes in app
> - **Auth:** Normalizes emails on user create via `databaseHooks`; gates OTP delivery by waitlist status (existing users bypass); sends "not ready" email for pending users; verify URL uses normalized email
> - **Email:** Refactors sending via `lib/resend` (`sendEmail`, `shouldSkipEmail`) and updates `sendSignInEmail` to use it
> - **Utils/Tests:** Adds `normalizeEmail`; comprehensive tests for waitlist API, auth integration, normalization, and email failure handling
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6828f7a9339ae74ea905b15396f90a9ebddd981c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


backend/src/auth/auth.ts
```diff
@@ -47,6 +50,29 @@ export const createAuth = (database: typeof DbType) =>
...
+            const waitlistEntry = await database
+              .select({ status: waitlist.status })
+              .from(waitlist)
+              .where(and(eq(waitlist.email, email.toLowerCase()), isNull(waitlist.deletedAt)))
+              .limit(1)
+
+            // If not on waitlist or not approved, don't send OTP
+            // Return silently to not reveal waitlist status
+            if (waitlistEntry.length === 0 || waitlistEntry[0].status !== 'approved') {
+              console.info(`🚫 Blocked sign-in attempt for non-approved email: ${email}`)
```
cjroth: I'd not log the email to prevent PII from going into our logs

backend/src/auth/auth.ts
```diff
@@ -47,6 +50,29 @@ export const createAuth = (database: typeof DbType) =>
...
+          // If user doesn't exist, check waitlist status
+          if (existingUser.length === 0) {
+            const waitlistEntry = await database
+              .select({ status: waitlist.status })
+              .from(waitlist)
+              .where(and(eq(waitlist.email, email.toLowerCase()), isNull(waitlist.deletedAt)))
+              .limit(1)
+
+            // If not on waitlist or not approved, don't send OTP
+            // Return silently to not reveal waitlist status
```
cjroth: It could be nice to send an email to them saying "sorry we're not quite ready for you yet, but you're on the waitlist" or something here - maybe better than a completely silent rejection from the user's perspective

backend/src/waitlist/README.md
```diff
@@ -0,0 +1,111 @@
+# Waitlist Module
```
cjroth, line 1: Interesting idea to put a readme here. Do you think we should put this in docs? I'm fine with either, just wanted to make sure we thought about it.

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,106 @@
+import { Resend } from 'resend'
+
+const resend = process.env.RESEND_API_KEY ? new Resend(process.env.RESEND_API_KEY) : null
```
cjroth: We should create the Resend instance somewhere (maybe utils?) and export the instance so that we don't end up creating multiple new instances when the backend server boots. And just cleaner if we ever want to add config settings or anything. (And we should update anywhere else in the backend that we're using it already.)

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Send waitlist confirmation email
+ * Uses a simple template saying "Thanks for signing up! We'll let you know when it's ready."
+ */
+export const sendWaitlistConfirmationEmail = async ({
+  email,
+  isProduction,
+}: SendWaitlistConfirmationParams): Promise<void> => {
+  console.info(`📧 Sending waitlist confirmation to ${email}`)
+
+  // Skip email in test/dev mode or if resend is not configured
```
cjroth: I'd love to see this boolean logic turned into a reusable function - maybe defined in suggested resend file - so that we can call it anywhere needed without redefining.

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+      throw new Error('Email service not configured')
+    }
+    console.info(`📝 [DEV] Would send waitlist confirmation to: ${email}`)
+    return
+  }
+
+  const { data, error } = await resend.emails.send({
+    from: 'hello@auth.thunderbolt.io',
+    to: email,
+    subject: "You're on the Thunderbolt waitlist!",
```
cjroth: We should define the email template in resend instead of having to hard code html here. I'm guessing we'll have a few emails as part of the waitlist (and maybe other future features) that are essentially just sending the user a message like this maybe with a link. Perhaps we could create some kind of very generic email template that just has a variable for body text and optional link and link text?

cjroth: (Take a look at the magic link auth feature for an example)

cjroth (changes requested):

backend/src/waitlist/README.md
ital0: @cjroth since this is context for a very specific feature, I think it makes sense to keep it here.
Inside docs, we can keep more “generic” things, for example `testing.md,` which provides general testing context, or `webview.md`, which explains web-related details, etc.

backend/src/auth/auth.ts
```diff
@@ -47,6 +45,41 @@ export const createAuth = (database: typeof DbType) =>
...
           }
 
+          const normalizedEmail = email.toLowerCase().trim()
+
+          // Check if user already has an account (existing users bypass waitlist)
+          const existingUser = await database
+            .select({ id: user.id })
+            .from(user)
+            .where(eq(user.email, normalizedEmail))
+            .limit(1)
```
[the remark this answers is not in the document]
ital0: the change I applied will prevent that for production cases

backend/src/lib/resend.ts
```diff
@@ -0,0 +1,26 @@
...
+if (!resend) {
+  console.warn('⚠️ RESEND_API_KEY is not set - emails will not be sent')
+}
+
+/**
+ * Check if email sending should be skipped (test/dev mode or resend not configured)
+ * Throws an error if in production but resend is not configured
+ * @returns true if email should be skipped, false if it should be sent
+ */
+export const shouldSkipEmail = (isProduction: boolean): boolean => {
```
cjroth: Could we move the `=== production` here so we don't have to pass it in every time?

cjroth: Also: what do you think of creating a function that sends the email here, this way the parent function that calls it doesn't even need to remember to check. It would just console.log instead of sending if not in prod.

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+export const sendWaitlistNotReadyEmail = async ({ email, isProduction }: SendWaitlistEmailParams): Promise<void> => {
+  console.info('📧 Sending waitlist not-ready email')
+
+  if (shouldSkipEmail(isProduction)) {
+    console.info('📝 [DEV] Would send waitlist not-ready email')
+    return
+  }
+
+  const { data, error } = await resend!.emails.send({
+    from: 'hello@auth.thunderbolt.io',
```
cjroth: I think it would be smart to put this from address into resend file so that we can change in one place

backend/src/auth/auth.ts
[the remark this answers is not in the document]
cjroth: Can we confirm that we normalize emails when user signs up?

backend/src/auth/auth.ts
```diff
@@ -47,11 +44,45 @@ export const createAuth = (database: typeof DbType) =>
...
+            // If not on waitlist or not approved, don't send OTP
+            if (waitlistEntry.length === 0 || waitlistEntry[0].status !== 'approved') {
+              console.info('🚫 Blocked sign-in attempt for non-approved email')
+
+              // If on waitlist but not approved, send a "not ready yet" email
+              if (waitlistEntry.length > 0) {
+                sendWaitlistNotReadyEmail({
+                  email: normalizedEmail,
+                  isProduction: process.env.NODE_ENV === 'production',
+                }).catch((error) => {
```
cjroth: Should we just await and let it throw here? (Just cleaner/simpler than catching and console.erroring I think)

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+    to: email,
+    template: {
+      id: 'waitlist-reminder',
+      variables: {},
+    },
+  })
+
+  if (error) {
```
cjroth: Same here, I'd just let it throw

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+  const { data, error } = await resend!.emails.send({
+    from: 'hello@auth.thunderbolt.io',
+    to: email,
+    template: {
+      id: 'waitlist-confirmation',
+      variables: {},
+    },
+  })
+
+  if (error) {
```
cjroth: And here

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,119 @@
...
+            email,
+            isProduction: process.env.NODE_ENV === 'production',
+          }).catch((error) => {
+            console.error('Failed to send waitlist reminder email:', error)
+          })
+
+          return { success: true }
+        }
+
+        // Check if email was soft-deleted (can be reactivated)
```
cjroth: What's the use case where we would need to soft delete an email? Personally I feel like we don't need this

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,119 @@
...
+            })
+          } catch (error) {
+            // Handle race condition: concurrent request already inserted this email
+            // PostgreSQL unique constraint violation code is '23505'
+            const isUniqueViolation =
+              error instanceof Error && 'code' in error && (error as Error & { code: string }).code === '23505'
+
+            if (isUniqueViolation) {
+              // Another request already inserted - send reminder instead
+              sendWaitlistReminderEmail({
```
cjroth: Should be awaited (to ensure that a promise rejection is thrown properly)

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,119 @@
...
+
+              return { success: true }
+            }
+
+            throw error
+          }
+        }
+
+        // Send confirmation email (fire and forget - don't block response)
+        sendWaitlistConfirmationEmail({
```
cjroth: Should be awaited

backend/src/lib/resend.ts
cjroth: One nice thing about that also is that we won't need to use `!` to reference the resend instance anymore, which I feel is a bit messy

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,119 @@
+import type { db } from '@/db/client'
+import { waitlist } from '@/db/schema'
+import { and, eq, isNotNull, isNull } from 'drizzle-orm'
+import { Elysia, t } from 'elysia'
+import { sendWaitlistConfirmationEmail, sendWaitlistReminderEmail } from './utils'
+
+export const createWaitlistRoutes = (database: typeof db) => {
+  return new Elysia({ prefix: '/waitlist' })
```
cjroth: Wouldn't this endpoint allow someone to check what emails are on the list?

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            throw error
+          }
+        }
+
+        // Send confirmation email (fire and forget - don't block response)
+        sendWaitlistConfirmationEmail({
+          email,
+          isProduction: process.env.NODE_ENV === 'production',
+        }).catch((error) => {
```
cjroth: No need to catch and re-log

backend/src/waitlist/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+type SendWaitlistEmailParams = {
+  email: string
+  isProduction: boolean
+}
+
+/**
+ * Send waitlist confirmation email
+ * Uses a simple template saying "Thanks for signing up! We'll let you know when it's ready."
+ */
+export const sendWaitlistConfirmationEmail = async ({
```
cjroth: Could we name this "sendJoinedWaitlistEmail"? The word "confirmation" threw me off bc I was thinking of where users have to click the link to confirm their email.

cjroth (changes requested):

backend/src/auth/auth.ts
ital0: @cjroth I was wondering about that. Because this change was intentional. I know the goal is to have `optimistic code over defensive code` (as per `AGENTS.md`), but wouldn't make sense to catch if it errored? I mean, I can convert this block of code to `async wait` inside a `try catch`, but just to keep track if this email failed or not. 

What do you think?

backend/src/waitlist/routes.ts
ital0: @cjroth that makes sense for this. I just followed what we have in `CLAUDE.md`. But yeah, it doesn't make sense to soft delete it.

backend/src/auth/auth.ts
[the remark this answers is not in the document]
ital0: @cjroth I'm sorry, I was wrong. I completely forgot about that BetterAuth was abstracting this whole layer.
I think we can easily fix this adding a props called `databaseHooks` where we can normalize the emails provided or we can use this plugin: https://github.com/gekorm/better-auth-harmony/blob/main/packages/plugins/README.md

Honestly, I think it makes sense to normalize the email addresses before persiting them in the database.

backend/src/waitlist/routes.ts
ital0: @cjroth yeap, thats correct! When I was building this I was counting with the `authentication` layer that is planned.
Let me know if that doesn't make sense so I can check if user is already registered/approved in a different way.

backend/src/auth/auth.ts
cjroth: My thought was that await will cause it to throw (if we don't await, you either have to add the .catch like you did or it will silently disappear) - but IMO adding the `await` is simpler and more readable than having to add the .catch and then log it. If it gets thrown because of the await, then we let the Elysia middleware catch it and handle it, which seems reasonable to me.

cjroth: Just to clarify - I don't think we need a try/catch here either (because Elysia's middleware should catch it) so it still falls into the guidance of being optimistic. I'm also not worried about the slight delay from waiting for the email to send that awaiting it would cause - I think that's expected for an endpoint like this

backend/src/auth/auth.ts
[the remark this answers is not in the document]
cjroth: I think simply lowercasing the emails before we create the row in the db should be simple and effective here - or is there more to it than that?

backend/src/auth/auth.ts
```diff
@@ -47,16 +44,44 @@ export const createAuth = (database: typeof DbType) =>
...
+              .where(eq(waitlist.email, normalizedEmail))
+              .limit(1)
+
+            // If not on waitlist or not approved, don't send OTP
+            if (waitlistEntry.length === 0 || waitlistEntry[0].status !== 'approved') {
+              console.info('🚫 Blocked sign-in attempt for non-approved email')
+
+              // If on waitlist but not approved, send a "not ready yet" email
+              if (waitlistEntry.length > 0) {
+                try {
```
cjroth: Still don't think we need a try / catch error (just let it throw and let Elysia error middleware catch it)

backend/src/auth/utils.ts
```diff
@@ -60,65 +62,32 @@ export const buildVerifyUrl = (origin: string, email: string, otp: string, reque
...
-export const sendSignInEmail = async ({
-  resend,
-  email,
-  otp,
-  verifyUrl,
-  isProduction,
-}: SendSignInEmailParams): Promise<void> => {
-  console.info(`📧 Sending sign-in email to ${email}`)
+export const sendSignInEmail = async ({ email, otp, verifyUrl }: SendSignInEmailParams): Promise<void> => {
+  console.info('📧 Sending sign-in email')
```
cjroth: I think we could skip this log

backend/src/lib/resend.ts
```diff
@@ -0,0 +1,63 @@
+import { Resend } from 'resend'
+
+/** Default sender address for all outgoing emails */
+const EMAIL_FROM = 'hello@auth.thunderbolt.io'
```
cjroth: Let's stick to camel case const names

backend/src/lib/resend.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!resend || process.env.NODE_ENV === 'test') {
+    if (isProduction && !resend) {
+      throw new Error('Email service not configured')
+    }
+    return true
+  }
+  return false
+}
+
+type SendEmailParams = {
```
cjroth: It could be useful to export this

backend/src/waitlist/README.md
```diff
@@ -0,0 +1,83 @@
...
+4. Users on waitlist but not approved receive a "not ready yet" email instead of the OTP
+5. Users not on the waitlist receive no email (prevents email enumeration)
+
+## Email Templates
+
+Three email types are sent via Resend:
+
+| Function                    | Template ID          | When sent                                           |
+| --------------------------- | -------------------- | --------------------------------------------------- |
+| `sendJoinedWaitlistEmail`   | `waitlist-joined`    | User joins the waitlist                             |
```
cjroth: Haha this is super nitpicky but it would be more consistent with other function names to name it `sendWaitlistJoinedEmail`

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,71 @@
...
+      // Check if email already exists
+      const existing = await database
+        .select({ id: waitlist.id })
+        .from(waitlist)
+        .where(eq(waitlist.email, email))
+        .limit(1)
+
+      // If entry exists, send reminder email (prevents email enumeration)
+      if (existing.length > 0) {
+        try {
```
cjroth, line 23: And here

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,71 @@
...
+          await sendWaitlistReminderEmail({ email })
+        } catch (error) {
+          console.error('Failed to send waitlist reminder email:', error)
+        }
+
+        return { success: true }
+      }
+
+      // Add new entry to waitlist
+      try {
```
cjroth: And here

backend/src/waitlist/routes.ts
```diff
@@ -0,0 +1,71 @@
...
+            console.error('Failed to send waitlist reminder email:', err)
+          }
+
+          return { success: true }
+        }
+
+        throw error
+      }
+
+      try {
```
cjroth, line 41: And here

cjroth (changes requested):

ital0:
@cjroth thread about the email normalization.

I know it's a pretty basic thing, but since we have BetterAuth in the mid to handle this, I ended up using `databaseHook`, as I early mentioned to you. What I added normalizes the email right before the user gets saved. This ensures that no matter how the user ends up in our database (OTP flow, future OAuth, etc.), their email is always normalized.

One thing I learned while testing: BetterAuth validates email format before our hooks run, so if someone types " test@example.com " with spaces, BetterAuth rejects it before we can trim it. That's actually fine for us because the frontend's email input should handle that, but good to know.

Another something worth mentioning: I initially looked into using Elysia middleware to centralize the email normalization (so we could just do it once at the request level for everything). Turns out that doesn't work for BetterAuth.

The reason is that BetterAuth is mounted using `.mount(auth.handler)`, which means it receives the raw `Request` object directly and does its own body parsing internally. Elysia middleware that modifies `body.email` never actually touches what BetterAuth sees — they're operating on different parsed versions of the request.

So for the waitlist (which is a regular Elysia route), middleware would've worked fine. But for BetterAuth user creation, we need the `databaseHooks` approach to intercept right before the user gets persisted.

Just wanted to document this in case you wonder why we didn't go the "simpler" here.

PR #346, opened 2026-01-27
THU-250: Failed inference response retry

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Adds timed auto-retry behavior to the chat send/regenerate flow and new session state to drive UI; mistakes could cause unintended repeated requests or confusing error states. Changes are localized to chat transport/UI and analytics, with updated tests/stories reducing risk.
> 
> **Overview**
> Implements **automatic retries for failed/empty assistant responses** in `createChatInstance`, using exponential backoff + jitter, cancelling retries on abort/session switch, and resetting retry state on manual `regenerate()` or new `sendMessage()`.
> 
> Extends chat session state with `retryCount`/`retriesExhausted`, updates the chat UI to show a **two-phase error banner** (retrying spinner vs. manual Retry button) and to hide broken empty assistant messages during errors, and adds PostHog tracking via new `chat_auto_retry` event.
> 
> Updates hydration/test mocks and adds/adjusts tests + Storybook stories to cover retrying vs. exhausted vs. stale-error scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be1aaa5cfd829472d185155b2758983fad128cbc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/chats/chat-instance.ts
```diff
@@ -6,6 +6,8 @@ import { DefaultChatTransport } from 'ai'
 import { v7 as uuidv7 } from 'uuid'
 import { useChatStore } from './chat-store'
 
+const MAX_RETRIES = 3
```
cjroth: Should be camel case

src/components/chat/error-message.tsx
```diff
@@ -1,14 +1,37 @@
...
 }
 
-export const ErrorMessage = memo(({ message }: ErrorMessageProps) => {
+export const ErrorMessage = memo(({ retriesExhausted, onRetry }: ErrorMessageProps) => {
+  if (!retriesExhausted) {
+    return (
+      <div className="p-4 rounded-md bg-amber-500/10 border border-amber-500/20 mr-auto w-full mt-2">
+        <div className="flex items-center gap-2">
+          <Loader2 className="size-3.5 text-amber-500 animate-spin" />
+          <p className="text-amber-500/80 text-sm">Hm, something went wrong. Retrying...</p>
```
cjroth: The "hm" feels weird here, I think just "something went wrong... etc" part is good

cjroth (changes requested):
Looks good. Here's what Opus said:

I'll review the branch diff. Let me read a few of the key files to get full context on the implementation.

[5 tools called]

## Code Review: Inference Call Retry Feature

Overall this is a well-structured implementation. The retry logic, UI feedback, and offline detection are cleanly implemented. Here's my detailed review:

---

### Issues to Address

**1. Potential race condition with `retryCount` closure**

```48:82:src/chats/chat-instance.ts
    onFinish: async ({ message, isError, isAbort }) => {
      if (!isError && !isAbort) {
        retryCount = 0
        useChatStore.getState().updateSession(id, { retriesExhausted: false })
        // ...
      }
      // ...
      if (retryCount < MAX_RETRIES) {
        retryCount++
        // ...
      }
    },
```

The `retryCount` is captured in a closure and persists across the lifetime of the chat instance. If the user sends a new message while a retry is pending (in the `setTimeout`), the retry will fire and potentially cause unexpected behavior. Consider:
- Clearing any pending retry timeout when `sendMessage` is called
- Or storing a message ID to verify the retry is still relevant

**2. Missing `retriesExhausted` reset on new message**

When the user sends a new message, `retriesExhausted` should be reset to `false`. Currently it's only reset on successful completion. If a user sends a new message after retries were exhausted, the error state would incorrectly persist until the new message succeeds.

**3. The `ErrorMessage` component always receives `onRetry` but doesn't use it when retrying**

```92:92:src/components/chat/chat-messages.tsx
      {hasError && <ErrorMessage retriesExhausted={retriesExhausted} onRetry={() => regenerate()} />}
```

During automatic retries (`retriesExhausted: false`), the manual `onRetry` prop is passed but unused. This is fine, but consider whether users should have the option to manually retry early (skip waiting for auto-retry).

---

### Suggestions

**4. Consider exponential backoff cap**

```75:79:src/chats/chat-instance.ts
        setTimeout(() => {
          instance.regenerate().catch((err) => {
            console.error('Auto-retry failed:', err)
          })
        }, 2000 * retryCount)
```

The linear backoff (2s, 4s, 6s) is reasonable, but you might want to add jitter to avoid thundering herd issues if multiple users hit errors simultaneously:

```typescript
const delay = 2000 * retryCount + Math.random() * 500
```

**5. The `useOnlineStatus` hook is clean and idiomatic** 

The `useSyncExternalStore` usage is the correct modern approach.

**6. Consider debouncing the offline banner**

Brief network blips could cause the banner to flash. You might want a small delay (500ms) before showing the offline state to avoid flicker.

**7. Minor: `trackEvent` missing from `useCallback` deps**

```94:101:src/components/chat/chat-prompt-input.tsx
    const handleShowOverflowModal = useCallback((model: Model, length: number, prompt_number: number) => {
      setShowOverflowModal(true)
      trackEvent('chat_send_prompt_overflow', {
        model,
        length,
        prompt_number,
      })
    }, [])
```

`trackEvent` is a dependency but not in the array. Since it's injected via props (for testing), this could cause stale closure issues. Add it to deps or assert it's stable.

---

### What's Good

- Clean separation of concerns between retry logic (chat-instance) and UI (components)
- Hiding the broken assistant message during retry is a nice UX detail
- The `useSyncExternalStore` approach for online status is the correct pattern
- Tests updated appropriately
- The "retrying" vs "exhausted" UI distinction provides good feedback

---

### Recommended Fix for Issue #1 & #2

In `chat-instance.ts`, track the pending timeout and clear it appropriately:

```typescript
let retryCount = 0
let retryTimeoutId: ReturnType<typeof setTimeout> | null = null

// In the sendMessage override, add:
if (retryTimeoutId) {
  clearTimeout(retryTimeoutId)
  retryTimeoutId = null
  retryCount = 0
}
useChatStore.getState().updateSession(id, { retriesExhausted: false })

// In onFinish, use:
retryTimeoutId = setTimeout(() => {
  retryTimeoutId = null
  instance.regenerate().catch(...)
}, 2000 * retryCount)
```

Would you like me to implement these fixes?

src/components/chat/chat-messages.tsx
```diff
@@ -12,23 +12,29 @@ type ChatMessagesProps = {
...
-      return chatError.message
-    }
+  const hasError = useMemo(() => {
+    if (chatError) return true
 
     const lastMessage = messages[messages.length - 1]
-    if (lastMessage?.role === 'assistant' && !lastMessage.parts?.length && !isStreaming) {
-      return 'Something went wrong. Please try again.'
-    }
+    if (lastMessage?.role === 'assistant' && !lastMessage.parts?.length && !isStreaming) return true
```
ital0: I think this condition can be simplified to `return lastMessage?.role === 'assistant' && !lastMessage.parts?.length && !isStreaming`

ital0: and you can get rid of the `return false` below

src/chats/chat-instance.ts
```diff
@@ -34,31 +36,53 @@ export const createChatInstance = (
...
   const instance = new Chat<ThunderboltUIMessage>({
     id,
     messages,
     transport: new DefaultChatTransport({ fetch: customFetch }),
     generateId: uuidv7,
     // Automatically send messages when the last one is a user message (used for automations)
     sendAutomaticallyWhen: ({ messages }) => messages.length > 0 && messages[messages.length - 1].role === 'user',
-    onFinish: async ({ message }) => {
-      const { sessions } = useChatStore.getState()
+    onFinish: async ({ message, isError, isAbort }) => {
```
ital0, line 55: suggestion: the AI SDK provides `isDisconnect`. It might make sense to have a retry strategy here using this flag to wait for the reconnection.

src/components/chat/chat-messages.tsx
```diff
@@ -12,23 +12,29 @@ type ChatMessagesProps = {
...
-  const error = useMemo(() => {
-    if (chatError) {
-      return chatError.message
-    }
+  const hasError = useMemo(() => {
+    if (chatError) return true
 
     const lastMessage = messages[messages.length - 1]
-    if (lastMessage?.role === 'assistant' && !lastMessage.parts?.length && !isStreaming) {
-      return 'Something went wrong. Please try again.'
```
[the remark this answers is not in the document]
ital0: suggestion: show `Retrying 2/3` to the user.

ital0 (changes requested):
nice work! Just pointed out some minor detail.

src/chats/chat-instance.ts
```diff
@@ -34,38 +36,83 @@ export const createChatInstance = (
...
+      if (retryCount < maxRetries) {
+        retryCount++
+        useChatStore.getState().updateSession(id, { retryCount })
+        console.info(`Auto-retrying (${retryCount}/${maxRetries})...`)
+        retryTimeout = setTimeout(() => {
+          retryTimeout = null
+          instance.regenerate().catch((err) => {
+            console.error('Auto-retry failed:', err)
+          })
+        }, 2000 * retryCount)
```
cjroth: Can we add some Math.random here so that we don't overload servers with synchronized retries if our backend goes offline?

src/components/chat/chat-messages.tsx
```diff
@@ -12,23 +12,34 @@ type ChatMessagesProps = {
...
-    }
+    return lastMessage?.role === 'assistant' && !lastMessage.parts?.length && !isStreaming
   }, [chatError, messages, isStreaming])
 
+  // Find the last assistant message index so we can hide it during errors
```
cjroth: What would happen if the error occurs before any assistant message is sent?

src/components/chat/chat-messages.tsx
```diff
@@ -63,6 +74,11 @@ export const ChatMessages = ({ useChat = useChat_default }: ChatMessagesProps) =
         }
 
         if (message.role === 'assistant') {
+          // Skip the last assistant message when there's an error — it's the broken
+          // response that regenerate() will remove. This prevents layout shift when
+          // the retry fires and the message disappears.
+          if (hasError && i === lastAssistantIndex) return null
```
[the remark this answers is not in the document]
cjroth: Yeah this

src/components/chat/error-message.tsx
```diff
@@ -1,14 +1,41 @@
...
-interface ErrorMessageProps {
-  message: string | null
+type ErrorMessageProps = {
+  retryCount: number
+  retriesExhausted: boolean
+  onRetry?: () => void
 }
 
-export const ErrorMessage = memo(({ message }: ErrorMessageProps) => {
+export const ErrorMessage = memo(({ retryCount, retriesExhausted, onRetry }: ErrorMessageProps) => {
```
cjroth, line 11: Can we add a storybook story for this so that we can test this?

I think it's nice for UI components like this that are tricky to get to because of the state needed to recreate them

src/chats/chat-instance.ts
```diff
@@ -34,38 +36,83 @@ export const createChatInstance = (
...
-        length: message.parts?.reduce((acc, part) => acc + (part.type === 'text' ? part.text.length : 0), 0) || 0,
-        reply_number: instance.messages.length + 1,
-      })
+    onFinish: async ({ message, isError, isAbort }) => {
+      if (isAbort) return
+
+      // Empty response without an error flag — treat as a retryable failure
+      const isEmptyResponse = !isError && !message.parts?.length
+
+      if (!isError && !isEmptyResponse) {
```
cjroth: One consideration for the future is that we might want some errors to not retry, such as if the user hits their own rate limit for Thunderbolt or something like that. Nothing we need to change right now, just noting for future.

cjroth (changes requested):
Looks good! Almost there - just a few more things I noticed this time.

src/chats/chat-instance.ts
cjroth: @darkbanjo should we do this?

cjroth (changes requested):
I just did another review and made a few changes based on Claude's review.

Two things we need to address:
1. The app can be used offline with Ollama, so we need to make sure the offline status makes sense here.
2. When I put the network into "offline" in the devtools, I wasn't seeing the offline status badge anymore. Need to verify if it's working.
3. You can get it into a state where it will say "Something went wrong. Retrying (0/3)..." and just be stuck if you interrupt it in the middle of retrying (eg refreshing page). We need to fix this.

cjroth (approved):

PR #360, opened 2026-02-05
PowerSync: Pre-release [BE]

raivieiraadriano92:
> [!NOTE]
> **High Risk**
> Adds new sync and account-deletion APIs and introduces JWT-based PowerSync token issuance plus server-side application of client-provided CRUD operations, which are security- and data-integrity sensitive. Also includes a DB migration and cascade-delete behavior that could cause irreversible data loss if misused.
> 
> **Overview**
> Adds **backend PowerSync support**: new `/powersync/token` issues signed JWTs (requires `x-device-id`, enforces device revocation/ownership, upserts device heartbeat/name) and `/powersync/upload` applies queued `PUT`/`PATCH`/`DELETE` operations via Drizzle while scoping all writes by authenticated `user_id` and filtering/typing payload columns (including ISO timestamp coercion).
> 
> Introduces **self-service account/device management** endpoints under `/account` (`DELETE /account` hard-deletes the user and relies on FK cascade to remove synced rows; `POST /account/devices/:id/revoke` soft-disconnects a device via `revokedAt`).
> 
> Extends auth+schema to track `user.isNew` (DB column + Better Auth additional field) and flips it to `false` after first successful email-OTP sign-in; adds PowerSync configuration to settings/env, updates docs, adds shared table-name typing via TS path mapping, and includes extensive new tests plus dependency/tooling updates (notably `@elysiajs/jwt` and ESLint v9).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4918cc95d9ae63cfe9a128940c73bd4d3b46b7e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


backend/src/api/account.ts
```diff
@@ -0,0 +1,54 @@
...
+        .where(and(eq(devicesTable.id, deviceId), eq(devicesTable.userId, userId)))
+      set.status = 204
+    })
+    .delete('/', async ({ set, user: sessionUser }) => {
+      const userId = sessionUser!.id
+
+      await database.transaction(async (tx) => {
+        for (const name of POWERSYNC_TABLE_NAMES) {
+          const table = POWERSYNC_TABLES_BY_NAME[name]
+          await tx.delete(table).where(eq(table.userId, userId))
```
ital0: should this really delete or should it follow the `soft delete` approach?

backend/src/api/powersync.ts
```diff
@@ -0,0 +1,263 @@
...
+ * Apply a single PowerSync operation to the database using raw SQL.
+ * The user_id is always set to the authenticated user to ensure data isolation.
+ */
+const applyOperation = async (op: PowerSyncOperation, userId: string, database: typeof DbType): Promise<void> => {
+  if (!VALID_TABLES.has(op.type)) {
+    console.warn(`Unknown table: ${op.type}`)
+    return
+  }
+
+  console.info(`Applying ${op.op} to ${op.type} with id=${op.id} for user=${userId}`)
```
ital0: might no be a good idea to log user IDs

backend/src/api/powersync.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      // Build UPDATE SET clause for all columns except id
+      // Only update if the row belongs to this user
+      const updateColumns = Object.keys(allData).filter((k) => k !== 'id')
+      const updateClause =
+        updateColumns.length > 0
+          ? `DO UPDATE SET ${updateColumns.map((k) => `"${k}" = EXCLUDED."${k}"`).join(', ')} WHERE "${op.type}"."user_id" = ${escapeValue(userId)}`
+          : 'DO NOTHING'
+
+      const query = `INSERT INTO "${op.type}" (${columns}) VALUES (${values}) ON CONFLICT (id) ${updateClause}`
```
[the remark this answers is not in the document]
ital0: One addition to this:

Do you think we should use Drizzle ORM instead? It would eliminate most of the SQL injection concerns and give us better type safety.

I don't know the reason of using raw SQL. Maybe PowerSync recommends it

backend/src/auth/auth.ts
```diff
@@ -40,6 +41,36 @@ export const createAuth = (database: typeof DbType) =>
...
+        const newSession = ctx.context.newSession
+        if (!newSession?.user) {
+          return
+        }
+
+        const sessionUser = newSession.user
+
+        // Check if user is new by comparing createdAt and updatedAt timestamps
+        // A new user has never had their updatedAt modified after creation
+        const isNewUser = new Date(sessionUser.createdAt).getTime() === new Date(sessionUser.updatedAt).getTime()
```
ital0: if the user signs in exactly when created (same millisecond), this could have edge cases. Probably rare enough not to matter, but wanted to mention it.

backend/src/api/powersync.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { Settings } from '@/config/settings'
+import { session as sessionTable, user as userTable } from '@/db/auth-schema'
+import type { db as DbType } from '@/db/client'
+import { devicesTable } from '@/db/schema'
+import { POWERSYNC_TABLE_NAMES } from '@shared/powersync-tables'
+import { jwt } from '@elysiajs/jwt'
+import { eq, sql } from 'drizzle-orm'
+import { Elysia, t } from 'elysia'
+
+const VALID_TABLES = new Set<string>(POWERSYNC_TABLE_NAMES)
```
ital0: constants must use `camelCase` format

ital0 (changes requested):
I really liked the approach used here and how you documented everything. **This will help a lot.**

An important mention is about the raw SQL construction with client-supplied data. IMO, it creates critical security vulnerabilities that must be addressed before moving forward.

I’m not sure if this solution is definitive, but I’d also suggest mentioning the addition of some tests that I couldn’t find in your PR:

* SQL injection via column names — 0% coverage

* Object spread ID override — 0% coverage

* `user_id` isolation enforcement — 0% coverage

* Cross-user PATCH/DELETE authorization — 0% coverage

* Transaction rollback on failure — 0% coverage

* Non-primitive value handling — 0% coverage

* Device revocation authorization — 0% coverage

* JWT validation — 30% coverage

(The coverage percentages were generated by Claude.)

That said, I think it’s only worth adding these if we decide to move forward with this approach as the definitive path.

backend/src/api/account.ts
raivieiraadriano92: in this case it should be hard deleted once the user is actually deleting his account

backend/src/api/powersync.ts
raivieiraadriano92: ✅

backend/src/api/powersync.ts
raivieiraadriano92: ✅

raivieiraadriano92:
> I really liked the approach used here and how you documented everything. **This will help a lot.**
> 
> An important mention is about the raw SQL construction with client-supplied data. IMO, it creates critical security vulnerabilities that must be addressed before moving forward.
> 
> I’m not sure if this solution is definitive, but I’d also suggest mentioning the addition of some tests that I couldn’t find in your PR:
> 
> * SQL injection via column names — 0% coverage
> * Object spread ID override — 0% coverage
> * `user_id` isolation enforcement — 0% coverage
> * Cross-user PATCH/DELETE authorization — 0% coverage
> * Transaction rollback on failure — 0% coverage
> * Non-primitive value handling — 0% coverage
> * Device revocation authorization — 0% coverage
> * JWT validation — 30% coverage
> 
> (The coverage percentages were generated by Claude.)
> 
> That said, I think it’s only worth adding these if we decide to move forward with this approach as the definitive path.

good catch - just applied some changes and now we are using the drizzle schema and also validating all columns

backend/src/config/settings.test.ts
```diff
@@ -93,6 +93,11 @@ describe('Config Settings', () => {
           port: 8000,
           posthogHost: 'https://us.i.posthog.com',
           posthogApiKey: '',
+          waitlistEnabled: false,
```
cjroth, line 96: Should this come from an env var?

cjroth (changes requested):
Overall, this is GREAT. Really, really, really well architected and high quality PR.

Things to look at:
1. Take a look at the security issues Claude found.
2. Can you resolve or fix other Claude / Cursor comments?
3. Here's what I found while chatting with Claude about the PR. I think I agree with most of it.

---

OK, I've now re-read the full diffs of both PRs with the correct understanding that PowerSync stores data as schemaless JSON with SQLite views on top, meaning schema evolution is handled automatically. Here's a fresh review.

---

## PR Summary

Two PRs adding PowerSync-based multi-device sync. Backend (#360, 28 files, +2478) adds sync token issuance, client-driven upload, account deletion, and device revocation. Frontend (#361, 116 files, +2482/-14424) replaces the local database layer with PowerSync, adds sync UI, device management, and a credentials-invalid reset flow. Most of the deletions are the old SQL migration system (files + snapshots), which PowerSync's internal schema management replaces.

The architecture is clean: PowerSync stores data as schemaless JSON in `ps_data__*` tables and presents SQLite views to the app. Schema changes between app versions are handled automatically by `powersync_replace_schema()` -- no migration code needed. Sync is opt-in and off by default, so the app works fully locally without sign-up.

---

## Concerns

### 1. `waitForInitialSync` can hang indefinitely

In `database.ts`, `waitForInitialSync()` returns a Promise that only resolves when `hasSynced` becomes true. In `use-app-initialization.ts`, this replaced the old migration step:

```typescript
// Step 3
await DatabaseSingleton.instance.waitForInitialSync()
```

If sync is enabled but the network is down (or PowerSync Cloud is unreachable), this never resolves and the app never finishes initializing. The `catch` around it is non-critical (just warns), but the `await` itself blocks. Needs a timeout -- something like `Promise.race([waitForInitialSync(), sleep(10_000)])`.

### 2. `setAuthToken` changed from async to sync but callers still `await` it

`auth-token.ts` was refactored from storing the token in the settings database (async) to localStorage (sync). `setAuthToken` is now `(token: string): void` but the sign-in flow still does `await setAuthToken(result.data.token)`. That's fine functionally (awaiting a non-Promise is a no-op), but it's misleading -- the function signature should match usage, or the `await` should be removed.

### 3. `isNewUser` detection via timestamp equality is fragile

```typescript
const isNewUser = new Date(sessionUser.createdAt).getTime() === new Date(sessionUser.updatedAt).getTime()
```

Any middleware, plugin, or hook that touches `updatedAt` during user creation would break this. The hook then sets `updatedAt` to `new Date()` to mark first sign-in complete, so this is a one-time flag encoded as a side effect of timestamp equality. A dedicated column or counting sessions would be more reliable.

### 4. `updateSettings` lost batch atomicity

The old code did a single batch upsert:

```typescript
await db.insert(settingsTable).values(values).onConflictDoUpdate(...)
```

The new code loops one-by-one with try/catch per row:

```typescript
for (const [key, value] of entries) {
  try { await db.insert(...) }
  catch (err) { if (!isInsertConflictError(err)) throw err; await db.update(...) }
}
```

This means if you call `updateSettings({ a: '1', b: '2', c: '3' })` and `b` fails with a non-conflict error, `a` is already committed but `c` never runs. The old upsert was atomic. This is a real regression if any call site depends on all-or-nothing behavior. Consider wrapping the loop in a transaction.

### 5. `createSetting` has a TOCTOU race

```typescript
const existing = await db.select(...).where(eq(settingsTable.key, key)).get()
if (!existing) {
  await db.insert(settingsTable).values({ key, value })
}
```

The check-then-insert has a race window. Two concurrent callers can both see `existing` as null and both try to insert. The old `onConflictDoNothing` was atomic. If `onConflictDoNothing` doesn't work with PowerSync views, this should at least catch the conflict error and ignore it.

### 6. Upload endpoint swallows individual operation errors

```typescript
for (const op of operations) {
  try {
    await applyOperation(op, user.id, database)
  } catch (error) {
    console.error(`Failed to apply operation:`, op, error)
    // Continue processing other operations
  }
}
return { success: true }
```

PowerSync recommends returning 2xx, which is fine. But logging just the op object may not include enough context to debug issues in production (user id, table, which column failed). More importantly, the console.error stringification of `op` could log sensitive user data. Consider logging only `op.type`, `op.op`, and the error message.

### 7. Device name validation skips upsert silently

In `issuePowerSyncToken`, if `x-device-name` is empty or > 100 chars, the token is still issued but the device is never upserted:

```typescript
const isDeviceNameValid = deviceName && deviceName.length > 0 && deviceName.length <= 100
if (isDeviceNameValid) { /* upsert */ }
```

This means the device won't appear in the devices list at all. A token is issued for a device that doesn't exist in the database. If the user then tries to revoke it, there's nothing to revoke. The device should probably be upserted with a fallback name (e.g., "Unknown device") rather than silently skipped.

### 8. `usePowerSyncInvalidation` runs a `SELECT` per table continuously

```typescript
for await (const _ of powerSync.watch(`SELECT 1 FROM ${tableName} LIMIT 1`, {
  signal: abortController.signal,
})) {
  for (const queryKey of queryKeys) {
    queryClient.invalidateQueries({ queryKey })
  }
}
```

This sets up 10 concurrent async iterators (one per synced table), each running a watch query. Any write to any row in a table invalidates all associated query keys. For high-frequency tables like `chat_messages`, this could cause excessive React Query invalidation and re-renders during active sync. PowerSync has a `registerListener` with `tablesUpdated` that batches notifications -- that might be more efficient than 10 independent watch queries.

### 9. `getDatabaseType` hardcodes `'powersync'` with dead code

```typescript
export const getDatabaseType = async (): Promise<DatabaseType> => {
  return 'powersync'
  // For web, use PowerSync as default
  // if (!isTauri()) {
  // }
  // const { libsql } = await getCapabilities()
  // return libsql ? 'libsql-tauri' : 'wa-sqlite'
}
```

The Tauri/libsql paths are commented out. This means Tauri desktop builds also get PowerSync instead of libsql. If that's intentional, the dead code should be removed. If Tauri support is still needed, this needs a proper conditional.

### 10. `performCredentialsInvalidReset` has no error handling

```typescript
const performCredentialsInvalidReset = async (): Promise<void> => {
  await setSyncEnabled(false)
  await resetAppDir()
  localStorage.clear()
  window.location.reload()
}
```

If `setSyncEnabled(false)` or `resetAppDir()` throws, the user gets stuck in a broken state -- they never reach `localStorage.clear()` and `reload()`. Since the whole point is to force a clean state, this should be resilient:

```typescript
try { await setSyncEnabled(false) } catch { /* ignore */ }
try { await resetAppDir() } catch { /* ignore */ }
localStorage.clear()
window.location.reload()
```

---

## Smaller items

- **`console.info` in connector**: `"Uploading N operations"` and `"PowerSync upload completed successfully"` will be noisy in production. Use the app's logger or remove.
- **Docker hardcoded password**: `myhighlyrandompassword` in config.yaml and init SQL. Fine for local dev, but the ironic name will confuse someone eventually. Consider `powersync-local-dev` or similar.
- **`clearPendingCrudOperations` uses raw SQL**: `DELETE FROM ps_crud` ties to PowerSync internals. The PowerSync SDK might have a proper API for this; worth checking.
- **Settings table column rename**: `key` in TypeScript maps to `id` in the DB. The inline comment explains why, but it's a footgun for anyone writing raw queries.

---

## What's done well

- **Architecture**: The split between backend (Postgres + Drizzle) and frontend (PowerSync + SQLite views) is clean. The `shared/powersync-tables.ts` module is a nice single source of truth.
- **Security**: Upload operations strip `id` and `user_id` from client payloads and always use the session user. Column allowlisting via Drizzle schema prevents injection. User isolation is tested.
- **Reset flow**: The credentials-invalid listener is thoughtfully placed in `AuthProvider` above the early return, with clear documentation of why. The `hadDeviceOnceRef` guard to avoid false resets on first load is well-considered.
- **Test coverage**: Both PRs have extensive tests covering auth edge cases, cross-user isolation, cascade deletes, device revocation, and the insert-first pattern.
- **Documentation**: The PR descriptions, inline docs, and separate markdown files (`composite-primary-keys-and-default-data.md`, `delete-account-and-revoke-device.md`, `powersync-account-devices.md`) are thorough.
- **`isInsertConflictError`**: Handles multiple SQLite environments (Bun, wa-sqlite, PowerSync) with numeric codes, string codes, and message patterns. Well-engineered.

PR #361, opened 2026-02-05
PowerSync: Pre-release [FE]

raivieiraadriano92:
# PowerSync, Account & Device Management

This document consolidates documentation for:

- **PowerSync**: multi-device sync (synced tables, local dev, adding tables)
- **Account**: deletion flow and how other devices reset
- **Devices**: registration, list, revoke, and how a revoked device resets

---

## 1. PowerSync overview

PowerSync provides offline-first sync between the backend (PostgreSQL) and clients (SQLite). Data is scoped by `user_id` from the JWT. The backend issues PowerSync JWTs and can apply client uploads (PUT/PATCH/DELETE) to Postgres. Production uses PowerSync Cloud; local development uses the Docker stack in `powersync-service/`.

---

### Requirements

- Every synced table must have a **`user_id`** column (sync rules and backend scope by `user_id`).
- Define the table in **both**:
  - Frontend: [src/db/tables.ts](src/db/tables.ts) (SQLite)
  - Backend: [backend/src/db/powersync-schema.ts](backend/src/db/powersync-schema.ts) (PostgreSQL)

### Current tables

Defined in [shared/powersync-tables.ts](shared/powersync-tables.ts):

`settings`, `chat_threads`, `chat_messages`, `tasks`, `models`, `mcp_servers`, `prompts`, `triggers`, `modes`, `devices`.

### Adding a new synced table

1. Create the table in both `src/db/tables.ts` and `backend/src/db/powersync-schema.ts` (include `user_id`).
2. Register in [src/db/powersync/schema.ts](src/db/powersync/schema.ts) (`drizzleSchema`).
3. Add the table name and query keys in [shared/powersync-tables.ts](shared/powersync-tables.ts) (`POWERSYNC_TABLE_NAMES` and `POWERSYNC_TABLE_TO_QUERY_KEYS`).
4. Update [powersync-service/config/config.yaml](powersync-service/config/config.yaml): add a line under `sync_rules.content` → `bucket_definitions.user_data.data` (e.g. `- SELECT * FROM my_table WHERE my_table.user_id = bucket.user_id`).
5. Run migrations for frontend and backend as needed.

**Before merging to production:** Deploy the new sync rules in the PowerSync Cloud dashboard (production uses PowerSync Cloud; local uses `powersync-service` config).

---

## 3. Local development (PowerSync Docker)

See [powersync-service/README.md](../powersync-service/README.md) for full steps. Summary:

- From `powersync-service/`: `docker compose up -d`
- PowerSync API: http://localhost:8080
- Postgres: localhost:5433 (use this for the backend so PowerSync and app share one database)
- Backend `.env`: set `DATABASE_DRIVER=postgres`, `DATABASE_URL=postgresql://postgres:postgres@localhost:5433/postgres`, and PowerSync vars (see below)
- Sync rules in `powersync-service/config/config.yaml` must match backend tables; when you add/change tables, update that file and `VALID_TABLES` in `backend/src/api/powersync.ts` (which uses `POWERSYNC_TABLE_NAMES` from shared).

### Backend PowerSync env vars (local)

```env
POWERSYNC_URL=http://localhost:8080
POWERSYNC_JWT_SECRET=powersync-dev-secret-change-in-production
POWERSYNC_JWT_KID=powersync-dev
POWERSYNC_TOKEN_EXPIRY_SECONDS=3600
```

The local `config/config.yaml` uses HS256 with the same secret (base64) and kid so backend-issued tokens are accepted.

---

## 4. Account deletion

- **Where:** Settings > Preferences → “Delete my account” (with confirmation).
- **Request:** Frontend calls `DELETE /v1/account` with the current auth token.
- **Backend:** Deletes the user and all related data (settings, chats, models, devices, etc.).
- **Other devices:** When PowerSync refreshes the token, the backend returns **410 Gone** with `code: 'ACCOUNT_DELETED'`. The app treats this as credentials invalid and runs the reset flow (see section 7).

---

### Devices table

- **Backend:** `devices` table: `id`, `user_id`, `name`, `last_seen`, `created_at`, `revoked_at`. Synced via PowerSync.
- **Frontend:** Same schema in the local DB; used for Settings > Devices and for “current device revoked?” checks.

### Listing devices

- **Where:** Settings > Devices.
- **Data:** Devices from the local DB (synced `devices` table) via `getAllDevices()` and React Query key `['devices']`.
- **UI:** Name, last seen, “This device” for current device, “Revoked” when `revoked_at` is set. “Revoke” only for other, non-revoked devices.

### Revoking a device

1. User chooses “Revoke” on another device (with confirmation). Frontend calls `POST /v1/account/devices/:id/revoke`.
2. Backend sets `revoked_at` on that device row (soft revoke). PowerSync syncs the updated `devices` table.
3. On the **revoked device**:
   - **Immediate:** The app watches the current device’s row via React Query (`getDevice(deviceId)`). When the synced row has `revoked_at` set, the app runs the reset flow.
   - **On token refresh:** Backend returns **403 Forbidden** with `code: 'DEVICE_DISCONNECTED'`; the connector dispatches credentials invalid and the app resets.

### Auth token and device id

- **Auth token:** In `localStorage` (fixed key). Cleared on reset via `localStorage.clear()`.
- **Device id:** In `localStorage`. Sent as `X-Device-ID` (and optional `X-Device-Name`) on PowerSync token requests so the backend can register/update the device and enforce revoke.

---

### PowerSync token (`GET /powersync/token`)

- **With `X-Device-ID`:**
  - Backend checks the `devices` row for that id. If `revoked_at` is set → **403** with `{ code: 'DEVICE_DISCONNECTED' }`, no token.
  - Otherwise: issues a PowerSync JWT and upserts the device (id, user_id, name, last_seen, created_at).
- **Bearer token only (e.g. credential refresh):**
  - If the user no longer exists (account deleted) → **410 Gone** with `{ code: 'ACCOUNT_DELETED' }`.
  - Otherwise may return **401** (invalid/expired token).

Summary for client:

- **410** → account deleted (reset).
- **403** with `DEVICE_DISCONNECTED` → this device revoked (reset).
- **401** → generic auth failure.

### Revoke device (`POST /v1/account/devices/:id/revoke`)

- Requires authenticated user (session).
- Sets `revoked_at` to current timestamp for the device `id` that belongs to the current user.
- **204** on success (idempotent for already-revoked devices).

---

## 7. Frontend: credentials-invalid and reset

When the app should reset (account deleted or device revoked), it runs a single flow:

1. `setSyncEnabled(false)` – disconnect from PowerSync.
2. `localStorage.clear()` – remove auth token and device id.
3. `resetAppDir()` – clear the app directory (DB and related files).
4. `window.location.reload()` – reload to a clean, signed-out state.

Triggered in two ways:

1. **Event `POWERSYNC_CREDENTIALS_INVALID`**  
   Dispatched when the token request returns **410** or **403** with body `code: 'DEVICE_DISCONNECTED'`.
2. **Devices table (current device revoked)**  
   `usePowerSyncCredentialsInvalidListener` uses React Query `getDevice(deviceId)` and key `['devices', deviceId]`. When the synced `devices` row has `revoked_at` set for the current device, the hook runs the same reset flow (immediate, without waiting for next token refresh).

After revoke, the Settings > Devices list is updated by invalidating `['devices']` so the list reflects the new state after sync.

---

## 8. Summary

| Action         | Where       | Backend / sync behavior                       | Other device behavior                                                          |
| -------------- | ----------- | --------------------------------------------- | ------------------------------------------------------------------------------ |
| Delete account | Preferences | User and data deleted; 410 on token refresh    | Reset when 410 received or when sync reflects deletion                         |
| Revoke device  | Devices     | Set `revoked_at`; 403 on that device’s refresh  | Revoked device resets when it sees `revoked_at` (useQuery) or gets 403 on refresh |

Both paths use the same reset: disable sync, clear localStorage, reset app dir, reload.

---

> [!NOTE]
> **High Risk**
> Large, cross-cutting change that alters database schema/types, write semantics (upsert→insert/update), and auth-adjacent reset behavior, so regressions could impact data integrity, sync correctness, or sign-in/session flows.
> 
> **Overview**
> Adds **PowerSync-based multi-device sync** to the frontend, including a new `PowerSyncDatabaseImpl` (Drizzle-wrapped), a backend connector for token fetch + upload, and a global sync enable/disable preference with UI controls (new header `PowerSyncStatus` + warning dialog).
> 
> Introduces **device management** on the client: a new synced `devices` table + DAL, a Settings > Devices page with revoke flow, and a credentials-invalid/reset mechanism that triggers a full local wipe/reload on account deletion or device revocation (wired into `AuthProvider` and token refresh events).
> 
> Refactors the local data layer for PowerSync compatibility: replaces SQLite upserts with insert-then-update fallbacks, implements DAL-level cascade soft deletes for chat messages, shifts `deletedAt`/timestamps to ISO strings, adds `user_id` columns across synced tables, and replaces migration-based test setup with schema application from Drizzle table definitions (`applySchema`).
> 
> Updates docs and deps for PowerSync (new consolidated doc, README pointers) and bumps `wa-sqlite` to support the new PowerSync web stack.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e724375d0ac1845df1362875f2a941b9564f6edf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/settings/preferences.tsx
```diff
@@ -255,6 +287,32 @@ export default function PreferencesSettingsPage() {
     }
   }
 
+  const handleDeleteAccount = async () => {
+    setDeleteAccountError(null)
+    dispatch({ type: 'SET_IS_DELETING_ACCOUNT', payload: true })
+    try {
+      await setSyncEnabled(false)
+      const token = getAuthToken()
+      if (!token) {
```
ital0, line 289: suggestion: I would add a `finally` block to handle (or before the early return) to handle with `SET_IS_DELETING_ACCOUNT` correctly. 

`dispatch({ type: 'SET_IS_DELETING_ACCOUNT', payload: false })`

On line 292, you're setting it as `true` but if it enters in this block, it doesn't reset its state.

src/hooks/use-powersync-credentials-invalid-listener.ts
```diff
@@ -0,0 +1,81 @@
...
+  const hasTriggeredRef = useRef(false)
+  const hadDeviceOnceRef = useRef(false)
+  const deviceId = getDeviceId()
+
+  const { data: device, isFetched } = useQuery({
+    queryKey: ['devices', deviceId],
+    queryFn: () => getDevice(deviceId),
+  })
+
+  if (device != null) hadDeviceOnceRef.current = true
```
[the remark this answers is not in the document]
ital0: This one makes sense @raivieiraadriano92

src/hooks/use-powersync-invalidation.ts
```diff
@@ -0,0 +1,91 @@
...
+    const database = DatabaseSingleton.instance.database
+    // Check if it's a PowerSync database with the powerSyncInstance getter
+    if ('powerSyncInstance' in database) {
+      return (database as { powerSyncInstance: PowerSyncDatabase | null }).powerSyncInstance
+    }
+  } catch {
+    // Not initialized or not PowerSync
+  }
+  return null
+}
```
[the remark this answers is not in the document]
ital0: @raivieiraadriano92 this one is true

ital0 (changes requested):
This looks amazing and I can't wait to get it deployed :) 

I think a good portion of Cursor’s reports make sense. I didn’t flag them, but the ones I did flag were valid.

I think it’s worth at least taking a look at them and maybe asking Claude (locally) to review the other issues the Claude reviewer reported and see whether they make sense.

src/settings/preferences.tsx
raivieiraadriano92: it makes sense - just added that

src/settings/devices.tsx
```diff
@@ -0,0 +1,151 @@
...
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from '@/components/ui/alert-dialog'
+
+const formatLastSeen = (ts: string | null): string => {
```
cjroth, line 23: Since we have dayjs included, we should use https://day.js.org/docs/en/display/from-now

src/settings/devices.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export default function DevicesSettingsPage() {
+  const queryClient = useQueryClient()
+  const currentDeviceId = getDeviceId()
+  const { data: devices = [], isLoading } = useQuery({
+    queryKey: ['devices'],
+    queryFn: getAllDevices,
+  })
+  const { cloudUrl } = useSettings({ cloud_url: 'http://localhost:8000/v1' })
+  const baseUrl = cloudUrl.value ?? 'http://localhost:8000/v1'
```
cjroth: this line `const baseUrl = cloudUrl.value ?? 'http://localhost:8000/v1'` should be unnecessary - `useSettings` already does this no?

src/settings/integrations.tsx
```diff
@@ -42,33 +44,118 @@ const ThunderboltProIcon = () => (
   </svg>
 )
 
+const INTEGRATION_SETTINGS_KEYS = [
```
cjroth: For consistency, I would use camelCase

cjroth (changes requested):
Nice work! Just added a few things I found

src/settings/integrations.tsx
raivieiraadriano92: done

src/settings/devices.tsx
raivieiraadriano92: agreed - I didn't notice that

src/settings/devices.tsx
raivieiraadriano92: done

cjroth (approved):

src/settings/preferences.tsx
```diff
@@ -255,6 +279,35 @@ export default function PreferencesSettingsPage() {
...
+      })
+      await clearAuthToken()
+      await resetAppDir()
+      window.location.reload()
+    } catch (error) {
+      console.error('Failed to delete account:', error)
+      setDeleteAccountError(error instanceof Error ? error.message : 'Failed to delete account.')
+      dispatch({ type: 'SET_IS_DELETING_ACCOUNT', payload: false })
+    } finally {
+      dispatch({ type: 'SET_IS_DELETING_ACCOUNT', payload: false })
```
ital0, line 306: Small bug: in `handleDeleteAccount`, the `catch` block dispatches `SET_IS_DELETING_ACCOUNT: false` and then the `finally` block does the exact same thing again. Double dispatch in one flow. Since `finally` always runs, you can just remove the one in `catch`.

src/lib/platform.ts
```diff
@@ -129,12 +126,39 @@ export const getDatabasePath = async (databaseType: DatabaseType, appDataDirPath
...
+
+/**
+ * Returns an auto-filled device display name (e.g. "Chrome on macOS", "Safari on iOS").
+ * Used for the synced devices table; not editable by the user.
+ */
+export const getDeviceDisplayName = (): string => {
+  if (isTauri()) {
+    const p = getPlatform()
+    const name = p.charAt(0).toUpperCase() + p.slice(1)
+    return `Thunderbolt on ${name}`
```
ital0, line 149: tiny privacy thought: `getDeviceDisplayName()` stores "Chrome on macOS" (parsed from user agent) in the synced devices table. For most users this is fine and it's genuinely useful UX, but it does reveal browser + OS to anyone who can read the database. Just flagging it.

What's your opinion on that @cjroth?

src/settings/preferences.tsx
```diff
@@ -729,37 +782,101 @@ export default function PreferencesSettingsPage() {
...
-                <AlertDialogAction
-                  onClick={handleResetDatabase}
-                  className="bg-destructive text-white hover:bg-destructive/90"
-                >
-                  Reset Database
-                </AlertDialogAction>
-              </AlertDialogFooter>
-            </AlertDialogContent>
-          </AlertDialog>
+      <SectionCard title="Sync">
```
ital0, line 784: The Sync section in Preferences shows the toggle to everyone, including unauthenticated users. The header's `PowerSyncStatus` correctly disables it when not signed in, but the Preferences toggle doesn't. Should we either gate this section behind `isAuthenticated` or add `disabled={!isAuthenticated}` to the Switch, similar to what the header does?

src/db/powersync/connector.ts
```diff
@@ -0,0 +1,130 @@
...
+   * Returns null if unable to get credentials (e.g., not authenticated or PowerSync not configured).
+   */
+  async fetchCredentials(): Promise<PowerSyncCredentials | null> {
+    const hadToken = Boolean(getAuthToken())
+    try {
+      if (!hadToken) {
+        return null
+      }
+
+      const response = await fetch(`${this.backendUrl}/powersync/token`, {
```
ital0: prefer using `ky` over `fetch`, per AGENTS.md definition.

src/db/powersync/connector.ts
```diff
@@ -0,0 +1,130 @@
...
+        op: op.op.toUpperCase() as 'PUT' | 'PATCH' | 'DELETE',
+        type: op.table,
+        id: op.id,
+        data: op.opData,
+      }))
+
+      console.info(`Uploading ${operations.length} operations to backend`)
+
+      // Send to backend
+      const response = await fetch(`${this.backendUrl}/powersync/upload`, {
```
ital0: prefer using `ky` over `fetch`, per AGENTS.md definition.

ital0 (changes requested):
Beautiful work here! I just found a bug and a few small details. 

Looking forward to seeing this merged 🚀

src/settings/preferences.tsx
raivieiraadriano92: fixed

src/lib/platform.ts
raivieiraadriano92: I'd say it's ok once the database will be encrypted soon - so nobody will be able to read the user's data even us.

src/settings/preferences.tsx
raivieiraadriano92: good catch - it's fixed

raivieiraadriano92: good catch - it's fixed

src/components/sign-in/use-sign-in-form-state.ts
```diff
@@ -142,9 +166,12 @@ export const useSignInFormState = ({
 
       // Store the token for bearer auth
       if (result.data?.token) {
-        await setAuthToken(result.data.token)
+        setAuthToken(result.data.token)
       }
 
+      const isNewUser = (result.data?.user as { isNew?: boolean })?.isNew ?? false
+      await onSignInSuccess(isNewUser)
```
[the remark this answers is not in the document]
raivieiraadriano92: The `isNew` is a new column we added to `user` table which we manually set to `false` when the user sign in for the first time

src/lib/platform.ts
ital0: makes sense

ital0 (approved):
just take a quick look on the Claude/Cursor comments (the last ones) to see if what is reported worth to get fixed. LGTM!

PR #362, opened 2026-02-05
THU-180: Improve Link Preview (image loading, better failure state UX)

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Adds new backend endpoints that fetch and proxy external images and tightens URL validation, which affects network I/O and introduces SSRF/size/timeout handling that must be correct to avoid abuse or regressions.
> 
> **Overview**
> Improves link previews by changing metadata extraction to **only** fall back to `<title>`/`<meta name="description">` when at least one Open Graph tag is present, returning all `null`s for captcha/block pages to avoid misleading previews; it also tightens URL parsing (no-empty URL, avoid double query-string appends, and restricts to `http/https`).
> 
> Adds image proxying capabilities via new backend routes `GET /link-preview/image/*` (fetch page → extract `og:image` → fetch image) and `GET /link-preview/proxy-image/*` (fetch image directly), with **SSRF protections**, 2MB size limits, and 2s image timeouts, and updates frontend link preview rendering to always show a consistent minimal card using hostname fallbacks and to load images through the new `proxy-image` endpoint. Extensive new test coverage validates proxying behavior, timeouts, size checks, and SSRF blocking (including IPv6 cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c6202cdc3eb56a00bbe87aaf72022cafe5f4b82a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/pro/link-preview.test.ts
+++ b/backend/src/pro/link-preview.test.ts
@@ -24,20 +24,6 @@ describe('Link Preview Routes', () => {
     return new Response(html, defaultOptions)
   }
 
-  const createMockImageResponse = (contentType = 'image/png') =>
-    new Response(new Uint8Array([0x89, 0x50, 0x4e, 0x47]), {
-      status: 200,
-      headers: { 'content-type': contentType },
-    })
-
-  /** Mock fetch that returns HTML for the target URL and a small image for any other URL */
-  const createPageAndImageMock = (targetUrl: string, html: string) => {
-    return (url: string) => {
-      if (url === targetUrl) return Promise.resolve(createMockHtmlResponse(html))
-      return Promise.resolve(createMockImageResponse())
-    }
-  }
-
   beforeAll(async () => {
     consoleSpies = setupConsoleSpy()
 
@@ -98,7 +84,7 @@ describe('Link Preview Routes', () => {
         </html>
       `
 
-      mockFetch.mockImplementation(createPageAndImageMock(targetUrl, html))
+      mockFetch.mockImplementation(() => Promise.resolve(createMockHtmlResponse(html)))
 
       const response = await app.handle(
         new Request(`http://localhost/link-preview/${targetUrl}`, {
@@ -113,7 +99,6 @@ describe('Link Preview Routes', () => {
       expect(body.data?.title).toBe('Test Article')
       expect(body.data?.description).toBe('This is a test article')
       expect(body.data?.image).toBe('https://example.com/image.jpg')
-      expect(body.data?.imageData).toMatch(/^data:image\/png;base64,/)
     })
 
     it('should fallback to title tag if og:title is missing', async () => {
@@ -172,7 +157,7 @@ describe('Link Preview Routes', () => {
         </html>
       `
 
-      mockFetch.mockImplementation(createPageAndImageMock(targetUrl, html))
+      mockFetch.mockImplementation(() => Promise.resolve(createMockHtmlResponse(html)))
 
       const response = await app.handle(new Request(`http://localhost/link-preview/${targetUrl}`, { method: 'GET' }))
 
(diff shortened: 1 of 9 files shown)


cjroth (changes requested):
1. Images Cause Height Issue:

Screenshot 2026-02-05 at 5 29 19 PM

The url at the bottom is causing this white space below the image. I think we don't need to add the url there - or at least should increase image height to support two lines of description text + url below.

2. imageData not present in any of the links that I'm seeing responses for. I think responding with raw binary data in JSON is also kind of funky - I was picturing this as a new API endpoint that returns in the image mime type rather than trying to embed it into the response JSON. It would still be 2 API calls, but they would happen in parallel instead of series.

3. I was thinking we could use the page title from Exa search results (the results we already get) instead of the page content. Right now the title for each link must wait on the page to get fetched (which can fail) but since we already know the title from the exa search results, we could cache it and reuse it somehow. If that's too complicated, then I think this is fine, but I think it should be possible.

cjroth (changes requested):
Right now the link preview widget has a waterfall: it first fetches metadata (title, description, image URL), waits for the response, then constructs an  src pointing to the proxy-image endpoint to load the image. That round trip between the two requests adds unnecessary latency before the user sees the preview image.
The fix is to use the /image/* endpoint instead of /proxy-image/*. The frontend should fire both requests in parallel at mount time: the metadata request for the text content, and the /image/* request for the image bytes. The /image/* endpoint already handles the full pipeline — it fetches the page, extracts the og:image URL, fetches the image, and returns the bytes. So the  src can be constructed immediately from the page URL without waiting for the metadata response. Yes, this means the backend fetches the page HTML twice, but that's a fast server-to-server call and the origin will likely cache it. The user-facing win of removing the waterfall is worth it.
Once that's done, the proxy-image endpoint and its tests can be removed since nothing will use it anymore. Also while you're in there — the twitter:image fallback was dropped from extractMetadata in this PR, which looks unintentional. The old code fell back to <meta name="twitter:image"> when og:image was missing, and some sites only have Twitter Card tags. That should be added back.

Concerns
Significant
1. No SSRF protection on the main metadata endpoint. validateImageUrl is only applied to the image proxy endpoints, but the main GET /link-preview/* endpoint happily fetches any URL including http://169.254.169.254/latest/meta-data/ or internal services. This is the more dangerous vector since the response body (HTML) is parsed and partially returned. An attacker could probe internal services and potentially extract information via the title/description fields.
2. No size limit on HTML page fetches. Both the metadata endpoint and the page-fetch step in /image/* call response.text() without any size check. A malicious URL returning a multi-gigabyte response could cause memory exhaustion. Consider adding a Content-Length check and/or streaming with a limit, similar to what fetchAndProxyImage does.
3. twitter:image fallback silently dropped. The old code fell back to <meta name="twitter:image"> when og:image was absent. The new code removes this entirely. Some sites (especially older ones) only have Twitter Card metadata, not Open Graph. This means those pages will lose their preview images without any obvious reason. Seems unintentional.
Minor
4. /image/* endpoint appears unused. The frontend uses proxy-image (passing the image URL directly from metadata). The /image/* endpoint — which re-fetches the page HTML to extract the image — isn't called anywhere. It adds surface area and duplicates the page-fetching logic. If it's intended for future use, it might be worth noting; otherwise it could be removed.
5. hostname.includes('metadata') is overly broad. This blocks any domain containing the substring "metadata" anywhere (e.g., metadata-api.example.com, docs.mymetadata.io). This could produce false positives for legitimate domains.
6. Duplicate URL parsing boilerplate. The URL extraction pattern (split path, decodeURIComponent, query string handling, protocol validation) is repeated nearly identically across all three route handlers. Extracting this into a shared helper would reduce duplication and make future changes less error-prone.
7. let usage for contentType and imageUrl. Per the project style guide, const is preferred. contentType could use a helper function with early returns; imageUrl in the /image/* handler being let ... = null and assigned in a nested try block is a bit awkward (though functionally safe since all error paths return before it's used).

cjroth:
Replaced with #373

PR #364, opened 2026-02-05
THU-236: Add a source citations feature

ital0:
> [!NOTE]
> **Medium Risk**
> Touches both inference request handling and core chat rendering/tooling paths; errors could break streaming output formatting or citation rendering, but changes are well-covered by targeted tests and include URL-safety validation.
> 
> **Overview**
> **Adds a new source-citations feature across the stack.** Tool executions now collect a capped, de-duplicated source registry (with stable `[Source N]`/`[N]` indices) that is attached to streamed `UIMessageMetadata`, and the UI renders inline, clickable citation badges and source lists from `{{CITE:N}}` placeholders.
> 
> This updates prompts and mode instructions to require `[N]` inline citations, enhances widget parsing (single-quoted JSON attrs + stripping model-native `【...】` citations), adds link-preview `source` support plus link-preview deduplication, and introduces URL safety/proxy utilities for links and favicons.
> 
> Separately, the inference route now sanitizes incoming chat messages by downgrading injected `system`/`developer` roles beyond the first message, with new tests covering the behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 460f876c0b71cc6f41570e92788985240ac5faf7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/inference/routes.ts
+++ b/backend/src/inference/routes.ts
@@ -6,6 +6,14 @@ import { Elysia } from 'elysia'
 import { APIConnectionError, APIConnectionTimeoutError } from 'openai'
 import { getInferenceClient, type InferenceProvider } from './client'
 
+type Message = { role: string; content: unknown }
+
+const privilegedRoles = new Set(['developer', 'system'])
+
+/** Downgrade developer/system roles to user for all messages except the first (the legitimate system prompt). */
+const sanitizeMessageRoles = (messages: Message[]): Message[] =>
+  messages.map((msg, i) => (i > 0 && privilegedRoles.has(msg.role) ? { ...msg, role: 'user' } : msg))
+
 type ModelConfig = {
   provider: InferenceProvider
   internalName: string
@@ -59,7 +67,7 @@ export const createInferenceRoutes = () => {
       try {
         const completion = await (client as PostHogOpenAI).chat.completions.create({
           model: internalName,
-          messages: body.messages,
+          messages: sanitizeMessageRoles(body.messages),
           temperature: body.temperature,
           tools: body.tools,
           tool_choice: body.tool_choice,
--- a/backend/src/inference/routes.ts
+++ b/backend/src/inference/routes.ts
@@ -4,6 +4,7 @@ import { createSSEStreamFromCompletion } from '@/utils/streaming'
 import type { OpenAI as PostHogOpenAI } from '@posthog/ai'
 import { Elysia } from 'elysia'
 import { APIConnectionError, APIConnectionTimeoutError } from 'openai'
+import type { ChatCompletionMessageParam } from 'openai/resources/chat/completions'
 import { getInferenceClient, type InferenceProvider } from './client'
 
 type Message = { role: string; content: unknown }
@@ -67,7 +68,7 @@ export const createInferenceRoutes = () => {
       try {
         const completion = await (client as PostHogOpenAI).chat.completions.create({
           model: internalName,
-          messages: sanitizeMessageRoles(body.messages),
+          messages: sanitizeMessageRoles(body.messages) as ChatCompletionMessageParam[],
           temperature: body.temperature,
           tools: body.tools,
           tool_choice: body.tool_choice,
--- a/src/ai/widget-parser.test.ts
+++ b/src/ai/widget-parser.test.ts
@@ -685,25 +685,25 @@ describe('parseContentParts', () => {
       expect(result[1].type).toBe('widget')
     })
 
-    it('strips brackets with various content formats', () => {
-      const text = 'Fact【source_name】 and another【12】 and【title with spaces】.'
+    it('strips numbered bracket citations', () => {
+      const text = 'Fact【12】 and another【3】.'
       const result = parseContentParts(text)
 
-      expect(result).toEqual([{ type: 'text', content: 'Fact and another and.' }])
+      expect(result).toEqual([{ type: 'text', content: 'Fact and another.' }])
     })
 
-    it('strips bracket with trailing period', () => {
+    it('strips bracket with dagger and title', () => {
       const text = 'The EU passed the AI Act【6†title】.'
       const result = parseContentParts(text)
 
       expect(result).toEqual([{ type: 'text', content: 'The EU passed the AI Act.' }])
     })
 
-    it('strips orphan opening bracket without closing', () => {
-      const text = '$2,499 USD 【'
+    it('preserves legitimate CJK brackets', () => {
+      const text = '価格は【税込み】です'
       const result = parseContentParts(text)
 
-      expect(result).toEqual([{ type: 'text', content: '$2,499 USD' }])
+      expect(result).toEqual([{ type: 'text', content: '価格は【税込み】です' }])
     })
 
     it('preserves text when no bracket citations present', () => {
--- a/src/ai/widget-parser.test.ts
+++ b/src/ai/widget-parser.test.ts
@@ -658,6 +658,42 @@ describe('parseContentParts', () => {
       expect(result[2]).toEqual({ type: 'text', content: 'Fact.' })
       expect((result[3] as { type: 'widget'; widget: { widget: string } }).widget.widget).toBe('citation')
     })
+
+    it("handles apostrophe in JSON title (NASA's Mission)", () => {
+      const json = `[{"id":"1","title":"NASA's Mission","url":"https://nasa.gov","siteName":"NASA"}]`
+      const text = `Space exploration advances. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
+      expect(result).toHaveLength(2)
+      expect(result[1]).toEqual({
+        type: 'widget',
+        widget: { widget: 'citation', args: { sources: json } },
+      })
+    })
+
+    it("handles multiple apostrophes in JSON (Rock 'n' Roll)", () => {
+      const json = `[{"id":"1","title":"Rock 'n' Roll History","url":"https://music.com","siteName":"Music DB"}]`
+      const text = `Music evolved. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
+      expect(result).toHaveLength(2)
+      expect(result[1]).toEqual({
+        type: 'widget',
+        widget: { widget: 'citation', args: { sources: json } },
+      })
+    })
+
+    it("handles apostrophe in siteName (McDonald's)", () => {
+      const json = `[{"id":"1","title":"Earnings Report","url":"https://mcdonalds.com","siteName":"McDonald's"}]`
+      const text = `Revenue grew. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
+      expect(result).toHaveLength(2)
+      expect(result[1]).toEqual({
+        type: 'widget',
+        widget: { widget: 'citation', args: { sources: json } },
+      })
+    })
   })
 
   describe('bracket citation stripping', () => {
--- a/src/ai/widget-parser.test.ts
+++ b/src/ai/widget-parser.test.ts
@@ -632,4 +632,85 @@ describe('parseContentParts', () => {
       expect(result3).toEqual([{ type: 'text', content: 'Some text' }])
     })
   })
+
+  describe('single-quoted attributes', () => {
+    it('parses citation with single-quoted JSON sources', () => {
+      const json = '[{"id":"1","title":"Test","url":"https://example.com","siteName":"Example"}]'
+      const text = `Some fact. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
+      expect(result).toHaveLength(2)
+      expect(result[0]).toEqual({ type: 'text', content: 'Some fact.' })
+      expect(result[1]).toEqual({
+        type: 'widget',
+        widget: { widget: 'citation', args: { sources: json } },
+      })
+    })
+
+    it('parses link preview with double-quoted attributes alongside citation with single-quoted', () => {
+      const json = '[{"id":"1","title":"Test","url":"https://a.com"}]'
+      const text = `Link: <widget:link-preview url="https://b.com" /> Fact. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
+      expect(result).toHaveLength(4)
+      expect(result[0]).toEqual({ type: 'text', content: 'Link:' })
+      expect((result[1] as { type: 'widget'; widget: { widget: string } }).widget.widget).toBe('link-preview')
+      expect(result[2]).toEqual({ type: 'text', content: 'Fact.' })
+      expect((result[3] as { type: 'widget'; widget: { widget: string } }).widget.widget).toBe('citation')
+    })
+  })
+
+  describe('bracket citation stripping', () => {
+    it('strips OpenAI-style bracket citations', () => {
+      const text = 'The AI Act was passed【2†title】.'
+      const result = parseContentParts(text)
+
+      expect(result).toEqual([{ type: 'text', content: 'The AI Act was passed.' }])
+    })
+
+    it('strips multiple bracket citations', () => {
+      const text = 'First fact【1†source】 and second fact【3†source】.'
+      const result = parseContentParts(text)
+
+      expect(result).toEqual([{ type: 'text', content: 'First fact and second fact.' }])
+    })
+
+    it('strips bracket citations alongside valid widget citations', () => {
+      const json = '[{"id":"1","title":"Test","url":"https://example.com"}]'
+      const text = `AI regulation passed【2†title】 today. <widget:citation sources='${json}' />`
+      const result = parseContentParts(text)
+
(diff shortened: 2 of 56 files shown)


src/components/chat/memoized-markdown.tsx
```diff
@@ -1,35 +1,43 @@
...
 MemoizedMarkdownBlock.displayName = 'MemoizedMarkdownBlock'
 
-export const MemoizedMarkdown = memo(({ content, id }: { content: string; id: string }) => {
+type MemoizedMarkdownProps = {
+  content: string
+  id: string
+  citations?: CitationMap
+}
+
+export const MemoizedMarkdown = memo(({ content, id, citations }: MemoizedMarkdownProps) => {
```
cjroth: I'm a little skeptical of adding specific functionality for citations in this function since it's a pretty general-purpose function - tightly coupling the logic around citation feels funky to me. If citations are just widgets, should all widgets be included here?

cjroth: Separately, I suspect this memoized markdown thing is related to the citations not being clickable when streaming is in progress

src/components/chat/source-card.tsx
```diff
@@ -0,0 +1,86 @@
...
+  /** Base URL for the proxy endpoint (e.g., "http://localhost:8000/v1") to bypass COEP */
+  proxyBase?: string
+}
+
+/**
+ * Generates a color for the initial badge based on the site name
+ */
+const getBadgeColor = (siteName: string = '') => {
+  const colors = [
+    'bg-[#7b8fff]', // blue
```
cjroth: I think we should find a way to pull these colors from our tailwind theme's colors so that they match our pallette

src/components/chat/source-card.tsx
```diff
@@ -0,0 +1,86 @@
...
+    'bg-[#e11d48]', // red
+    'bg-[#8b5cf6]', // purple
+    'bg-[#f59e0b]', // amber
+  ]
+  const index = (siteName.charCodeAt(0) || 0) % colors.length
+  return colors[index]
+}
+
+/** Derives a favicon URL from the source domain, proxied through the backend to bypass COEP */
+const getFaviconUrl = (pageUrl: string, proxyBase?: string): string | null => {
```
cjroth: Don't we already have some logic for this somewhere, such as when it's fetching websites?

src/lib/citation-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  } catch {
+    return null
+  }
+}
+
+/**
+ * Validates that a URL uses a safe protocol (http or https).
+ * Returns false for javascript:, data:, and other potentially dangerous schemes.
+ */
+export const isSafeUrl = (url: string): boolean => {
```
cjroth: Interesting. I guess this is to protect against some kind of prompt injection, right?

We should probably move this outside of the citations logic and apply this in our markdown links to right?

cjroth (changes requested):
This is a good start and the UI components look great.

After reviewing carefully, I want to do a little bit more exploration around the architecture of this feature. My main concern with the current approach is that the model needs to generate the same JSON repeatedly despite it being the same source that's being cited throughout the response.

I'd like to find a way to allow the AI to respond with something like:

`According to Wired, the best robot vacuum is Matic [1]`

Where `[1]` might refer to the index of a source that was returned from the "search" tool.

If we could store the list of sources that the search tool finds in a list and refer to them, we would solve that.

ALSO if we could store the list of sources, including their titles and favicons, and refer back to them, it would simultaneously solve some performance issues with our link previews widget where it's having to re-generate the URL for the website, favicon, etc every time.

It's probably worth talking through this in a call.

src/components/chat/source-card.tsx
ital0: Done. Replaced the hardcoded hex colors with `bg-chart-1` through `bg-chart-5` theme tokens.

src/lib/citation-utils.ts
ital0: That's correct. 

I’m changing this here so it can be reused in the links rendered by `react-markdown`.

src/components/chat/memoized-markdown.tsx
ital0: Well, the logic behind this change was because `Citations` are the only widget that needs to render inline within markdown text, after a sentence, inside a paragraph.

The other widgets renders as a block-level element between paragraphs.

But I agree with your observation about the coupling. Let me see what are the alternatives.

ital0:
> This is a good start and the UI components look great.
> 
> After reviewing carefully, I want to do a little bit more exploration around the architecture of this feature. My main concern with the current approach is that the model needs to generate the same JSON repeatedly despite it being the same source that's being cited throughout the response.
> 
> I'd like to find a way to allow the AI to respond with something like:
> 
> `According to Wired, the best robot vacuum is Matic [1]`
> 
> Where `[1]` might refer to the index of a source that was returned from the "search" tool.
> 
> If we could store the list of sources that the search tool finds in a list and refer to them, we would solve that.
> 
> ALSO if we could store the list of sources, including their titles and favicons, and refer back to them, it would simultaneously solve some performance issues with our link previews widget where it's having to re-generate the URL for the website, favicon, etc every time.
> 
> It's probably worth talking through this in a call.

@cjroth that’s true. During the implementation I noticed this repetition of the same widget multiple times. Apart from the multiple calls to fetch the favicon's as well...

I even thought about putting all the citations in a final footer with some kind of reference, but that would drift quite a bit from the suggested layout.

Let me see what I can get before we talk.

src/components/chat/memoized-markdown.tsx
ital0: I did some research and found a few viable alternatives, and the one that turned out to be the least complex was this one, passing the components via props.

The other alternatives were:

* Creating a custom plugin — an extremely complex alternative that would also require a `ref` to avoid re-rendering the components;
* Creating a `CitationProvider` in the context tree and then “capturing” all the `citations` and replacing them with the widget — extremely complex, and it would be very difficult to avoid re-rendering the other consumers, with a big performance impact.

cjroth:
> @cjroth that’s true. During the implementation I noticed this repetition of the same widget multiple times. Apart from the multiple calls to fetch the favicon's as well...

Multiple calls to fetch the favicon is fine - the browser should handle caching this.

The main issue is that the LLM is outputting a large <widget /> tag with JSON in it repeatedly, which is slow and uses a lot of tokens.

> I even thought about putting all the citations in a final footer with some kind of reference, but that would drift quite a bit from the suggested layout.

This is close to what I am suggesting in https://linear.app/mozilla-thunderbolt/issue/THU-292/search-results-caching-idea. My proposal here is to have the LLM output references like `[1]` but then intercept and replace all of the occurrences of `[n]` (`[1]`, `[2]`, etc) in the markdown as it is streaming with the React component that you've made here. Instead of parsing out JSON from a bunch of `<widget data={...} />` tags, the widget components would have logic that obtains the JSON data from the previous search tool call responses.

Basically, instead of the LLM needing to output the long string `<widget data={...} />` repeatedly and using tons of tokens for this, it would only need to output `[1]` repeatedly (much fewer tokens) and then we could update our code to replace these `[1]`s with real widgets.

Does this make sense?

src/components/chat/widget-renderer.tsx
```diff
@@ -41,6 +43,7 @@ export const WidgetRenderer = memo(({ widget, messageId }: WidgetRendererProps)
     {
       ...widget.args,
       messageId,
+      sources,
```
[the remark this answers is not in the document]
ital0: Both triggering scenarios require the AI to behave in ways it's explicitly instructed not to, in combinations that are extremely unlikely.

src/components/chat/citation-popover.tsx
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Provides citation popover state and renders the overlay (Popover/Sheet)
+ * outside the markdown tree so streaming re-renders don't destroy it.
+ */
+export const CitationPopoverProvider = ({ children }: { children: ReactNode }) => {
+  const [state, setState] = useState<{
+    id: number | null
+    sources: CitationSource[] | null
+    rect: DOMRect | null
+  }>({ id: null, sources: null, rect: null })
```
raivieiraadriano92: I noticed that when we open/close this popover component a re-render is triggered for the entire message  component + for each citation component in that message.

See the video below:
https://github.com/user-attachments/assets/9934d747-339e-4acf-9277-46f3bc80f853

Although this is re-renders alone are not a big performance bottleneck I'd say we could try to avoid it since we are trying to improve the chat performance.

One suggestion here would be moving the state + open/close method from the `CitationPopoverProvider` to `CitationOverlay` component - and use `forwardref` + `useimperativehandle` to set the make the open/close methods available.

Something like we have in `src/components/delete-all-chats-dialog.tsx`

Let me know your thoughts.

backend/src/inference/routes.ts
```diff
@@ -59,7 +68,7 @@ export const createInferenceRoutes = () => {
       try {
         const completion = await (client as PostHogOpenAI).chat.completions.create({
           model: internalName,
-          messages: body.messages,
+          messages: sanitizeMessageRoles(body.messages) as ChatCompletionMessageParam[],
```
raivieiraadriano92, line 71: Do you think it'd make sense to change the return type of `sanitizeMessageRoles` to `ChatCompletionMessageParam[]` instead of forcing a type casting?

backend/src/inference/routes.ts
```diff
@@ -4,8 +4,17 @@ import { createSSEStreamFromCompletion } from '@/utils/streaming'
 import type { OpenAI as PostHogOpenAI } from '@posthog/ai'
 import { Elysia } from 'elysia'
 import { APIConnectionError, APIConnectionTimeoutError } from 'openai'
+import type { ChatCompletionMessageParam } from 'openai/resources/chat/completions'
 import { getInferenceClient, type InferenceProvider } from './client'
 
+type Message = { role: string; content: unknown }
```
raivieiraadriano92, line 10: Not sure we should do it in this PR - but I think this could be an use case where we can start sharing the application types between FE and BE using a `shared/` folder. I started that folder in the PowerSync PR.

Feel free to implement this or not if it makes sense.

src/ai/widget-parser.test.ts
```diff
@@ -632,4 +632,173 @@ describe('parseContentParts', () => {
       expect(result3).toEqual([{ type: 'text', content: 'Some text' }])
     })
   })
+
+  describe('single-quoted attributes', () => {
+    it('parses citation with single-quoted JSON sources', () => {
+      const json = '[{"id":"1","title":"Test","url":"https://example.com","siteName":"Example"}]'
```
raivieiraadriano92, line 638: This is not an issue and I don't think it should be addressed in this PR ... just a suggestion for future improvements.

Maybe we should start considering replace our `jsons` with `toon` format which can help us save tokens while working with LLMs.

At this moment we have just a few widgets which transports json data but as the application grows this could have a huge impact in the amount of tokens used.

Some references:
https://medium.com/data-science-in-your-pocket/toon-bye-bye-json-for-llms-91e4fe521b14
https://www.npmjs.com/package/@jojojoseph/toon-json-converter

src/components/chat/citation-badge.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+// --- Context-managed variant (inline in streaming markdown) ---
+
+const ManagedBadge = memo(({ sources, citationId }: { sources: CitationSource[]; citationId: number }) => {
+  const ctx = useCitationPopover()!
+  const isOpen = ctx.popover?.citationId === citationId
+  const { displayName, additionalCount, ariaLabel } = getBadgeLabel(sources)
+
+  const toggle = (rect: DOMRect) => {
+    if (isOpen) ctx.close()
```
raivieiraadriano92, line 54: maybe we should add two new rules in our eslint.config file to keep the code style consistent:

```
// Enforce brackets after if statements
curly: ['error', 'all'],
// Prefer early returns instead of else statements
'no-else-return': ['error', { allowElseIf: false }],
```

This aligns with the code style `Prefer early return over long if statements and nested code` that we have defined in AGENTS.md and CLAUDE.md.

src/components/chat/markdown-utils.tsx
```diff
@@ -1,18 +1,38 @@
...
+ * new component types on each render. Components read from this context
+ * to render inline CitationBadge elements.
+ */
+export const CitationContext = createContext<CitationMap | undefined>(undefined)
+
 /**
  * Regex to split on <br> tags (case-insensitive, global)
  */
-const BR_SPLIT_REGEX = /<br\s*\/?>/gi
+const brSplitRegex = /<br\s*\/?>/gi
```
raivieiraadriano92, line 21: sometimes we forget this patter - to avoid this we could enforce the naming convention adding the below rule to our eslint.config file:

```
'@typescript-eslint/naming-convention': [
        'error',
        // Enforce camelCase for variables (const, let)
        {
          selector: 'variable',
          format: ['camelCase'],
          leadingUnderscore: 'allow', // Allow _unused variables
        },
        // Allow PascalCase for React components
        {
          selector: 'variable',
          modifiers: ['const'],
          format: ['camelCase', 'PascalCase'],
          filter: {
            // Allow React components (functions that return JSX)
            regex: '^[A-Z]',
            match: true,
          },
        },
        // Enforce PascalCase for types, interfaces, classes
        {
          selector: 'typeLike',
          format: ['PascalCase'],
        },
        // Enforce camelCase for functions
        {
          selector: 'function',
          format: ['camelCase', 'PascalCase'], // PascalCase for React components
        },
      ],
```

src/integrations/thunderbolt-pro/tools.ts
```diff
@@ -28,27 +30,96 @@ export {
...
+      verb: 'fetching {url}',
+      parameters: fetchContentSchema,
+      execute: async (params: FetchContentParams) => {
+        const result = await fetchContent(params, httpClient)
+
+        if (!result) return result
+
+        const existingSource = sourceCollector?.find((s) => s.url === result.url)
+        const sourceIndex = existingSource ? existingSource.index : nextIndex
+
```
raivieiraadriano92, line 91: I see we have some repeated code in `fetch_content` and `search` - would it be possible to extract that into a new method to reuse the logic?

raivieiraadriano92 (changes requested):
I left a few comments with some improvement suggestions - but other than that it seems really good, congrats. This new component brings a really nice UX while chatting.

A few points a caught while testing the deploy preview that worth to mention:

GPT OSS: the model didn't use the citations widget even after many tries - I tried to explicitly say to give me the source citations and it showed the Preview Link widget instead of the Citations one.

Mistral Medium 3.1: it worked at the first try but I noticed something weird when I explicitly asked for the source citations:

Image

Basically it keeps `[1]`, [2] and so on at the end - I know I forced the model a little bit here haha - so I'm not sure if it worth to investigate this. 

Also I noticed all citations were grouped at the end of the message:

Image

Sonet 4.5: it was the model with the best results - it worked at the first try and added citation after phrases and paragraphs in the message.

backend/src/inference/routes.ts
ital0: Good catch.

The cast is intentional because the function works with a looser Message type to keep it SDK-agnostic. Making it return `ChatCompletionMessageParam[]` would couple the sanitizer to the OpenAI SDK.

src/components/chat/citation-badge.tsx
ital0: Good idea for consistency! Since you suggested some other similar changes, might worth creating a specific for this and try to address separate.

src/integrations/thunderbolt-pro/tools.ts
ital0: Good eye on the duplication. I looked into extracting it, but the two paths diverge significantly. `fetch_content` updates existing source metadata (title, image, etc.) while `search` doesn't. A shared helper would need conditional logic that arguably obscures the differences. If we add a third tool with source registration, that'd be the right time to extract. But let me know your thoughts and understand if it really makes sense to extract!

src/integrations/thunderbolt-pro/tools.ts
```diff
@@ -28,27 +30,96 @@ export {
...
+              author: result.author,
+              publishedDate: result.publishedDate,
+              toolName: 'search',
+            })
+            nextIndex++
+          } else if (!existingSource) {
+            nextIndex++
+          }
+
+          return { sourceLabel: `[Source ${sourceIndex}] (cite as [${sourceIndex}])`, sourceIndex, ...result }
```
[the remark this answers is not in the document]
ital0: `sourceIndex` is never present on the raw API response. It exists on the type to describe the enriched return shape. Swapping the spread order would be defensive coding against a scenario that can't happen. Per our `AGENTS.md`, we prefer optimistic code.

raivieiraadriano92 (approved):

PR #371, opened 2026-02-17
Prompt Separation by Model/Chat Mode

ital0:
> [!NOTE]
> **Medium Risk**
> Touches the core AI request pipeline (prompt construction, step/retry behavior, provider options) and introduces a new persisted `model_profiles` table, so misconfiguration could change response quality or runtime behavior across models/modes. Changes are scoped but impact user-facing generation and default-data reconciliation.
> 
> **Overview**
> Adds a new per-model configuration layer (`model_profiles`) with defaults + hashing, reconciliation, and a DAL for upsert/reset/soft-delete; model creation now auto-seeds a default profile when available and model deletion also soft-deletes its profile.
> 
> Updates `aiFetchStreamingResponse` and prompt/step logic to use the model profile + explicit `modeName` to drive system prompt overrides (tools/link preview addenda), inference settings (temperature/steps/attempts/nudge threshold), provider options, nudges, and retry handling (retry now considers tool calls across attempts and escalates the retry nudge).
> 
> Introduces a standalone AI eval runner (`bun run eval`) with scenario generation, stream parsing, scoring, terminal UI, and markdown reporting, plus a single-scenario debug script; also factors mode prompts into separate files and makes minor docs/.gitignore/platform guard tweaks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 005a7a7ad94e73b0ce9742fa238eb408642b0af4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/ai/prompts/vendors/mistral/search.ts
```diff
@@ -0,0 +1,9 @@
+import type { PromptOverride } from '../../types'
+
+/** Mistral search mode — URL validation to prevent re-linking to homepages/sections */
```
cjroth: This approach feels funky. I suppose it's a valid way to prevent index pages, but really what we want the model to care about is the actual page content - the URL really should not matter here. It seems like this is fixing it in kind of the wrong way.

src/ai/prompts/vendors/openai/models/gpt-oss-120b/config.ts
```diff
@@ -0,0 +1,17 @@
+import type { PartialVendorConfig } from '../../../../types'
+
+/**
+ * GPT-OSS 120B model-specific config — tuned through E2E eval testing.
+ *
+ * These override the vendor-level OpenAI config:
```
cjroth: This is a really interesting idea to have model-specific tuning like this.

src/ai/prompts/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import { mistralResearchOverride } from './vendors/mistral/research'
+import { mistralSearchOverride } from './vendors/mistral/search'
+import { openaiChatOverride } from './vendors/openai/chat'
+import { openaiVendorConfig } from './vendors/openai/config'
+import { openaiGlobalOverride } from './vendors/openai/global'
+import { gptOss120bConfig } from './vendors/openai/models/gpt-oss-120b/config'
+import { openaiResearchOverride } from './vendors/openai/research'
+import { openaiSearchOverride } from './vendors/openai/search'
+
+/** Vendor-level partial configs — shared by all models from that vendor */
```
cjroth: I'm not sure if I see the use case for vendor overrides. I definitely understand model overrides - but do all of the models from the same vendor really tend to need shared overrides?

.gitignore
```diff
@@ -36,6 +36,9 @@ tauri.key
 *storybook.log
 storybook-static
 
+# Eval results
+.evals/
```
cjroth: I kind of like the idea of keeping this as a visible folder (no `.` prefix) - kind of like `node_modules`. So still gitignore, but don't hide.

cjroth (changes requested):

src/ai/prompts/vendors/mistral/chat.ts
```diff
@@ -0,0 +1,12 @@
+import type { PromptOverride } from '../../types'
```
cjroth: Here and everywhere else in this codebase - we should use `@/types` instead of `../../types`. This makes it more readable and cleaner

cjroth (changes requested):

src/ai/eval/scenarios.ts
```diff
@@ -0,0 +1,321 @@
+import { defaultModelGptOss120b, defaultModelMistralMedium31, defaultModelSonnet45 } from '@/defaults/models'
+import type { EvalCriteria, EvalScenario } from './types'
+
+const MODELS = [
```
raivieiraadriano92: maybe use lowerCase?

src/ai/eval/scoring.ts
```diff
@@ -0,0 +1,156 @@
+import type { EvalCriteria, EvalResult, EvalScenario, ParsedStream } from './types'
+
+const REVIEW_SITE_DOMAINS = [
```
raivieiraadriano92: same here, maybe we should keep lowerCase

src/ai/eval/scoring.ts
```diff
@@ -0,0 +1,156 @@
...
+  '/health',
+  '/health/',
+  '/sections',
+  '/hub',
+  '/category',
+  '/categories',
+  '/topics',
+  '/tags',
+])
+
```
raivieiraadriano92, line 52: same here SECTION_PATHS and GENERIC_SUBDOMAINS

src/ai/eval/ui.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/** Move cursor to absolute row, col */
+const moveTo = (row: number, col = 1) => write(`\x1b[${row};${col}H`)
+
+const SPINNER_FRAMES = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏']
+const FIXED_BASE_LINES = 3 // separator + progress + stats
+const HEADER_LINES = 5 // title + separator + info + separator + blank
+const BAR_WIDTH = 30
+const PROMPT_MAX_WIDTH = 48
+const ID_WIDTH = 18
```
raivieiraadriano92: same here

src/lib/platform.ts
```diff
@@ -20,6 +20,7 @@ export const isPrPreview = (): boolean => {
  * @returns true if running in Tauri, false otherwise
  */
 export const isTauri = (): boolean => {
+  if (typeof window === 'undefined') return false
```
raivieiraadriano92, line 23: just want to confirm - these new `if` statements are necessary for the eval calls, is it right?

raivieiraadriano92 (changes requested):
It looks great and is pretty cool to see the eval running. I left just a few comment.

Also, I'm not sure if it makes sense or not - but should we add tests for the eval methods?

src/ai/eval/scenarios.ts
ital0: You're right. CLAUDE.md says camelCase for consts. I'll rename MODELS to models (or evalModels to avoid shadowing). Same for the other UPPER_CASE constants in the eval files.

src/lib/platform.ts
ital0: Yes, exactly. The eval runner runs in Bun (no browser, no window). The existing typeof window === 'undefined' check in isTauri() already handles this, but the additional guard ensures we don't hit any edge cases in test environments where window might be partially defined (e.g., jsdom). Safe defensive check for a platform boundary.

src/ai/fetch.ts
```diff
@@ -223,17 +229,29 @@ export const aiFetchStreamingResponse = async ({
       ],
     })
 
-    const maxSteps = 20
-    const maxAttempts = 2
+    // Load inference config from profile (fallback to code defaults)
```
cjroth: These values could almost go into a typescript typed object like model config or something. Not that important, just an idea

src/ai/fetch.ts
```diff
@@ -251,21 +269,34 @@ export const aiFetchStreamingResponse = async ({
         providerOptions,
 
         prepareStep: ({ steps, stepNumber, messages: stepMessages }) => {
+          const hadToolCallSteps = steps.some((s) => s.finishReason === 'tool-calls')
```
cjroth: I feel like this prepareStep logic should have unit tests if possible - lots of branches in here

src/defaults/model-profiles.ts
```diff
@@ -0,0 +1,150 @@
+import { hashValues } from '@/lib/utils'
```
cjroth: This is going to be really hard to read and edit - and I think we'll be editing it fairly often. Would it be possible to create a similar folder structure to what we had where we define the prompts in folders and files and pull it into here?

Eg:

/prompts
  /mistral-medium-3.1.tsx

export const prompt = `
Prompt...
`

export const profile = {
  setting1: 1,
  setting2: 2
}


It would still import into this file and be used the same way as the defaults using the sqlite databse- just breaking into separate files to import.

cjroth (changes requested):
Added some suggestions

cjroth (approved):

src/ai/eval/debug-single.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
raivieiraadriano92, line 1: Is this code inside of `/src/ai/eval/*` required in prod?
If not I'd say we could skip that during prod build to avoid increase the bundle size.

PR #372, opened 2026-02-18
THU-191: New BE DB schemas for PowerSync

raivieiraadriano92:
> [!NOTE]
> **Medium Risk**
> Applies new database migrations and introduces new tables/constraints in a separate schema; main risk is migration/constraint correctness and downstream query expectations rather than changes to existing business logic.
> 
> **Overview**
> Introduces a new Postgres schema `powersync` and creates the initial set of PowerSync-synced tables (`chat_threads`, `chat_messages`, `tasks`, `settings`, `models`, `modes`, `prompts`, `triggers`, `devices`, `mcp_servers`) via Drizzle migrations, including per-user foreign keys (cascade delete), supporting indexes, and a self-referential parent FK for `chat_messages`.
> 
> Adds corresponding Drizzle table definitions in `backend/src/db/powersync-schema.ts` and re-exports them from `backend/src/db/schema.ts` so they can be used by the backend.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 88c3dc24d4ec33c640e65813d8a284d07634e4a8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/db/powersync-schema.ts
+++ b/backend/src/db/powersync-schema.ts
@@ -1,0 +1,194 @@
+import { index, integer, pgSchema, primaryKey, text, timestamp } from 'drizzle-orm/pg-core'
+import { user } from './auth-schema'
+
+/**
+ * PowerSync tables - mirror of frontend SQLite schema.
+ * These tables sync bidirectionally with the frontend via PowerSync.
+ */
+
+const powersyncSchema = pgSchema('powersync')
+
+export const settingsTable = powersyncSchema.table(
+  'settings',
+  {
+    // Column is named 'id' in DB for PowerSync compatibility, but accessed as 'key' in TypeScript
+    key: text('id').notNull(),
+    value: text('value'),
+    updatedAt: timestamp('updated_at').defaultNow(),
+    defaultHash: text('default_hash'),
+    userId: text('user_id')
+      .notNull()
+      .references(() => user.id, { onDelete: 'cascade' }),
+  },
+  (table) => [primaryKey({ columns: [table.key, table.userId] }), index('idx_settings_user_id').on(table.userId)],
+)
+
+export const chatThreadsTable = powersyncSchema.table(
+  'chat_threads',
+  {
+    id: text('id').primaryKey(),
+    title: text('title'),
+    isEncrypted: integer('is_encrypted').default(0),
+    triggeredBy: text('triggered_by'),
+    wasTriggeredByAutomation: integer('was_triggered_by_automation').default(0),
+    contextSize: integer('context_size'),
+    modeId: text('mode_id'),
+    deletedAt: timestamp('deleted_at'),
+    userId: text('user_id')
+      .notNull()
+      .references(() => user.id, { onDelete: 'cascade' }),
+  },
+  (table) => [index('idx_chat_threads_user_id').on(table.userId)],
+)
+
+export const chatMessagesTable = powersyncSchema.table(
+  'chat_messages',
+  {
+    id: text('id').primaryKey(),
+    content: text('content'),
+    role: text('role'),
+    parts: text('parts'),
+    chatThreadId: text('chat_thread_id').references(() => chatThreadsTable.id),
+    modelId: text('model_id'),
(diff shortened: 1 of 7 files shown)


backend/src/db/powersync-schema.ts
```diff
@@ -0,0 +1,308 @@
...
+    updatedAt: integer('updated_at').default(sql`extract(epoch from now())::integer`),
+    defaultHash: text('default_hash'),
+    userId: text('user_id')
+      .notNull()
+      .references(() => user.id, { onDelete: 'cascade' }),
+  },
+  (table) => [primaryKey({ columns: [table.key, table.userId] }), index('idx_settings_user_id').on(table.userId)],
+)
+
+export const chatThreadsTable = pgTable(
```
ital0: noticed the frontend table includes `modeId: text('mode_id')`, but the backend schema does not. Is that expected? I also made sure to check the other PR to understand if this is being addressed there and looks like it's not.

shared/powersync-tables.ts
```diff
@@ -0,0 +1,41 @@
...
+] as const
+
+export type PowerSyncTableName = (typeof powersyncTableNames)[number]
+
+/**
+ * Map of PowerSync table names to React Query keys to invalidate when the table changes.
+ * Keys are type-checked against powersyncTableNames; every table must have an entry.
+ * Prefix keys (e.g. ['settings']) invalidate all queries starting with that prefix.
+ */
+export const powersyncTableToQueryKeys: {
```
cjroth: Would this logic make sense to generate a utility function that just generates the names instead of hardwiring?

The downside here is (a) our code (and thus our bundle size) will grow as then number of tables grows and (b) we have to remember to update this every time we add or rename a table.

backend/src/db/powersync-schema.ts
```diff
@@ -0,0 +1,308 @@
...
+  modes: modesTable.id,
+  devices: devicesTable.id,
+}
+
+/**
+ * Conflict target for each PowerSync table (for INSERT ON CONFLICT).
+ * Tables with default data (settings, models, modes, tasks, prompts) use composite primary keys (id/key + user_id)
+ * so each user can have their own row with the same default ID. See docs/composite-primary-keys-and-default-data.md.
+ */
+export const powersyncConflictTarget: Record<PowerSyncTableName, AnyPgColumn[]> = {
```
cjroth: Would this logic make sense to generate a utility function that just generates the names instead of hardwiring?

The downside here is (a) our code (and thus our bundle size) will grow as then number of tables grows and (b) we have to remember to update this every time we add or rename a table.

(Also added comment below that is similar)

backend/src/db/powersync-schema.ts
```diff
@@ -0,0 +1,308 @@
...
+} from 'drizzle-orm/pg-core'
+import { getTableColumns, sql } from 'drizzle-orm'
+import { user } from './auth-schema'
+
+/**
+ * PowerSync tables - mirror of frontend SQLite schema.
+ * These tables sync bidirectionally with the frontend via PowerSync.
+ */
+
+export const settingsTable = pgTable(
```
ital0: noticed this table doesn't have `deleted_at`. Is that intentional?

ital0 (changes requested):
in general LGTM! Just left some minor details so you can take a look!

backend/src/db/powersync-schema.ts
ital0: I mean, the user can delete any setting? Or is this going to be allowed in the future?

backend/src/db/powersync-schema.ts
raivieiraadriano92: good catch - it was an error while rebasing - it's fixed

darkbanjo (approved):

shared/powersync-tables.ts
raivieiraadriano92: yes, I'll implement something there
I just removed that from this PR once it's not really necessary ... I'll handle that in the other branches

ital0 (approved):

PR #373, opened 2026-02-18
THU-180: Improve the link preview feature

ital0:
> [!NOTE]
> **High Risk**
> Introduces new network-facing proxy endpoints and changes URL validation/SSRF logic; mistakes could reintroduce SSRF or break preview fetching for legitimate URLs despite added test coverage.
> 
> **Overview**
> Improves the `/pro/link-preview` backend to add **SSRF protection** (blocks localhost/private IPs incl. IPv6/IPv4-mapped), robust URL decoding/validation, and **2MB HTML size limits**, while extending extracted metadata to include `og:site_name` and avoiding title/description fallbacks on pages with no social tags.
> 
> Adds new endpoints to serve preview images safely: `/link-preview/image/*` (fetch page, extract `og:image`, then proxy) and `/link-preview/proxy-image/*` (proxy a direct image URL), both with **2MB image limits**, short timeouts, inferred content types, and clearer status/error handling.
> 
> Updates the frontend link preview widget to use the new proxy endpoints, prefer hostname/`siteName`/title fallbacks (including a new `getHostname` helper + tests and new Storybook scenarios), and drops legacy citation widget placeholder support in `TextPart`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45732334ea607b53e84a816a69ddee04c54a804f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/pro/link-preview.ts
+++ b/backend/src/pro/link-preview.ts
@@ -150,14 +150,6 @@ const validateImageUrl = (url: string): { valid: boolean; error?: string } => {
       return { valid: false, error: 'Internal URLs are not allowed' }
     }
 
-    // Block IPv4-mapped IPv6 addresses (::ffff:x.x.x.x)
-    if (hostname.startsWith('::ffff:')) {
-      const ipv4Part = hostname.slice(7)
-      if (ipv4Part === '127.0.0.1' || ipv4Part === '0.0.0.0') {
-        return { valid: false, error: 'Internal URLs are not allowed' }
-      }
-    }
-
     const ipv4Regex =
       /^(?:(?:10|127)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|172\.(?:1[6-9]|2[0-9]|3[01])\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|192\.168\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|169\.254\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))$/
     const ipv6LinkLocalRegex = /^fe[89ab][0-9a-f]/
@@ -165,6 +157,22 @@ const validateImageUrl = (url: string): { valid: boolean; error?: string } => {
 
     if (ipv4Regex.test(hostname) || ipv6LinkLocalRegex.test(hostname) || ipv6UniqueLocalRegex.test(hostname)) {
       return { valid: false, error: 'Internal URLs are not allowed' }
+    }
+
+    // Block IPv4-mapped IPv6 (::ffff:XXYY:ZZWW) — Bun normalizes ::ffff:127.0.0.1 to ::ffff:7f00:1
+    if (hostname.startsWith('::ffff:')) {
+      const mapped = hostname.slice(7)
+      const hexParts = mapped.split(':')
+      if (hexParts.length === 2) {
+        const high = parseInt(hexParts[0], 16)
+        const low = parseInt(hexParts[1], 16)
+        if (!Number.isNaN(high) && !Number.isNaN(low)) {
+          const ipv4 = `${(high >> 8) & 0xff}.${high & 0xff}.${(low >> 8) & 0xff}.${low & 0xff}`
+          if (ipv4Regex.test(ipv4)) {
+            return { valid: false, error: 'Internal URLs are not allowed' }
+          }
+        }
+      }
     }
 
     if (hostname.includes('metadata') || hostname.includes('169.254.169.254')) {
--- a/backend/src/pro/link-preview.ts
+++ b/backend/src/pro/link-preview.ts
@@ -30,38 +30,196 @@ const resolveUrl = (baseUrl: string, relativeUrl: string): string => {
 }
 
 /**
- * Extracts Open Graph metadata from HTML content
+ * Fetches and proxies an image with size limits and timeout.
+ * Returns a Response with the image data or an error response.
+ */
+const fetchAndProxyImage = async (
+  imageUrl: string,
+  fetchFn: typeof fetch,
+  ctx: { set: { status?: number | string } },
+): Promise<Response> => {
+  try {
+    const controller = new AbortController()
+    const timeoutId = setTimeout(() => controller.abort(), 2000)
+
+    try {
+      const response = await fetchFn(imageUrl, {
+        method: 'GET',
+        headers: {
+          'User-Agent':
+            'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
+        },
+        signal: controller.signal,
+      })
+
+      if (!response.ok) {
+        ctx.set.status = response.status
+        const errorMessage = response.statusText || `HTTP ${response.status}`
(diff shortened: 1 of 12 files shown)


src/widgets/link-preview/utils.ts
```diff
@@ -0,0 +1,21 @@
+/** Extracts a clean hostname from a URL (strips "www." prefix) */
+export const getHostname = (url: string): string => {
```
raivieiraadriano92, line 2: I'd say it would be good to have some tests for this new method

src/widgets/link-preview/widget.tsx
```diff
@@ -91,27 +96,34 @@ const FetchLinkPreview = ({
...
+        description: preview.description,
         image: preview.image,
+        siteName: 'siteName' in preview ? (preview.siteName as string | null) : null,
       }
     },
   })
 
+  const imageUrl =
+    data?.image && cloudUrl && cloudUrl.trim()
+      ? `${cloudUrl}/pro/link-preview/proxy-image/${encodeURIComponent(data.image)}`
```
raivieiraadriano92: I see that above we are repeating the same logic to get the imageUrl - maybe we could extract this in a small utility method?

src/integrations/thunderbolt-pro/tools.ts
```diff
@@ -82,10 +82,15 @@ export const createConfigs = (httpClient: HttpClient, sourceCollector?: SourceMe
       verb: 'fetching {url}',
       parameters: fetchContentSchema,
       execute: async (params: FetchContentParams) => {
-        const result = await fetchContent(params, httpClient)
+        // Fetch content and link preview in parallel — link preview gives us og:site_name
+        const [result, preview] = await Promise.all([
+          fetchContent(params, httpClient),
+          fetchLinkPreview({ url: params.url }, httpClient).catch(() => null),
```
raivieiraadriano92, line 93: I noticed that not necessarily the model will show a link preview widget after requesting `/v1/pro/link-preview` - so I'm wondering if `fetchLinkPreview` couldn't be executed inside of the link preview widget.

This way we can avoid making requests which will not be used. Right now it's just a few request but when scaling this could save lots of bandwidth.

Not sure this thought makes sense - let me know what you think.

raivieiraadriano92 (changes requested):
LGTM - I just added a few minor questions there.

src/integrations/thunderbolt-pro/tools.ts
ital0: well the main reason that this call lives in `tools.ts` is for Citation badges, not link previews. 

When the AI fetches a page via `fetch_content`, the source gets registered with `siteName` derived from the hostname (e.g., nytimes.com). The parallel `fetchLinkPreview` call enriches that with `og:site_name` (e.g., The New York Times) which shows up in the citation badge and citations always render (well, it should be...), even when there's no link preview widget.

The link preview widget itself has its own separate fetch path via `useMessageCache`, so there's no duplication there.

That said, you raise a valid point about bandwidth at scale. If we want to optimize later, we could make this opt-in (only fetch when the source doesn't already have a good siteName, maybe? [but it's hard to validate that]) or move it to a lazy background enrichment. For now the parallel `Promise.all` keeps it zero-latency relative to the Exa call.

@raivieiraadriano92 What do you think?

backend/src/pro/link-preview.ts
```diff
@@ -30,38 +30,200 @@ const resolveUrl = (baseUrl: string, relativeUrl: string): string => {
...
+        signal: controller.signal,
+      })
+
+      if (!response.ok) {
+        ctx.set.status = response.status
+        const errorMessage = response.statusText || `HTTP ${response.status}`
+        return new Response(`Failed to fetch image: ${errorMessage}`, {
+          headers: { 'Content-Type': 'text/plain' },
+        })
+      }
```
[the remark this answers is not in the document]
ital0: Elysia does apply ctx.set.status to raw Response objects.

src/integrations/thunderbolt-pro/tools.ts
raivieiraadriano92: I see - since the citations should always be rendered I think it makes sense to keep it as it is

cjroth (changes requested):
PR Review: Source Citations Feature
Nice work — the architecture is solid, especially the streaming-safe CitationPopoverProvider pattern and the source registry approach. A few things to address before merge:
1. SSRF on the metadata endpoint (security, blocking)
The /image/* and /proxy-image/* endpoints run validateImageUrl() for SSRF protection, but the main GET /link-preview/* metadata endpoint only checks protocol. It will happily fetch http://169.254.169.254/latest/meta-data/ or any private IP.
Extract the IP-blocking logic from validateImageUrl into a shared validateSafeUrl function and apply it to the metadata endpoint too.
2. Unbounded HTML download (security, blocking)
The metadata endpoint calls await response.text() with no size limit. A malicious page could return a very large response. The image endpoints enforce 2MB — add a similar cap (or at minimum a Content-Length check) to the metadata fetch.
3. Image proxy waterfall (architecture)
The current FetchLinkPreview flow creates a request waterfall:
Fetch metadata JSON → wait for response → get image URL
Render  → second round trip
The /image/* endpoint was built to solve this (page fetch → extract og:image → proxy bytes in one request), but the frontend doesn't use it. Wire up FetchLinkPreview to use /link-preview/image/{pageUrl} as the  so the image loads in parallel with the metadata fetch.
Once that's done, consider whether /proxy-image/* still earns its keep — it's only useful for the InstantLinkPreview path where you already have a direct image URL from the source registry. If you want to keep one endpoint, /image/* can cover both cases (at the cost of a redundant page fetch for the instant path).
4. Duplicate type declaration (bug)
backend/src/pro/types.ts declares LocationSearchRequest twice. This should be a compile error — remove the duplicate.
5. React.ComponentType<any> (code quality)
src/widgets/index.ts line 85 uses React.ComponentType<any> which violates the project's "never use any" rule. Consider a union type of the actual widget prop types, or a generic widget props base type.
6. JSON.stringify in useMemo deps (perf)
In assistant-message.tsx, JSON.stringify(message.metadata?.sources) runs on every render. In research mode with up to 200 sources, that's serializing a large array every cycle. Consider a lighter comparison — e.g., just checking array length + reference equality, or computing a hash during collection.
7. Source registry cap is silent (robustness)
sourceRegistryCap = 200 in tools.ts silently drops sources beyond 200 — the index keeps incrementing but no metadata is stored, so [N] citations for those sources render as raw [N] text. Add a console.warn when the cap is hit so it's debuggable.
8. Old <widget:citation> path — still needed? (simplification)
The text-part.tsx rendering maintains two full code paths: the new [N] format (Path A) and the old <widget:citation> XML format (Path B). If no current model produces the old format, consider removing Path B — it adds significant complexity to buildTextWithCitationPlaceholders and the conditional branching in the useMemo.
Nits
The extractMetadata regex won't handle multi-line OG tag attributes (e.g., content on a separate line from property). Fine for 99% of pages, just worth knowing.
fetch_content's parallel fetchLinkPreview call uses .catch(() => null) which swallows all errors silently. A console.debug there would help debugging.
badgeColors in source-card.tsx uses charCodeAt(0) % 5 which clusters common letters to the same color. A hash of the full string would distribute better.

ital0:
@cjroth Thanks for the thorough review! Points #1 (SSRF), #2 (unbounded HTML), and #7 (silent cap) are spot on. I'll address those. A few notes on the rest:
                                                                                                                                                                                                                                                                                                                                                                          
  #3 — Image proxy waterfall: Agree about wiring up /image/* for the FetchLinkPreview path to kill the waterfall. Re: removing /proxy-image/*. It's still needed for the InstantLinkPreview path where we already have the direct image URL from the source registry. Using /image/* there would mean re-fetching the entire page just to extract the same URL we already have.                                                 

  #4 — Duplicate LocationSearchRequest: Checked types.ts. It only appears once (line 107). Might have been a duplicate in an earlier push that got cleaned up already.

  #5, #6, #8, and the nits about badgeColors/charCodeAt: these are all valid observations, but they reference files outside this PR's diff (widgets/index.ts, assistant-message.tsx, text-part.tsx, source-card.tsx). None of those files are touched here.

Specifically on #8 (<widget:citation>): the old XML path is intentional backward compat. Existing stored messages still use that format from before the [N] migration, so removing it would break rendering of historical conversations. Low-risk to keep for now.

PR #379, opened 2026-02-23
THU-295: Confirm external link modal before navigation

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Touches all outbound link handling and introduces new dialog/state + native webview hide/show behavior; regressions could block link navigation or affect desktop preview rendering.
> 
> **Overview**
> Adds a shared **external link confirmation modal** (`ExternalLinkDialog`) and a global `ExternalLinkDialogProvider`/`useExternalLinkDialog` hook that intercepts “safe” outbound links and only opens them after explicit confirmation (with error/loading states and an optional *Open in Sidebar* action on desktop).
> 
> Updates markdown link rendering (`SafeLink`), `SourceCard`, and link preview widgets to route clicks through the provider, and adjusts tests/stories to wrap relevant content with the provider.
> 
> Improves desktop preview layering by adding `previewHidden` to `ContentViewProvider` and wiring it through `SidebarWebview`/`useSidebarWebview` to call native webview `hide()`/`show()` while the modal is open; also tightens URL safety checks in `intercept-links` and `SidebarWebview` open-external actions. Removes the unused `tauri-plugin-shell` dependency and adds Tauri capabilities for webview hide/show.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d64f8aa3374cd611dfd58ef093891118c4bd4fd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/components/chat/markdown-utils.tsx
```diff
@@ -132,11 +133,35 @@ const processChildren = (children: ReactNode, citations?: CitationMap): ReactNod
...
+    setPendingUrl(safeHref)
+    setDialogOpen(true)
+  }
+
+  const handleConfirm = () => {
+    if (pendingUrl) {
+      window.open(pendingUrl, '_blank', 'noopener,noreferrer')
+    }
+    setDialogOpen(false)
+    setPendingUrl('')
```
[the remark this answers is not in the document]
darkbanjo: ^ this happens so fast I can't even see it visually. will leave this open unless you guys can catch it visually

src/components/chat/markdown-utils.tsx
```diff
@@ -126,17 +138,60 @@ const processChildren = (children: ReactNode, citations?: CitationMap): ReactNod
...
  */
 const SafeLink = memo(({ href, children, ...props }: React.ComponentProps<'a'>) => {
+  const context = useContext(ExternalLinkDialogContext)
+  const { dialogOpen, pendingUrl, openDialog, handleConfirm, setDialogOpen } = useExternalLinkDialog()
+
   const safeHref = href && isSafeUrl(href) ? href : undefined
+
+  if (context) {
+    const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
+      e.preventDefault()
```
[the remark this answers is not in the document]
darkbanjo: Isn't this still our responsibility? Or do we not care if they cmd+click

src/components/chat/markdown-utils.tsx
```diff
@@ -126,17 +138,60 @@ const processChildren = (children: ReactNode, citations?: CitationMap): ReactNod
...
+      if (!safeHref) return
+      context.openExternalLink(safeHref)
+    }
+    return (
+      <a {...props} href="#" onClick={handleClick}>
```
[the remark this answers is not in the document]
darkbanjo: again - how do we want to address accessibility?

src/components/chat/external-link-dialog.tsx
```diff
@@ -0,0 +1,42 @@
+import {
```
cjroth, line 1: Can we add a storybook for this?

src/hooks/use-external-link-dialog.ts
```diff
@@ -0,0 +1,53 @@
...
+  }, [])
+
+  const handleConfirm = useCallback(async () => {
+    const urlToOpen = pendingUrlRef.current
+    setDialogOpen(false)
+    setPendingUrl('')
+    pendingUrlRef.current = ''
+    if (urlToOpen) {
+      if (isTauri()) {
+        // In Tauri, use shell.open to open URLs in system browser
```
cjroth: I can't think of any reason why we would switch to a shell approach here. Also this makes me shiver from a security perspective.

src/components/chat/markdown-utils.tsx
```diff
@@ -126,19 +138,90 @@ const processChildren = (children: ReactNode, citations?: CitationMap): ReactNod
   return processTextContent(afterCitations)
 }
 
+/**
+ * Provider that renders a single ExternalLinkDialog for all SafeLinks in the tree.
```
cjroth, line 155: @raivieiraadriano92 i'm curious to get your opinion on this, since you did a lot of work on the performance issues of chat ui.

Is it worth the added complexity to add a modal provider here to avoid rendering one modal per link?

Personally I'm not so sure that it's worth it, but performance is very important. I think it's unnecessary *rerenders* that are usually the issue more than just too many UI components.

cjroth (changes requested):

src/hooks/use-external-link-dialog.ts
```diff
@@ -0,0 +1,52 @@
...
+  const [pendingUrl, setPendingUrl] = useState<string>('')
+  const pendingUrlRef = useRef<string>('')
+
+  const openDialog = useCallback((url: string) => {
+    pendingUrlRef.current = url
+    setPendingUrl(url)
+    setDialogOpen(true)
+  }, [])
+
+  const handleConfirm = useCallback(async () => {
```
ital0, line 76: Small thing. `handleConfirm` is async but the call site (`AlertDialogAction onClick`) doesn't await it. If the dynamic import or `openUrl()` rejects, the error gets silently swallowed and the user sees the dialog disappear with nothing happening.

src/hooks/use-external-link-dialog.ts
```diff
@@ -0,0 +1,52 @@
...
+    const urlToOpen = pendingUrlRef.current
+    setDialogOpen(false)
+    setPendingUrl('')
+    pendingUrlRef.current = ''
+    if (urlToOpen) {
```
ital0: I know user needs to confirm if they want to open a link, but all the callers currently validate with `isSafeUrl()` before calling `openDialog()`, which is great. Would it be worth adding a defense-in-depth check here too? Since the hook is exported, a future caller could skip validation.

src/components/chat/markdown-utils.tsx
ital0: to add some context to @cjroth's question -- the provider does help avoid N dialog DOM nodes (portals + overlays) per message, which can matter during streaming with lots of links. The `useMemo` on `contextValue` also prevents link re-renders when dialog state changes.

one simplification: since `MemoizedMarkdown` always wraps in `ExternalLinkDialogProvider`, does `SafeLinkStandalone` actually get used in practice? If not, the 3-component split could be simplified to just `SafeLink` always reading from context, and you'd throw if context is missing. That would cut the complexity in half while keeping the performance benefit.

src-tauri/capabilities/default.json
```diff
@@ -7,6 +7,7 @@
     "core:default",
     "opener:default",
     "os:default",
+    "shell:default",
```
ital0: I think you're going to remove that, right?

ital0 (changes requested):
Your PR is looking good! Other than the tiny issues I mentioned, LGTM!

src-tauri/capabilities/default.json
darkbanjo: yep good catch, thank you!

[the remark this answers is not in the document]
darkbanjo: removed, this was addressed in another comment as well. resolving.

src/components/chat/markdown-utils.tsx
darkbanjo: yeah good call out @ital0 -- i simplified to a single safelink

src/components/chat/external-link-dialog.tsx
darkbanjo: added!

src/components/chat/source-card.test.tsx
```diff
@@ -75,40 +79,80 @@ describe('SourceCard', () => {
...
+      const mockWindowOpen = mock(() => null)
+      window.open = mockWindowOpen as typeof window.open
+
+      renderWithProvider(<SourceCard source={mockSource} />)
+
+      const link = screen.getByRole('listitem')
+      fireEvent.click(link)
+
+      const cancelButton = screen.getByRole('button', { name: 'Cancel' })
+      fireEvent.click(cancelButton)
```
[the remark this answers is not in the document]
darkbanjo: removed cancel button anyway, replaced with X

src-tauri/src/lib.rs
```diff
@@ -32,7 +32,6 @@ pub fn create_app() -> tauri::Builder<tauri::Wry> {
         .plugin(tauri_plugin_process::init())
         .plugin(tauri_plugin_fs::init())
         .plugin(tauri_plugin_opener::init())
-        .plugin(tauri_plugin_shell::init())
```
cjroth, line 35: Wow, I didn't even realize we had this installed. Glad we're removing it.

cjroth (changes requested):
I think I agree with Claude here - thoughts?

Image

src-tauri/src/lib.rs
darkbanjo: :D

darkbanjo:
> I think I agree with Claude here - thoughts?
> Image

i agree, this is a good catch. smart to move it up rather than mount/remount. i'll make this change!

PR #381, opened 2026-02-24
THU-278: Back button from settings now returns to correct chat, or falls back to new chat

darkbanjo:
> [!NOTE]
> **Low Risk**
> Small client-side navigation change scoped to the sidebar’s back behavior; low risk aside from potential edge cases around pending query state and stale stored paths.
> 
> **Overview**
> Updates the settings sidebar “back” behavior to navigate to the *last visited* `/chats/*` path (tracked via a new `lastChatPathRef`) instead of always selecting the current/first thread.
> 
> Adds `isPending` awareness to `chatThreads` loading so the back action waits for data when no last-chat fallback exists, and otherwise falls back to the first loaded thread or `/chats/new` creation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c95ecd47bfa1fcf6b31093a0e74730894f822541. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/layout/sidebar/index.tsx
+++ b/src/layout/sidebar/index.tsx
@@ -161,9 +161,8 @@ export default function Sidebar() {
   }
 
   const goToMainMenu = async () => {
-    // Only block navigation if query is pending AND we have no safe fallback
-    // (no current chat ID from URL, no last chat path)
-    if (isPending && !currentChatThreadId && !lastChatPathRef.current) {
+    // Only block navigation if query is pending and we have no last chat path to return to
+    if (isPending && !lastChatPathRef.current) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


src/layout/sidebar/index.tsx
```diff
@@ -123,10 +131,47 @@ export default function Sidebar() {
...
-    if (chatThreadId) {
-      navigate(`/chats/${chatThreadId}`)
+    // Only block navigation if query is pending and we have no last chat path to return to
+    if (isPending && !lastChatPathRef.current) {
+      return
+    }
+
+    const allThreads = data ?? []
+    // Use lastChatPathRef if available (skip validation if query is pending since we can't validate against empty array)
+    if (lastChatPathRef.current && (isPending || isChatPathValid(lastChatPathRef.current, allThreads))) {
```
cjroth: This assumes that all threads will be returned in the useQuery, but we will need to add continuous scroll / pagination here soon to handle large numbers of chats.

Tbh this feels overly complex to me - what is the scenario that we're trying to prevent by adding this check? If we truly needed to check that a chat isn't deleted, could we check the database lazily as needed? A single SQL lookup should be fast enough in that scenario.

As part of just generally feeling like a heavy solution, I worry that adding a lot of logic here increases our bundle size. If there's a more concise solution, that would be idea.

cjroth (changes requested):

src/layout/sidebar/index.tsx
darkbanjo: yeah good call -- i actually realized i kind of solved for this situation already when i worked on the /not-found page and how it handles chats that don't exist. updated to use-hydrate-chat-store

cjroth (approved):

PR #382, opened 2026-02-24
THU-307: Reduce input box footprint on mobile

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Mostly UI/layout changes, but it alters `useIsMobile` initialization (reads `window` immediately) and removes the prior `isReady` guard, which could introduce SSR/hydration or runtime issues if the hook is ever used outside a browser-only context.
> 
> **Overview**
> Updates `PromptInput` to require an `isMobile` prop and render a tighter *mobile-first* layout (smaller textarea height, controls moved below), while preserving the existing desktop layout.
> 
> Switches `ChatPromptInput` to use the shared `useIsMobile` hook (instead of sidebar state), tweaks padding/gaps, makes `ModeSelector` icon-only on mobile, and hides the context-usage indicator on mobile; `AutomationFormModal` explicitly opts into desktop layout. Also simplifies `useIsMobile` (removes `isReady`) and adjusts `useAutosizeTextArea` to recompute sizing when `minHeight`/`maxHeight` change; tests are updated for the new mobile behavior and hook injection.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9e46e3e32d564a2db087d7a7ffb00ecf28391971. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/components/ui/prompt-input.tsx
```diff
@@ -72,7 +75,59 @@ export const PromptInput = forwardRef<HTMLFormElement, PromptInputProps>(
...
+            chatThread={chatThread}
+            models={models}
+            selectedModelId={selectedModelId}
+            onModelChange={onModelChange}
+          />
+        )}
+
+        {submitButton}
+      </div>
+    )
```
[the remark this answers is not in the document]
darkbanjo: this is intentional, visually it is more appealing and only minimally changes the vertical height

src/components/chat/chat-prompt-input.tsx
```diff
@@ -1,5 +1,6 @@
 import { useCurrentChatSession, useChatStore } from '@/chats/chat-store'
 import { useContextTracking as useContextTracking_default } from '@/hooks/use-context-tracking'
+import { useIsMobile as useIsMobile_default } from '@/hooks/use-mobile'
```
cjroth, line 3: Not sure if I follow why we need two instances of useIsMobile here? seems funky

src/components/ui/prompt-input.tsx
```diff
@@ -47,6 +49,7 @@ export const PromptInput = forwardRef<HTMLFormElement, PromptInputProps>(
       isStreaming = false,
       onStop,
       footerStartElements,
+      isMobile,
```
cjroth, line 52: I'm half leaning toward splitting into two child components, one for desktop and one for mobile. But I think this is fine.

Also, I would find it valuable to see a storybook of this so we can play around with different prop configurations and screen sizes in a controlled environment.

darkbanjo: i'll add a storybook!

src/components/chat/chat-prompt-input.tsx
```diff
@@ -113,10 +117,17 @@ export const ChatPromptInput = forwardRef<ChatPromptInputRef, ChatPromptInputPro
     const footerStartElements = (
```
ital0, line 117: small thing: the `isReady` guard on line 127 is placed after the `footerStartElements` JSX construction (line 117). That JSX gets built and immediately discarded when `isReady` is false. Could we move the guard above it? All hooks are already called by that point, so it's safe to early-return there. Fits CLAUDE.md's "prefer early return" principle!

src/components/chat/chat-prompt-input.test.tsx
```diff
@@ -32,21 +32,12 @@ const createMockUseContextTracking =
     estimateTokensForInput: (_input: string) => 0,
   })
```
ital0, line 33: The `isReady` guard is the core defensive mechanism of this PR, but there's no test for `isReady=false`. The mock factory already supports it! Would be great to add a quick test.

src/components/chat/chat-prompt-input.tsx
```diff
@@ -131,8 +142,9 @@ export const ChatPromptInput = forwardRef<ChatPromptInputRef, ChatPromptInputPro
           onStop={stop}
           autoFocus={!isMobile}
           submitOnEnter={!isStreaming && !isPlatformMobile()}
-          className="flex flex-col gap-2 bg-background dark:bg-input/30 border dark:border-input p-3 rounded-2xl w-full"
+          className={`flex flex-col gap-2 bg-background dark:bg-input/30 border dark:border-input rounded-2xl w-full ${isMobile ? 'px-3 py-1' : 'p-3'}`}
```
ital0: tiny style nit: the template literal on line 145 could use `cn()` instead, which is what the rest of the codebase uses for conditional classes. More consistent and handles Tailwind class merging better:

 ```tsx
 className={cn(
   'flex flex-col gap-2 bg-background dark:bg-input/30 border dark:border-input rounded-2xl w-full',
   isMobile ? 'px-3 py-1' : 'p-3'
 )}
 ```

src/components/ui/prompt-input.tsx
```diff
@@ -72,7 +75,59 @@ export const PromptInput = forwardRef<HTMLFormElement, PromptInputProps>(
```
ital0, line 75: `PromptInput` now has two structurally different layouts but no test file. Totally fine as a follow-up (especially with the Storybook coming!), but wanted to flag it.

ital0 (changes requested):
just noticed some small points that! Take a look before merging.

src/components/chat/chat-prompt-input.tsx
darkbanjo: i saw that pattern in a few different places (use-chat-scroll-handler.ts for example) where we import as useX_default? might be unnecessary here

src/components/chat/chat-prompt-input.tsx
darkbanjo: great call out, added

src/components/ui/prompt-input.tsx
darkbanjo: sounds good - i added a storybook to latest changes as per you and Chris' comments

src/components/chat/chat-prompt-input.tsx
```diff
@@ -131,8 +141,9 @@ export const ChatPromptInput = forwardRef<ChatPromptInputRef, ChatPromptInputPro
           onStop={stop}
           autoFocus={!isMobile}
           submitOnEnter={!isStreaming && !isPlatformMobile()}
-          className="flex flex-col gap-2 bg-background dark:bg-input/30 border dark:border-input p-3 rounded-2xl w-full"
+          className={`flex flex-col bg-background dark:bg-input/30 border dark:border-input rounded-2xl w-full ${isMobile ? 'gap-0 p-4' : 'gap-2 p-3'}`}
```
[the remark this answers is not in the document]
darkbanjo: resolved in latest commit, using cn()

PR #402, opened 2026-02-27
THU-312: implement db.transaction to optimize PowerSync operations

raivieiraadriano92:
> [!NOTE]
> **Medium Risk**
> Moderate risk: changes touch multiple DAL write paths (models/prompts/triggers/chat threads/messages/settings) and adjust function signatures to accept transactions, which could introduce subtle behavior differences or missed call-site updates, but the logic is largely refactoring for atomicity and PowerSync compatibility.
> 
> **Overview**
> **PowerSync write operations are now batched/atomic via `db.transaction(...)`** across key DAL flows to reduce sync overhead and avoid partial updates.
> 
> This refactors several DAL helpers to accept an optional `AnyDrizzleDatabase` transaction and threads it through multi-step operations (e.g. `createModel` + default profile creation, `deleteModel` + prompt/trigger/profile cleanup, `runAutomation` thread + seed message, `saveMessagesWithContextUpdate` message writes + thread context update, and chat thread/message soft-deletes).
> 
> Separately, `createChatThread` now takes a resolved `Model` object (instead of a `modelId`) to avoid redundant lookups and to support transactional callers; tests were updated accordingly. PowerSync connection is also tuned with `crudUploadThrottleMs: 5000`, and coding guidelines now prefer top-level imports over dynamic imports when possible.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit abdccd2aff801aea6854ebed005ce6f1846ea6f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/dal/models.ts
+++ b/src/dal/models.ts
@@ -145,21 +145,18 @@ export const resetModelToDefault = async (id: string, defaultModel: Model): Prom
  * Only updates records that haven't been deleted yet to preserve original deletion datetimes
  */
 export const deleteModel = async (id: string): Promise<void> => {
-  // Import locally to avoid circular dependency
-  const { deletePromptsForModel } = await import('./prompts')
-
-  // Soft-delete model profile
-  const { deleteModelProfileForModel } = await import('./model-profiles')
-  await deleteModelProfileForModel(id)
-
-  // Soft-delete prompts and their triggers first (replaces onDelete: 'cascade')
-  await deletePromptsForModel(id)
-
   const db = DatabaseSingleton.instance.db
-  await db
-    .update(modelsTable)
-    .set({ ...clearNullableColumns(modelsTable), deletedAt: nowIso() })
-    .where(and(eq(modelsTable.id, id), isNull(modelsTable.deletedAt)))
+  await db.transaction(async (tx) => {
+    const { deletePromptsForModel } = await import('./prompts')
+    const { deleteModelProfileForModel } = await import('./model-profiles')
+
+    await deleteModelProfileForModel(id, tx)
+    await deletePromptsForModel(id, tx)
+    await tx
+      .update(modelsTable)
+      .set({ ...clearNullableColumns(modelsTable), deletedAt: nowIso() })
+      .where(and(eq(modelsTable.id, id), isNull(modelsTable.deletedAt)))
+  })
 }
 
 /**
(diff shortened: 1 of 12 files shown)


ital0 (approved):
LGTM

src/dal/models.ts
```diff
@@ -145,21 +145,18 @@ export const resetModelToDefault = async (id: string, defaultModel: Model): Prom
...
-  // Soft-delete prompts and their triggers first (replaces onDelete: 'cascade')
-  await deletePromptsForModel(id)
-
   const db = DatabaseSingleton.instance.db
-  await db
-    .update(modelsTable)
-    .set({ ...clearNullableColumns(modelsTable), deletedAt: nowIso() })
-    .where(and(eq(modelsTable.id, id), isNull(modelsTable.deletedAt)))
+  await db.transaction(async (tx) => {
+    const { deletePromptsForModel } = await import('./prompts')
```
cjroth: Can these be normal imports?

src/dal/models.ts
```diff
@@ -169,9 +166,9 @@ export const createModel = async (
...
 ): Promise<void> => {
   const db = DatabaseSingleton.instance.db
-  await db.insert(modelsTable).values(data)
-
-  // Auto-create default profile if one exists in seed data
-  const { createDefaultModelProfile } = await import('./model-profiles')
-  await createDefaultModelProfile(data.id)
+  await db.transaction(async (tx) => {
+    await tx.insert(modelsTable).values(data)
+    const { createDefaultModelProfile } = await import('./model-profiles')
```
cjroth: And this one

cjroth (commented):
Looks good - but haven't tested. For the inline import()s - I know sometimes we use them to avoid circular dependencies or in edge cases where we really only want to load a dependency lazily, but I find that Claude tends to use them too often. Maybe we could update AGENTS.md to ask it not to do this?

src/dal/models.ts
raivieiraadriano92: Done - also updated the AGENTS.md to avoid this

src/dal/models.ts
raivieiraadriano92: done

cjroth (approved):

PR #404, opened 2026-02-27
THU-314: Verification failed error message displays before navigating away

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Touches OTP verification/sign-in flows and React Query settings caching, which can affect authentication UX and session initialization if regressions occur.
> 
> **Overview**
> Improves sign-in verification stability by preventing `MagicLinkVerify` from re-submitting an already-consumed OTP when the session `refetch` function changes and re-triggers the effect.
> 
> Adds a lightweight, one-shot `WelcomeDialog` (backed by a small Zustand store + tests) that is triggered after successful waitlist OTP verification for *returning* users, and mounts it at the app root so it appears after redirect.
> 
> Tweaks `useSettings` to use React Query `keepPreviousData` as `placeholderData`, reducing transient "missing settings" states during refetches (e.g., around auth client initialization).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30992ee4f6f2dc2bff8adf71e168139afee3cf48. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/welcome-dialog.tsx
+++ b/src/components/welcome-dialog.tsx
@@ -1,0 +1,43 @@
+'use client'
+
+import { useState } from 'react'
+
+import { SignInSuccessStep } from '@/components/sign-in/sign-in-success-step'
+import { Dialog, DialogContent, DialogDescription, DialogTitle } from '@/components/ui/dialog'
+import { useSettings } from '@/hooks/use-settings'
+
+/** Module-level flag — survives route changes since it's just a JS variable. */
+let pendingWelcome = false
+
+/** Call after successful OTP verification to show the welcome dialog on the next mount. */
+export const triggerWelcome = () => {
+  pendingWelcome = true
+}
+
+/**
+ * Shows a welcome dialog after successful sign-in from the waitlist page.
+ */
+export const WelcomeDialog = () => {
+  const { preferredName } = useSettings({ preferred_name: '' })
+  const [isOpen, setIsOpen] = useState(() => {
+    if (pendingWelcome) {
+      pendingWelcome = false
+      return true
+    }
+    return false
+  })
+
+  return (
+    <Dialog open={isOpen} onOpenChange={setIsOpen}>
+      <DialogContent className="sm:max-w-md" showCloseButton={false}>
+        <DialogTitle className="sr-only">Welcome</DialogTitle>
+        <DialogDescription className="sr-only">Sign-in successful</DialogDescription>
+        <SignInSuccessStep
+          displayName={preferredName.value as string}
+          onContinue={() => setIsOpen(false)}
+          variant="modal"
+        />
+      </DialogContent>
+    </Dialog>
+  )
+}
--- a/src/components/welcome-dialog.tsx
+++ b/src/components/welcome-dialog.tsx
@@ -1,31 +1,32 @@
 'use client'
 
 import { useState } from 'react'
+import { create } from 'zustand'
 
 import { SignInSuccessStep } from '@/components/sign-in/sign-in-success-step'
 import { Dialog, DialogContent, DialogDescription, DialogTitle } from '@/components/ui/dialog'
 import { useSettings } from '@/hooks/use-settings'
 
-/** Module-level flag — survives route changes since it's just a JS variable. */
-let pendingWelcome = false
-
-/** Call after successful OTP verification to show the welcome dialog on the next mount. */
-export const triggerWelcome = () => {
-  pendingWelcome = true
-}
+export const useWelcomeStore = create<{
+  pending: boolean
+  trigger: () => void
+  consume: () => boolean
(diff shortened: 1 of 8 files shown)


src/contexts/auth-context.tsx
```diff
@@ -93,13 +98,19 @@ export const AuthProvider = ({ children, authClient: overrideClient }: AuthProvi
...
+  // Wait for auth client to be ready before rendering children.
+  // Once initialized, keep children mounted during transient reloads
+  // to avoid unmounting in-flight auth flows (e.g. magic link verification).
+  const resolvedValue = value ?? lastValueRef.current
+  if (!resolvedValue) {
     return null
   }
 
-  return <AuthContext.Provider value={value}>{children}</AuthContext.Provider>
+  return <AuthContext.Provider value={resolvedValue}>{children}</AuthContext.Provider>
```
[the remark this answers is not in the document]
darkbanjo: i added a comment to the code, this is an edge case but nonetheless worth documenting

cjroth (changes requested):
The root cause is being treated, not fixed

  The actual bug chain is:

  1. OTP succeeds → refetchSession() updates auth state
  2. PowerSync syncs → invalidates the settings table
  3. usePowerSyncInvalidation calls queryClient.invalidateQueries({ queryKey: ['settings'] }) — this invalidates all settings queries
  4. useSettings({ cloud_url: String }) in AuthProvider transitions to isLoading: true during refetch
  5. useMemo returns null → AuthProvider returns null → entire tree unmounts
  6. Tree remounts → MagicLinkVerify effect re-fires with consumed OTP → error flash

  The real fix is one line — tell React Query to keep serving the previous data while refetching. In useSettings at line 207:

  useQuery({
    queryKey: ['settings', ...keys],
    queryFn: ...,
    placeholderData: keepPreviousData,  // ← this
  })

  This would prevent isLoading from flipping to true during a refetch, so AuthProvider would never unmount the tree, and the entire cascade never
  happens. No need for Parts 1, 2, or 3.

  Specific concerns with the current approach

  Part 1 (AuthProvider ref caching) — Reasonable defensive measure, but it's a workaround for the settings query lacking placeholderData. It also
  introduces subtle behavior: the provider serves a stale value during transient reloads, which is fine for cloudUrl but could mask real issues if
  this pattern gets copied to other providers.

  Part 2 (OTP dedup ref) — This one is independently valuable. Even with the root cause fixed, it's good hygiene to guard against double-submitting
  a consumed OTP. The refetchSession identity change could still re-trigger the effect. I'd keep this.

  Part 3 (WelcomeDialog with module-level flag) — This is the weakest part:

  - Module-level mutable state is a code smell. It's invisible coupling outside React's data flow, not testable in isolation, and fragile to mount
  ordering. If WelcomeDialog were already mounted when triggerWelcome() fires, the flag is never consumed (the useState initializer only runs once).
  - Is it even needed? The MagicLinkVerify component already has a success state with the same "Welcome" UI. If the unmount is fixed, does this
  dialog serve a purpose? It's only triggered from the waitlist flow for returning users — is that a different UX path that genuinely needs a
  separate welcome modal, or was it added to work around the unmount destroying the original success screen?

  What I'd recommend

  1. Fix the root cause: add placeholderData: keepPreviousData to the useQuery in useSettings. This is one import + one line and prevents the entire
   cascade.
  2. Keep Part 2 (the OTP dedup ref) — it's good defensive code independent of the root cause.
  3. Question Part 3 — clarify with the PR author whether WelcomeDialog is a feature requirement or a workaround. If it's a workaround, drop it. If
  it's a genuine feature, the module-level flag should be replaced with something in React's data flow (a URL search param like ?welcome=1, or a
  zustand atom).
  4. Drop Part 1 if the placeholderData fix is applied — it becomes unnecessary.

  Testing

  The OTP dedup guard (Part 2) would benefit from a test: verify the effect doesn't re-call signIn.emailOtp when refetchSession changes identity but
   email/otp stay the same. The useSettings fix wouldn't need a new test since it's a config change, but the existing settings tests should still
  pass.

  Risk of bugs

  The current PR is unlikely to break anything — the changes are additive and defensive. The main risk is the module-level flag in WelcomeDialog
  silently failing (not showing the dialog) if mount timing changes in a future refactor, which would be a silent regression rather than a crash.

src/components/welcome-dialog.tsx
```diff
@@ -0,0 +1,46 @@
+'use client'
```
cjroth: Use client doesn't make sense here

cjroth (approved):
:shipit:

src/components/welcome-dialog.tsx
darkbanjo: removed, good catch

PR #441, opened 2026-03-06
⚡ THU-331: Persist prompt input draft per chat thread

cjroth:
## Summary
- Adds `useDraftInput` hook that persists chat prompt text to localStorage (debounced, keyed per chat thread)
- On submit, the draft is cleared immediately
- Restores draft when navigating back to a chat thread

## Linear
[THU-331](https://linear.app/mozilla-thunderbolt/issue/THU-331/persist-prompt-input-to-disk-so-that-user-doesnt-lose-their-place-if)

## Test Plan
- [x] Type in prompt, navigate away, come back — text is restored
- [x] Submit a message — draft is cleared
- [x] Switch between chat threads — each has its own draft
- [x] 7 unit tests for `useDraftInput` hook
- [x] All 9 existing `ChatPromptInput` tests still pass

## Changes
- `src/hooks/use-draft-input.ts` — New hook with debounced localStorage persistence
- `src/hooks/use-draft-input.test.ts` — 7 tests covering init, save, debounce, clear, thread switching
- `src/components/chat/chat-prompt-input.tsx` — Replace `useState('')` with `useDraftInput(chatThreadId)`

---

> [!NOTE]
> **Low Risk**
> Low risk UX change that adds debounced localStorage persistence for chat prompt drafts; main risk is subtle draft key/cleanup behavior causing stale or missing drafts across thread switches.
> 
> **Overview**
> Chat prompt drafts are now persisted to `localStorage` per chat thread via a new debounced `useLocalStorage` hook and a `useDraftInput` wrapper, so typed text is restored when navigating away/back or switching threads.
> 
> `ChatPromptInput` switches from `useState` to `useDraftInput` (including a stable `'new'` key for unsaved chats) and clears the persisted draft immediately on submit. Adds unit tests for both new hooks. Updates `.thunderbot/thunderfix.md` to include replying to questions/suggestions before resolving threads and minimizing comments, and fetches comment IDs needed for replies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2861478cddddd5518e5e6965177d7fc74c8fddd3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/hooks/use-draft-input.ts
+++ b/src/hooks/use-draft-input.ts
@@ -28,6 +28,7 @@ export const useDraftInput = (chatThreadId: string) => {
   const [input, setInputState] = useState(() => readDraft(chatThreadId))
   const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const chatThreadIdRef = useRef(chatThreadId)
+  const pendingValueRef = useRef<string | null>(null)
 
   // When the chat thread changes, load that thread's draft
   useEffect(() => {
@@ -35,15 +36,21 @@ export const useDraftInput = (chatThreadId: string) => {
     setInputState(readDraft(chatThreadId))
 
     return () => {
+      // Flush any pending debounced write before switching threads
       if (timerRef.current) {
         clearTimeout(timerRef.current)
         timerRef.current = null
+      }
+      if (pendingValueRef.current !== null) {
+        writeDraft(chatThreadIdRef.current, pendingValueRef.current)
+        pendingValueRef.current = null
       }
     }
   }, [chatThreadId])
 
   const setInput = useCallback((value: string) => {
     setInputState(value)
+    pendingValueRef.current = value
 
     if (timerRef.current) {
       clearTimeout(timerRef.current)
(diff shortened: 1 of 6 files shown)


cjroth:
⚡ Addressed in the latest push.

cjroth:
Can we also persist the new chat screen one too?

src/hooks/use-draft-input.ts
```diff
@@ -0,0 +1,77 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+const draftKeyPrefix = 'draft:'
+const debounceMs = 300
+
+/** Get the localStorage key for a chat thread's draft. */
+const getDraftKey = (chatThreadId: string) => `${draftKeyPrefix}${chatThreadId}`
```
cjroth: Do we have a useLocalStorage hook? If not, should we? Also would using useQuery in such a hook make it cleaner? (This is an idea, it's ok if the answer is no, just curious.)

cjroth:
⚡ Addressed in the latest push.

src/hooks/use-draft-input.ts
cjroth: ⚡ Good question. We don't have a `useLocalStorage` hook today — localStorage is used directly in `auth-token.ts`, `theme-provider.tsx`, and `database.ts`.

A generic `useLocalStorage<T>(key, defaultValue)` hook could consolidate those patterns and would be a natural home for this draft logic too. Wrapping it with `useQuery` would give you cache consistency across components reading the same key (e.g., if two components both read the draft, they'd stay in sync via the query cache rather than reading stale localStorage values).

For this PR I kept it focused on the specific use case (draft input needs debounced writes + per-thread keying + clear-on-submit), but a `useLocalStorage` primitive underneath could be a nice follow-up to extract — especially if more local-only persistence needs come up.

PR #444, opened 2026-03-08
refactor: migrate to @powersync/tanstack-react-query

raivieiraadriano92:
## Summary

Migrates from manual PowerSync query management to the official `@powersync/tanstack-react-query` integration. This provides better query invalidation, caching, and state management through TanStack Query's established patterns.

## Changes

- Migrated all data access layers to use TanStack Query hooks
- Removed `usePowerSyncInvalidation` hook and manual invalidation logic
- Created `PowerSyncMock` test utility for consistent test setup
- Updated `toCompilableQuery` usage across DAL functions

## Benefits

- Automatic query invalidation on database changes
- Better caching and deduplication via TanStack Query
- Consistent patterns for reactive queries across the codebase
- Reduced boilerplate for PowerSync query management

---

> [!NOTE]
> **Medium Risk**
> Broad refactor of data-fetching and cache invalidation across the app to rely on PowerSync live queries; regressions could surface as stale UI, missing updates, or subtle query-shape/type mismatches.
> 
> **Overview**
> Migrates UI data fetching from manual React Query `queryFn` + explicit `invalidateQueries` to PowerSync’s reactive `useQuery` (`@powersync/tanstack-react-query`) using `toCompilableQuery`, so views update automatically from local writes and sync.
> 
> Refactors multiple DAL read functions (e.g. prompts, triggers, devices, settings, models, chat threads, MCP servers, tasks) to return Drizzle query builders (and adds `getSelectedModelQuery`/`getModelQuery` plus `mapModel(s)` helpers), then updates call sites to consume the new query results shape (arrays/rows) and drops the now-unneeded `usePowerSyncInvalidation` hook.
> 
> Adds test infrastructure for PowerSync (`PowerSyncMockProvider`, `renderWithReactivity` + triggerable table changes) and introduces/updates reactivity tests for key pages/hooks; bumps `@powersync/react` to `^1.9.0`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5c5e60cf9e81e548ea51c672fd7d609b56d7102. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/widgets/citation/widget.test.tsx
```diff
@@ -17,6 +18,13 @@ const renderWithProviders = (ui: React.ReactElement) => {
 }
 
 describe('CitationWidgetComponent', () => {
+  beforeAll(async () => {
```
cjroth, line 21: Hmm - are we sure we want to make these components that don't need database access dependent on the connected powersync version? I feel like this makes them more fragile / connected and less modular.

vite.config.ts
```diff
@@ -71,6 +71,11 @@ export default defineConfig({
     alias: {
       '@': path.resolve(__dirname, './src'),
       '@shared': path.resolve(__dirname, './shared'),
+      // Force single @powersync/react instance so PowerSyncContext is shared (usePowerSyncQuery needs it)
+      '@powersync/react': path.resolve(__dirname, 'node_modules/@powersync/react'),
```
cjroth: Does this line actually accomplish anything - isn't this just a no-op?

src/automations/use-trigger-scheduler.ts
```diff
@@ -17,7 +18,7 @@ export const useTriggerScheduler = () => {
       timers.current.forEach(clearTimeout)
       timers.current = []
 
-      const triggers = await getAllEnabledTriggers()
+      const triggers = (await getAllEnabledTriggers()) as Trigger[]
```
cjroth, line 21: This seems funky - shouldnt getAllEnabledTriggers return Trigger[] type?

src/dal/settings.test.ts
```diff
@@ -185,10 +185,8 @@ describe('Settings DAL', () => {
         { recomputeHash: true },
       )
 
-      const records = await getSettingsRecords({
-        hash_key_one: String,
-        hash_key_two: String,
-      })
+      const recordsArray = await getSettingsRecords(['hash_key_one', 'hash_key_two'])
```
cjroth, line 188: Hmmm this is hard to read compared to previously - is it possible to keep it the previous usage?

src/dal/settings.ts
```diff
@@ -25,38 +25,17 @@ type SettingSchema = Record<
...
-    (acc, setting) => {
-      if (setting) {
-        acc[setting.key] = setting
-      }
-      return acc
-    },
-    {} as Record<string, Setting>,
-  )
-}
+export const getSettingsRecords = (keys: string[]) =>
```
cjroth, line 34: (here - possible to keep usage the same?)

cjroth (changes requested):
Claude using /thunderimprove skill (without making the changes itself) came back with:


     Summary

     This branch migrates from manual React Query cache invalidation to PowerSync's
     @powersync/tanstack-react-query live queries. DAL functions are converted from async
     functions returning data to synchronous functions returning Drizzle query builders,
     which are then passed through toCompilableQuery() for PowerSync-powered reactive
     updates. The usePowerSyncInvalidation hook is removed entirely.

     The overall direction is correct and is a significant improvement -- live queries are
     the right architectural approach. However, there are several concrete issues.

     ---
     Issues Found

     1. Unsafe as unknown as cast in app.tsx (line 151-152)

     value={powerSyncInstance as unknown as ComponentProps<typeof
     PowerSyncContext.Provider>['value']}

     This double cast (as unknown as) silences a type mismatch between PowerSyncDatabase
     (from @powersync/web) and the context value expected by @powersync/react. This is a
     bandaid for a version mismatch between @powersync/common versions. The vite.config.ts
     dedupe addition partially addresses this, but the cast is a red flag that the dedupe
     may not be sufficient at the type level. If the runtime instances actually differ, live
      queries will silently fail.

     Fix: Verify that dedupe in vite resolves the runtime context correctly, and if so, the
     cast should become unnecessary. If the types genuinely don't match, the PowerSync
     version pinning needs alignment.

     2. Repeated data.map(mapModel) pattern -- missing reusable hook

     mapModel is called in 6 different places with the same useMemo(() =>
     data.map(mapModel), [data]) pattern:
     - src/settings/models/detail.tsx:75
     - src/settings/models/layout.tsx:19
     - src/settings/models/index.tsx:254
     - src/automations/automation-form-modal.tsx:69
     - src/chats/use-hydrate-chat-store.ts:99 and :148

     This mapping is needed because the DAL now returns raw ModelRow objects instead of
     Model objects. Every consumer has to know about mapModel. This is a DRY violation and a
      correctness risk -- if someone forgets mapModel, they'll get a row missing api_key and
      is_system fields.

     Fix: Create a small hook like useModelsQuery(queryBuilder) that internally calls
     useQuery + toCompilableQuery and maps through mapModel, returning Model[].
     Alternatively, the DAL query functions themselves could be wrapped to include the
     mapping information so consumers don't need to know about it.

     3. getContextSizeForThread test assertions are fragile

     In src/dal/chat-threads.test.ts, the tests changed from:
     const contextSize = await getContextSizeForThread(threadId)
     expect(contextSize).toBe(null)
     to:
     const contextSize = await getContextSizeForThread(threadId).get()
     expect(contextSize?.contextSize).toBeUndefined()

     The old test checked for null (no thread found). The new test calls .get() which
     returns a single row or undefined, then checks contextSize?.contextSize. When the
     thread does not exist, .get() returns undefined, so contextSize?.contextSize is
     undefined -- this technically passes but conflates "thread not found" with "thread
     exists but contextSize is null". Both would satisfy toBeUndefined(). The semantic
     meaning of the test has shifted.

     Fix: For the "thread does not exist" case, assert expect(contextSize).toBeUndefined()
     (i.e., the row itself is undefined). For the "thread exists but no context size" case,
     assert expect(contextSize).toBeDefined() then
     expect(contextSize?.contextSize).toBeNull().

     4. getDevice lost its return type narrowing

     In src/dal/devices.ts, getDevice went from returning Promise<Device | null> to
     returning a raw Drizzle query. Callers now do data[0] ?? null themselves (e.g.,
     use-powersync-credentials-invalid-listener.ts:61). But the DAL function no longer
     carries any Device type information -- the return type is the raw Drizzle select
     result, which is a ModelRow-like object, not Device.

     Compare to getAllDevices and getAllChatThreads which have the explicit as typeof query
     & { execute: () => Promise<Device[]> } cast. getDevice is missing this type narrowing.

     Fix: Add the same type assertion pattern to getDevice:
     export const getDevice = (deviceId: string) => {
       const query =
     DatabaseSingleton.instance.db.select().from(devicesTable).where(eq(devicesTable.id,
     deviceId))
       return query as typeof query & { execute: () => Promise<Device[]> }
     }

     5. getAllEnabledTriggers cast in use-trigger-scheduler.ts is a bandaid

     const triggers = (await getAllEnabledTriggers()) as Trigger[]

     The DAL function getAllEnabledTriggers already has the type assertion & { execute: ()
     => Promise<Trigger[]> }, so awaiting it should return Trigger[] if the Drizzle .then()
     is wired correctly. The as Trigger[] cast here suggests the types don't flow through
     properly.

     Fix: This is the same root issue as #4 -- the type assertion on the query builders
     should make the as cast unnecessary. Investigate whether await getAllEnabledTriggers()
     actually resolves to Trigger[] with the current typing. If the & { execute } pattern
     doesn't propagate through await, a different approach may be needed (e.g., wrapping the
      query in a function that returns the correctly typed promise).

     6. itemNotEmpty filter is a behavior change, not just a refactor

     In src/dal/tasks.ts, the old code filtered empty items in JS after the query:
     return result.filter((task) => task.item && task.item.trim() !== '')

     The new code moves this to SQL:
     const itemNotEmpty = and(isNotNull(tasksTable.item), sql`trim(${tasksTable.item}) !=
     ''`)

     This is an improvement (filter at DB level). However, it's applied to
     getIncompleteTasksCount too, which the old code did NOT filter by itemNotEmpty. This
     means the count may now differ from what it used to be. If there were empty-item tasks
     in the DB, the old count included them but the old list didn't show them -- a
     pre-existing bug that this change silently fixes.

     Fix: No action needed -- this is actually a bug fix. But worth noting in the commit/PR
     message that this changes behavior: getIncompleteTasksCount now excludes empty-item
     tasks, matching what the list showed.

     7. getIncompleteTasks lost its Task[] return type

     getIncompleteTasks used to return Promise<Task[]>. Now it returns a raw Drizzle query
     builder with no type assertion (unlike getAllChatThreads, getAllDevices,
     getAllTriggersForPrompt, etc. which all have as typeof query & { execute: () =>
     Promise<X[]> }).

     In src/tasks/index.tsx:279:
     query: toCompilableQuery(getIncompleteTasks(debouncedSearchQuery)) as
     CompilableQuery<Task>,

     The consumer has to add as CompilableQuery<Task> because the DAL function doesn't carry
      the type.

     Fix: Add the same type assertion to getIncompleteTasks and getIncompleteTasksCount that
      the other DAL functions have, then remove the as CompilableQuery<Task> cast at the
     call site.

     8. placeholderData may not be supported by PowerSync's useQuery

     Three components pass placeholderData: (previousData) => previousData to
     @powersync/tanstack-react-query's useQuery:
     - src/layout/sidebar/index.tsx:65
     - src/tasks/index.tsx:280
     - src/automations/index.tsx:54

     The @powersync/tanstack-react-query useQuery wraps TanStack's useQuery but passes its
     own options. If placeholderData is not forwarded to the underlying TanStack query, it
     will be silently ignored, causing UI flicker when query keys change (e.g., during
     search).

     Fix: Verify at runtime that placeholderData works with PowerSync's useQuery. If it
     doesn't, the previous data retention logic needs to be handled differently (e.g., with
     local state).

     ---
     Code that looks good

     - Removing usePowerSyncInvalidation and all the manual queryClient.invalidateQueries
     calls is the right call. Live queries are the proper solution.
     - The PowerSyncContext.Provider placement in app.tsx (before QueryClientProvider) is
     correct.
     - The vite.config.ts dedupe for @powersync/common, @powersync/react, and react is
     appropriate.
     - Moving the empty-item filter into SQL in tasks.ts is cleaner.

vite.config.ts
raivieiraadriano92: it should not be here anymore - just removed

src/dal/settings.test.ts
raivieiraadriano92: Hey, we could keep the same as before - however use an object here wouldn't have any practical utility once in this method we need and actually use only the `keys`.

Also, this method is used exclusively in `useSettings` which accept an object where we can provide the default values for each key.

Moving back to the previous one would keep only the `reading`.

Just want to double check it should be reverted.

src/dal/settings.ts
raivieiraadriano92: same explanation as above - just want to double check if we should revert this.

src/widgets/citation/widget.test.tsx
raivieiraadriano92: In this case it actually depends on DB due the internal usage of `useSettings`.
With this refactor, `useSettings` uses `useQuery` from PowerSync which requires `PowerSyncContext.Provider`. This last one requires a PowerSync instance which depends on the DB.

raivieiraadriano92:
> Claude using /thunderimprove skill (without making the changes itself) came back with:
> 
> ```
>  Summary
> 
>  This branch migrates from manual React Query cache invalidation to PowerSync's
>  @powersync/tanstack-react-query live queries. DAL functions are converted from async
>  functions returning data to synchronous functions returning Drizzle query builders,
>  which are then passed through toCompilableQuery() for PowerSync-powered reactive
>  updates. The usePowerSyncInvalidation hook is removed entirely.
> 
>  The overall direction is correct and is a significant improvement -- live queries are
>  the right architectural approach. However, there are several concrete issues.
> 
>  ---
>  Issues Found
> 
>  1. Unsafe as unknown as cast in app.tsx (line 151-152)
> 
>  value={powerSyncInstance as unknown as ComponentProps<typeof
>  PowerSyncContext.Provider>['value']}
> 
>  This double cast (as unknown as) silences a type mismatch between PowerSyncDatabase
>  (from @powersync/web) and the context value expected by @powersync/react. This is a
>  bandaid for a version mismatch between @powersync/common versions. The vite.config.ts
>  dedupe addition partially addresses this, but the cast is a red flag that the dedupe
>  may not be sufficient at the type level. If the runtime instances actually differ, live
>   queries will silently fail.
> 
>  Fix: Verify that dedupe in vite resolves the runtime context correctly, and if so, the
>  cast should become unnecessary. If the types genuinely don't match, the PowerSync
>  version pinning needs alignment.
> 
>  2. Repeated data.map(mapModel) pattern -- missing reusable hook
> 
>  mapModel is called in 6 different places with the same useMemo(() =>
>  data.map(mapModel), [data]) pattern:
>  - src/settings/models/detail.tsx:75
>  - src/settings/models/layout.tsx:19
>  - src/settings/models/index.tsx:254
>  - src/automations/automation-form-modal.tsx:69
>  - src/chats/use-hydrate-chat-store.ts:99 and :148
> 
>  This mapping is needed because the DAL now returns raw ModelRow objects instead of
>  Model objects. Every consumer has to know about mapModel. This is a DRY violation and a
>   correctness risk -- if someone forgets mapModel, they'll get a row missing api_key and
>   is_system fields.
> 
>  Fix: Create a small hook like useModelsQuery(queryBuilder) that internally calls
>  useQuery + toCompilableQuery and maps through mapModel, returning Model[].
>  Alternatively, the DAL query functions themselves could be wrapped to include the
>  mapping information so consumers don't need to know about it.
> 
>  3. getContextSizeForThread test assertions are fragile
> 
>  In src/dal/chat-threads.test.ts, the tests changed from:
>  const contextSize = await getContextSizeForThread(threadId)
>  expect(contextSize).toBe(null)
>  to:
>  const contextSize = await getContextSizeForThread(threadId).get()
>  expect(contextSize?.contextSize).toBeUndefined()
> 
>  The old test checked for null (no thread found). The new test calls .get() which
>  returns a single row or undefined, then checks contextSize?.contextSize. When the
>  thread does not exist, .get() returns undefined, so contextSize?.contextSize is
>  undefined -- this technically passes but conflates "thread not found" with "thread
>  exists but contextSize is null". Both would satisfy toBeUndefined(). The semantic
>  meaning of the test has shifted.
> 
>  Fix: For the "thread does not exist" case, assert expect(contextSize).toBeUndefined()
>  (i.e., the row itself is undefined). For the "thread exists but no context size" case,
>  assert expect(contextSize).toBeDefined() then
>  expect(contextSize?.contextSize).toBeNull().
> 
>  4. getDevice lost its return type narrowing
> 
>  In src/dal/devices.ts, getDevice went from returning Promise<Device | null> to
>  returning a raw Drizzle query. Callers now do data[0] ?? null themselves (e.g.,
>  use-powersync-credentials-invalid-listener.ts:61). But the DAL function no longer
>  carries any Device type information -- the return type is the raw Drizzle select
>  result, which is a ModelRow-like object, not Device.
> 
>  Compare to getAllDevices and getAllChatThreads which have the explicit as typeof query
>  & { execute: () => Promise<Device[]> } cast. getDevice is missing this type narrowing.
> 
>  Fix: Add the same type assertion pattern to getDevice:
>  export const getDevice = (deviceId: string) => {
>    const query =
>  DatabaseSingleton.instance.db.select().from(devicesTable).where(eq(devicesTable.id,
>  deviceId))
>    return query as typeof query & { execute: () => Promise<Device[]> }
>  }
> 
>  5. getAllEnabledTriggers cast in use-trigger-scheduler.ts is a bandaid
> 
>  const triggers = (await getAllEnabledTriggers()) as Trigger[]
> 
>  The DAL function getAllEnabledTriggers already has the type assertion & { execute: ()
>  => Promise<Trigger[]> }, so awaiting it should return Trigger[] if the Drizzle .then()
>  is wired correctly. The as Trigger[] cast here suggests the types don't flow through
>  properly.
> 
>  Fix: This is the same root issue as #4 -- the type assertion on the query builders
>  should make the as cast unnecessary. Investigate whether await getAllEnabledTriggers()
>  actually resolves to Trigger[] with the current typing. If the & { execute } pattern
>  doesn't propagate through await, a different approach may be needed (e.g., wrapping the
>   query in a function that returns the correctly typed promise).
> 
>  6. itemNotEmpty filter is a behavior change, not just a refactor
> 
>  In src/dal/tasks.ts, the old code filtered empty items in JS after the query:
>  return result.filter((task) => task.item && task.item.trim() !== '')
> 
>  The new code moves this to SQL:
>  const itemNotEmpty = and(isNotNull(tasksTable.item), sql`trim(${tasksTable.item}) !=
>  ''`)
> 
>  This is an improvement (filter at DB level). However, it's applied to
>  getIncompleteTasksCount too, which the old code did NOT filter by itemNotEmpty. This
>  means the count may now differ from what it used to be. If there were empty-item tasks
>  in the DB, the old count included them but the old list didn't show them -- a
>  pre-existing bug that this change silently fixes.
> 
>  Fix: No action needed -- this is actually a bug fix. But worth noting in the commit/PR
>  message that this changes behavior: getIncompleteTasksCount now excludes empty-item
>  tasks, matching what the list showed.
> 
>  7. getIncompleteTasks lost its Task[] return type
> 
>  getIncompleteTasks used to return Promise<Task[]>. Now it returns a raw Drizzle query
>  builder with no type assertion (unlike getAllChatThreads, getAllDevices,
>  getAllTriggersForPrompt, etc. which all have as typeof query & { execute: () =>
>  Promise<X[]> }).
> 
>  In src/tasks/index.tsx:279:
>  query: toCompilableQuery(getIncompleteTasks(debouncedSearchQuery)) as
>  CompilableQuery<Task>,
> 
>  The consumer has to add as CompilableQuery<Task> because the DAL function doesn't carry
>   the type.
> 
>  Fix: Add the same type assertion to getIncompleteTasks and getIncompleteTasksCount that
>   the other DAL functions have, then remove the as CompilableQuery<Task> cast at the
>  call site.
> 
>  8. placeholderData may not be supported by PowerSync's useQuery
> 
>  Three components pass placeholderData: (previousData) => previousData to
>  @powersync/tanstack-react-query's useQuery:
>  - src/layout/sidebar/index.tsx:65
>  - src/tasks/index.tsx:280
>  - src/automations/index.tsx:54
> 
>  The @powersync/tanstack-react-query useQuery wraps TanStack's useQuery but passes its
>  own options. If placeholderData is not forwarded to the underlying TanStack query, it
>  will be silently ignored, causing UI flicker when query keys change (e.g., during
>  search).
> 
>  Fix: Verify at runtime that placeholderData works with PowerSync's useQuery. If it
>  doesn't, the previous data retention logic needs to be handled differently (e.g., with
>  local state).
> 
>  ---
>  Code that looks good
> 
>  - Removing usePowerSyncInvalidation and all the manual queryClient.invalidateQueries
>  calls is the right call. Live queries are the proper solution.
>  - The PowerSyncContext.Provider placement in app.tsx (before QueryClientProvider) is
>  correct.
>  - The vite.config.ts dedupe for @powersync/common, @powersync/react, and react is
>  appropriate.
>  - Moving the empty-item filter into SQL in tasks.ts is cleaner.
> ```

Items reviewed and addressed:

Changes Summary

Issue 1 — app.tsx uses as AbstractPowerSyncDatabase single cast

Issue 2 — Eliminated repeated data.map(mapModel) pattern:           
- Created src/hooks/use-models.ts — new useModelsQuery hook wrapping useQuery + toCompilableQuery + mapModel
- Added mapModels helper to src/dal/models.ts for imperative callers
- Updated src/dal/index.ts to export mapModels
- Updated 4 React component consumers: detail.tsx, layout.tsx, index.tsx (models settings), automation-form-modal.tsx
- Updated 2 imperative consumers in src/chats/use-hydrate-chat-store.ts
- Removed unused imports (useMemo, mapModel, useQuery, toCompilableQuery) from updated files

Issue 3 — Fixed test assertion in src/dal/chat-threads.test.ts: assert expect(contextSize).toBeUndefined() (the row itself) instead of expect(contextSize?.contextSize).toBeUndefined()

Issue 4 — Added { execute: () => Promise<Device[]> } type assertion to getDevice in src/dal/devices.ts

Issue 5 — Kept as Trigger[] cast (.execute() doesn't help because Drizzle's native method type wins over the intersection). The cast is correct for enabled triggers.

Issue 6 — No code change (bug fix, note for PR description).

Issue 7 — Added type assertions to getIncompleteTasks and getIncompleteTasksCount in src/dal/tasks.ts. Removed as CompilableQuery<Task> cast and unused CompilableQuery import from src/tasks/index.tsx.

Issue 8 — No code change (placeholderData works correctly via rest-spread).

PR #463, opened 2026-03-16
THU-341: Desktop crash on update and restart

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Touches the desktop update/relaunch path and app bootstrap logic; mistakes could block startup or regress the in-app update flow, though changes are small and localized.
> 
> **Overview**
> Fixes desktop update relaunch issues by doing a best-effort PowerSync disconnect before `relaunch()` and setting a `localStorage` flag to force a clean post-update navigation state.
> 
> Adds `post-update-redirect` logic (with tests) that, on next startup, consumes the flag and redirects to `/` to avoid the WebView restoring a stale route; `src/index.tsx` now skips normal initialization when redirecting.
> 
> Introduces a Bun-based local Tauri update server plus documentation for end-to-end testing of in-app updates against a localhost “CDN” (manifest + bundle + `.sig`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be009beb392a0cfd54734bc7ccd98c047db38e8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

 
 
 
 
 
 
 
 
 
 
 
 **PR Summary by Typo**
------------

#### Overview
This PR addresses a desktop crash on update and restart by introducing a robust mechanism for handling post-update redirects and providing tools for local update testing. The changes ensure a clean application state after an update and relaunch, preventing issues caused by stale routes.

#### Key Changes
- Added comprehensive documentation (`docs/local-cdn-for-app-update-testing.md`) for setting up a local CDN to test in-app updates.
- Implemented a local update server script (`scripts/local-update-server.ts`) to simulate the production CDN for testing purposes.
- Modified the `useDesktopUpdate` hook to disconnect PowerSync gracefully before relaunch and to set a flag signaling a post-update redirect.
- Introduced `post-update-redirect.ts` to manage a flag in local storage, which forces a redirect to the root path after an update and relaunch if the WebView restores a stale route.
- Updated `src/index.tsx` to utilize the new post-update redirect logic, ensuring the app starts cleanly after an update.

#### Work Breakdown

| Category    | Lines Changed |
|-------------|---------------|
| New Work    | 302 (99.0%)   |
| Rework      | 3 (1.0%)      |
| Total Changes | 305         | 

 <h6>To turn off PR summary, please visit <a href="https://app.typoapp.io/settings/notification?tab=codeHealth">Notification settings</a>.</h6>

### DIFF
--- a/scripts/local-update-server.ts
+++ b/scripts/local-update-server.ts
@@ -1,0 +1,96 @@
+/**
+ * Local update server for testing Tauri updater flow.
+ *
+ * Serves update manifests and bundles from the release build output.
+ * Point the old build's updater endpoint to http://localhost:8888/update/{{target}}-{{arch}}/{{current_version}}
+ *
+ * Usage:
+ *   1. Build the new version: TAURI_SIGNING_PRIVATE_KEY=~/.tauri/test-update.key bun tauri build
+ *   2. Start this server: bun run scripts/local-update-server.ts
+ *   3. Open the old build and trigger the update
+ */
+
+const BUNDLE_DIR = new URL('../src-tauri/target/release/bundle', import.meta.url).pathname
+const PORT = 8888
+
+const server = Bun.serve({
+  port: PORT,
+  async fetch(req) {
+    const url = new URL(req.url)
+    console.log(`${req.method} ${url.pathname}`)
+
+    // Tauri updater hits: GET /update/{target}-{arch}/{current_version}
+    // It expects a JSON response with the update info, or 204 if no update
+    if (url.pathname.startsWith('/update/')) {
+      const parts = url.pathname.split('/')
+      const platform = parts[2] // e.g. "darwin-aarch64"
+      const currentVersion = parts[3]
+
+      // Find the .tar.gz update bundle and its .sig file
+      const glob = new Bun.Glob('macos/*.tar.gz')
+      const bundles = Array.from(glob.scanSync(BUNDLE_DIR))
+
+      if (bundles.length === 0) {
+        console.error('No update bundles found in', BUNDLE_DIR + '/macos/')
+        return new Response('No bundles found', { status: 500 })
+      }
+
+      const bundleName = bundles[0]
+      const bundlePath = `${BUNDLE_DIR}/${bundleName}`
+      const sigPath = `${bundlePath}.sig`
+
+      // Read signature
+      let signature: string
+      try {
+        signature = await Bun.file(sigPath).text()
+      } catch {
+        console.error('Signature file not found:', sigPath)
+        return new Response('Signature not found', { status: 500 })
+      }
+
+      // Read version from tauri.conf.json
+      const confPath = new URL('../src-tauri/tauri.conf.json', import.meta.url).pathname
+      const conf = await Bun.file(confPath).json()
+      const newVersion = conf.version
+
+      console.log(`Current: ${currentVersion}, Available: ${newVersion}, Bundle: ${bundleName}`)
+
+      if (currentVersion === newVersion) {
+        return new Response(null, { status: 204 })
+      }
+
+      const manifest = {
+        version: newVersion,
+        notes: `Update to ${newVersion}`,
+        pub_date: new Date().toISOString(),
+        url: `http://localhost:${PORT}/bundle/${bundleName}`,
+        signature,
+      }
+
+      console.log('Serving update manifest:', JSON.stringify(manifest, null, 2))
+      return Response.json(manifest)
+    }
+
+    // Serve the actual bundle file
+    if (url.pathname.startsWith('/bundle/')) {
+      const filePath = `${BUNDLE_DIR}/${url.pathname.replace('/bundle/', '')}`
+      const file = Bun.file(filePath)
+
+      if (!(await file.exists())) {
+        console.error('Bundle not found:', filePath)
+        return new Response('Not found', { status: 404 })
+      }
+
+      console.log('Serving bundle:', filePath, `(${(file.size / 1024 / 1024).toFixed(1)}MB)`)
+      return new Response(file, {
+        headers: { 'Content-Type': 'application/gzip' },
+      })
+    }
+
+    return new Response('Not found', { status: 404 })
+  },
+})
+
+console.log(`Local update server running at http://localhost:${PORT}`)
+console.log(`Bundle dir: ${BUNDLE_DIR}`)
+console.log(`\nWaiting for update requests...`)
--- a/scripts/local-update-server.ts
+++ b/scripts/local-update-server.ts
@@ -73,7 +73,12 @@ const server = Bun.serve({
 
     // Serve the actual bundle file
     if (url.pathname.startsWith('/bundle/')) {
-      const filePath = `${BUNDLE_DIR}/${url.pathname.replace('/bundle/', '')}`
+      const relativePath = url.pathname.substring('/bundle/'.length)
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -5,6 +5,15 @@ import './polyfills'
...
 import { initializeLinkInterception } from './lib/intercept-links'
 
+// After an update+relaunch, the WebView may restore a stale route (e.g. /waitlist
+// verify screen). Detect this and force a clean start at root.
+if (localStorage.getItem('thunderbolt_post_update')) {
+  localStorage.removeItem('thunderbolt_post_update')
+  if (window.location.pathname !== '/') {
+    window.location.replace('/')
+  }
+}
```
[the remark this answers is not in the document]
darkbanjo: @cursor push 775a80e55a

src/hooks/use-desktop-update.ts
```diff
@@ -112,6 +113,14 @@ export const useDesktopUpdate = (): DesktopUpdateState => {
 
   const restartApp = useCallback(async () => {
     try {
+      // Best-effort disconnect — don't block the relaunch if PowerSync fails
+      try {
+        await getPowerSyncInstance()?.disconnect()
+      } catch {
+        /* proceed anyway */
```
cjroth: We need to at least console.error here. This will be painful someday

src/index.tsx
```diff
@@ -5,8 +5,22 @@ import './polyfills'
 import './index.css'
 import { initializeLinkInterception } from './lib/intercept-links'
 
-initializeLinkInterception()
+// After an update+relaunch, the WebView may restore a stale route (e.g. /waitlist
+// verify screen). Detect this and force a clean start at root.
+const postUpdateFlag = localStorage.getItem('thunderbolt_post_update')
```
cjroth: This should go into a util function and we should have unit tests for it. Also we should use the local storage hook.

src/hooks/use-desktop-update.ts
```diff
@@ -112,6 +113,14 @@ export const useDesktopUpdate = (): DesktopUpdateState => {
...
+      }
+      // Signal the new process to reset navigation (WebView may restore stale route)
+      localStorage.setItem('thunderbolt_post_update', '1')
```
cjroth: Would our local storage hook work here?

cjroth (changes requested):
Left a few comments. This seems fine - my main concern is that this is an edge case that is hard to reproduce - how do we know that it works and doesn't break anything, and how do we know that we don't break this logic? I think having tests to ensure this would be good.

src/hooks/use-desktop-update.ts
darkbanjo: ⚡ Good call — added `console.error` in the latest push.

src/index.tsx
darkbanjo: ⚡ Done — extracted into `src/lib/post-update-redirect.ts` with `setPostUpdateFlag()` and `handlePostUpdateRedirect()`, plus unit tests. Re: the local storage hook — since this runs outside React (at the top of `index.tsx` before render), we can't use a React hook here. The util uses raw localStorage directly.

src/hooks/use-desktop-update.ts
darkbanjo: ⚡ The `useLocalStorage` hook wouldn't work here since `restartApp` is a callback inside `useCallback`, not a hook call site. We now use `setPostUpdateFlag()` from the extracted util instead of raw `localStorage.setItem`, which keeps the key centralized.

src/lib/post-update-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+    return false
+  }
+
+  localStorage.removeItem(postUpdateKey)
+
+  if (window.location.pathname !== '/') {
+    window.location.replace('/')
+    return true
+  }
+
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Per project guidelines we prefer optimistic code — localStorage is available in all our target environments (desktop Tauri WebView and modern browsers). Adding try/catch here would be defensive coding that obscures intent.

src/lib/post-update-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+ * current pathname isn't "/", redirects to root via `window.location.replace`
+ * and returns true (caller should skip normal initialization). Otherwise returns false.
+ */
+export const handlePostUpdateRedirect = (): boolean => {
+  const flag = localStorage.getItem(postUpdateKey)
+  if (!flag) {
+    return false
+  }
+
+  localStorage.removeItem(postUpdateKey)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ The function already has the correct return statements — returns `true` when redirecting, `false` otherwise. The suggested code is identical to what's already there.

src/lib/post-update-redirect.test.ts
```diff
@@ -0,0 +1,65 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
```
cjroth: Oh this should be bun test

cjroth: (and maybe we should update agents.md for that)

cjroth (changes requested):

cjroth (commented):
Also can we add the Notion doc to docs folder in the report for this?

src/lib/post-update-redirect.test.ts
darkbanjo: ⚡ Good call — this was already migrated to bun:test in commit 8b8f9d89. The test file imports from `bun:test` and runs with `bun test`. AGENTS.md already says to use `bun test` instead of `vitest`, so no update needed there.

src/hooks/use-desktop-update.ts
```diff
@@ -112,6 +114,14 @@ export const useDesktopUpdate = (): DesktopUpdateState => {
...
   const restartApp = useCallback(async () => {
     try {
+      // Best-effort disconnect — don't block the relaunch if PowerSync fails
+      try {
+        await getPowerSyncInstance()?.disconnect()
+      } catch (err) {
+        console.error('Failed to disconnect PowerSync before relaunch:', err)
+      }
+      // Signal the new process to reset navigation (WebView may restore stale route)
+      setPostUpdateFlag()
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in the latest push. `clearPostUpdateFlag()` is now called in the catch block so a failed relaunch won't leave an orphaned flag.

scripts/local-update-server.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // Serve the actual bundle file
+    if (url.pathname.startsWith('/bundle/')) {
+      const filePath = `${BUNDLE_DIR}/${url.pathname.replace('/bundle/', '')}`
+      const file = Bun.file(filePath)
+
+      if (!(await file.exists())) {
+        console.error('Bundle not found:', filePath)
+        return new Response('Not found', { status: 404 })
+      }
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Adopted the suggestion — path traversal guard added in the latest push.

cjroth (approved):
code looks good to me but want @arienemaiara's review first

arienemaiara (approved):
LGTM

PR #482, opened 2026-03-19
UI Refinements

cjroth:
Continuation of #476 with clean history.

> [!NOTE]
> **Medium Risk**
> Touches CI workflows and adds an automated PR-commenting script (plus baseline caching), which could fail builds or spam/update PR comments if misconfigured. The rest is mostly UI refactors and documentation, but spread across several areas.
> 
> **Overview**
> Adds a new **PR Metrics** GitHub Action that computes prod-code line deltas, bundle size (via `size-limit`), test coverage, and Lighthouse performance against the Render preview, then posts/updates a `## PR Metrics` comment via a new `.github/scripts/post-pr-metrics.cjs` script (with bun tests). Main CI now persists a `.metrics-baseline/metrics.json` cache on merges to `main` so PRs can show deltas.
> 
> Expands Thunderbot ops: introduces `.thunderbot/references/*` (implementation/review/PR workflow/commit conventions/subagent/team orchestration), updates Thunderbot docs and skills (`thunderbot.md`, `thunderimprove.md`, `thunderfix.md`, `thunderpush.md`), adds a Bun-based `.thunderbot/cli.ts` with PR/CI subcommands, and updates `.gitignore`/`setup.sh` to track/validate the new rules/references.
> 
> Includes product changes: refactors Automations UI (new search UX, prompt/model input changes, card/layout tweaks), adjusts chat layout/scroll/padding and citation popover anchoring (pass element instead of `DOMRect`), extends widget parser tests for `[N]` citations + widget tags, adds a local Tauri updater test server (`scripts/local-update-server.ts`) with documentation, and bumps app version to `0.1.71`.


src/components/ui/mode-selector/mode-selector.tsx
```diff
@@ -41,15 +43,14 @@ const createModeGroups = (modes: Mode[]): SearchableMenuGroup<ModeItemData>[] =>
...
   const renderTrigger = (selected: SearchableMenuItem<ModeItemData> | undefined, isOpen: boolean) => (
     <div
       className={cn(
-        'flex items-center rounded-lg cursor-pointer transition-colors text-[length:var(--font-size-sm)] border border-border',
-        iconOnly
-          ? 'size-[var(--touch-height-default)] justify-center'
-          : 'gap-1 px-[var(--spacing-x-sm)] py-[var(--spacing-y-sm)]',
+        'flex items-center rounded-lg cursor-pointer transition-colors text-[length:var(--font-size-body)] border border-border',
+        iconOnly ? 'size-[var(--touch-height-sm)] justify-center' : 'gap-2 px-3 h-[var(--touch-height-default)]',
         isOpen ? 'bg-secondary' : 'hover:bg-secondary/50',
```
raivieiraadriano92: The hover effect here it's not noticeable - probably because the prompt input now uses `bg-secondary` too.

src/components/ui/combobox.tsx
```diff
@@ -0,0 +1,152 @@
...
+      setSearch('')
+    }
+  }, [open])
+
+  useEffect(() => {
+    if (open) {
+      // Focus the search input after the popover opens
+      requestAnimationFrame(() => inputRef.current?.focus())
+    }
+  }, [open])
```
raivieiraadriano92: I think this is a good example where we could avoid the `useEffect`. We could replace both with:

```ts
 const handleOpenChange = (nextOpen: boolean) => {                                                                                                        
    setOpen(nextOpen)
    if (nextOpen) {                                                                                                                                        
      requestAnimationFrame(() => inputRef.current?.focus())                                                                                               
    } else {                                                                                                                                               
      setSearch('')                                                                                                                                        
    }                                                                                                                                                      
  } 
```

and then replace `onOpenChange={setOpen}` with `onOpenChange={handleOpenChange}`.

src/components/location-search-combobox.tsx
```diff
@@ -0,0 +1,106 @@
...
+  onSelect,
+  placeholder = 'Select location...',
+  autoOpen = false,
+  className,
+}: LocationSearchComboboxProps) => {
+  const locationSearch = useLocationSearch()
+  const buttonRef = useRef<HTMLButtonElement>(null)
+  const searchInputRef = useRef<HTMLInputElement>(null)
+
+  useEffect(() => {
```
raivieiraadriano92: This `useEffect` could be removed as well and with it the refs `buttonRef` and `searchInputRef`.

To achieve that we'd need to initialize `useLocationSearch` with `autoOpen` - something like this:

```ts
const locationSearch = useLocationSearch({autoOpen})
```

and then initialize the initial state of `open` inside of useLocationSearch based on the new prop `autoOpen`.

raivieiraadriano92 (changes requested):
LGTM
I left just a few comments related to the useEffects

src/components/sidebar-footer.tsx
```diff
@@ -21,19 +24,66 @@ import {
...
+    )
+  }
+
+  return (
+    <button type="button" onClick={onClick} className={className}>
+      {icon}
+      <span>{label}</span>
+    </button>
+  )
+}
```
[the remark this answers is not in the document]
cjroth: ⚡ This was already addressed in a prior commit — `onNavigate={handleMenuNavigate}` is passed to `AccountMenuItemButton` for both Dev Settings and Message Simulator, and `handleMenuNavigate` calls `setMenuOpen(false)`. The popover closes on navigation.

src/layout/sidebar/sidebar-header.tsx
```diff
@@ -54,13 +54,17 @@ export const SidebarHeader = ({ onToggle }: SidebarHeaderProps) => {
       {isExpanded && (
         <div className="flex items-center">
           {isMobile ? (
-            <SidebarCloseButton onClick={onToggle} />
+            <PowerSyncStatus />
```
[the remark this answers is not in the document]
cjroth: ⚡ Good catch — restored the `SidebarCloseButton` alongside `PowerSyncStatus` in the mobile sidebar header.

src/components/ui/mode-selector/mode-selector.tsx
cjroth: ⚡ Good call — switched from `bg-secondary` to `bg-accent` for the hover/active states so they're visually distinct from the prompt input background.

src/components/ui/combobox.tsx
cjroth: ⚡ Good call — done in the latest push. Replaced both `useEffect` hooks with a single `handleOpenChange` callback.

src/components/location-search-combobox.tsx
cjroth: ⚡ Done — `useLocationSearch` now accepts `{ autoOpen }` and initializes the `open` state accordingly. Removed the `useEffect`, `buttonRef`, and `searchInputRef` from `LocationSearchCombobox`.

src/layout/sidebar/index.tsx
```diff
@@ -97,9 +97,12 @@ export default function Sidebar() {
...
   })
 
-  const handleRename = (threadId: string, title: string) => {
-    renameChatMutation.mutate({ id: threadId, title })
-  }
+  const handleRename = useCallback(
+    (threadId: string, title: string) => {
+      renameChatMutation.mutate({ id: threadId, title })
+    },
+    [renameChatMutation],
```
ital0: the `memo()` on `ChatListItem` is a nice addition, just needs the callback to cooperate.

I think can depend on `renameChatMutation.mutate` instead of the whole object, or use a ref to stabilize the callback.

src/settings/models/index.tsx
```diff
@@ -398,40 +369,12 @@ export default function ModelsPage() {
 
   const handleDialogOpenChange = (open: boolean) => {
```
ital0, line 371: I see this method has been significantly refactored, removing some calls like `form.reset` and `form.clearErrors`.

Do you think it worth to quick sanity check to confirm if users don't see red error text from a previous attempt when they reopen the dialog fresh?

src/components/ui/page-search.tsx
```diff
@@ -0,0 +1,79 @@
...
+  delay?: number
+  tooltip?: string
+}
+
+/**
+ * A search toggle button + collapsible search input for page-level search.
+ * Renders a ghost search icon button (to place in a PageHeader) and a
+ * slide-down SearchInput that appears when toggled.
+ */
+export const usePageSearch = ({
```
ital0: This hook is returning pre-built JSX elements, which inverts the logics of the patterns we've been using.

Not a blocker, it gets the job done. But do you think we can improve this and split a proper `PageSearch` component?

src/lib/constants.ts
```diff
@@ -7,3 +7,9 @@ export const privacyPolicyUrl = 'https://www.thunderbird.net/en-US/privacy/'
 
 /** URL to Mozilla terms of service */
 export const termsOfServiceUrl = 'https://www.mozilla.org/en-US/about/legal/terms/mozilla/'
+
+/** Tailwind class for the responsive default icon size */
+export const iconSizeClass = 'size-[var(--icon-size-default)]'
```
ital0: is there an alternative for this class?

ital0 (changes requested):
huge change! Check the details I've noticed

src/layout/sidebar/index.tsx
cjroth: ⚡ Good call — extracted `renameChatMutation.mutate` into a stable reference so the `useCallback` dependency doesn't change on every render. Done in the latest push.

src/settings/models/index.tsx
cjroth: ⚡ Good catch — added an explicit `form.clearErrors()` call alongside `form.reset()` in the `useEffect` that fires when the dialog closes. This ensures any lingering validation messages are cleaned up even if `reset()` alone doesn't clear them in edge cases.

src/components/ui/page-search.tsx
cjroth: ⚡ Agreed — refactored `usePageSearch` into a proper `PageSearch` compound component with `PageSearch.Button` and `PageSearch.Input` sub-components backed by React context. Much cleaner pattern.

src/lib/constants.ts
cjroth: ⚡ Removed the constant and inlined `size-[var(--icon-size-default)]` directly where it was used. It was just a Tailwind class string — no need for an abstraction.

src/components/ui/combobox.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    {value === item.id && <CheckIcon className="size-4" />}
+                  </span>
+                </button>
+              ))}
+          </div>
+        </PopoverPrimitive.Content>
+      </PopoverPrimitive.Portal>
+    </PopoverPrimitive.Root>
+  )
+}
```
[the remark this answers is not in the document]
cjroth: ⚡ This is an automated bot comment about the Combobox component — it serves a different use case than SearchableMenu (form-integrated selection vs standalone dropdown). No action needed.

src/components/sidebar-footer.tsx
```diff
@@ -48,122 +98,200 @@ export const SidebarFooter = ({ className }: SidebarFooterProps) => {
...
+  const showBlur = isMobile && menuOpen
+
+  // Capture trigger position when blur activates so we can render a clone above it
+  useEffect(() => {
+    if (showBlur && triggerRef.current) {
+      setTriggerRect(triggerRef.current.getBoundingClientRect())
+    } else {
+      setTriggerRect(null)
+    }
+  }, [showBlur])
```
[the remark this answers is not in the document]
cjroth: ⚡ The triggerRect is only used momentarily during the blur animation transition and the element is removed when the animation completes. The brief window where a layout shift could cause misalignment is negligible. No action needed.

src/automations/index.tsx
```diff
@@ -107,40 +107,42 @@ export default function AutomationsPage() {
...
+                  setIsCreateModalOpen(true)
+                  trackEvent('automation_modal_create_open')
+                }}
+              >
+                <Plus className="h-4 w-4" />
+              </Button>
+            </PageHeader>
+
+            <PageSearch.Input placeholder="Search automations..." onSearch={setDebouncedSearchQuery} />
+          </PageSearch>
```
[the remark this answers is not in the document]
cjroth: ⚡ The PageSearch compound component is intentionally designed this way — the parent onSearch handles cleanup on close, while PageSearch.Input onSearch handles typing. This is standard compound component API design. No action needed.

PR #483, opened 2026-03-19
Rename chat fix for Safari/Chrome/Mobile keyboard

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Moderate UI-behavior change: renaming flow is reworked and introduces new state/optimistic title handling in `ChatListItem`, which could affect navigation and render timing across browsers.
> 
> **Overview**
> Sidebar chat renaming is refactored to use a new modal `RenameChatDialog` (input auto-focus/select, Save/Cancel, Enter-to-save, trim/"New Chat" fallback) instead of inline editing inside `ChatListItem`.
> 
> `ChatListItem` now always navigates on item click, opens the rename dialog from the menu, and keeps an optimistic title locally until the thread title prop catches up; the old `useChatListItemState` hook and its unit tests are removed and replaced with dialog-focused tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 341cf230b9a729169385f607628d0a9e65ad4c74. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

 
 
 
 
 
 
 
 **PR Summary by Typo**
------------

#### Overview
This PR refactors the chat renaming functionality by replacing the inline editing experience with a dedicated `RenameChatDialog` component. This change aims to improve the user experience and address potential issues with keyboard interactions on various browsers and mobile devices.

#### Key Changes
- Replaced the inline chat renaming input within `ChatListItem` with a new `RenameChatDialog` component.
- Introduced `rename-chat-dialog.tsx` and its corresponding test file `rename-chat-dialog.test.tsx`.
- Removed the `useChatListItemState` hook and its associated test file, as its functionality is now handled by the new dialog.
- Updated `ChatListItem` to manage the dialog's open state and implement optimistic title updates.
- Modified `chat-list-item.test.tsx` to remove tests related to inline editing and update navigation tests.

#### Work Breakdown

| Category    | Lines Changed |
|-------------|---------------|
| New Work    | 142 (34.4%)   |
| Churn       | 202 (48.9%)   |
| Rework      | 69 (16.7%)    |
| Total Changes | 413         | 

 <h6>To turn off PR summary, please visit <a href="https://app.typoapp.io/settings/notification?tab=codeHealth">Notification settings</a>.</h6>


src/layout/sidebar/chat-list-item.tsx
```diff
@@ -31,20 +32,21 @@ export const ChatListItem = ({
...
-    displayTitle,
-    inputRef,
-    handleRenameStart,
-    handleRenameSubmit,
-    handleRenameCancel,
-  } = useChatListItemState({
-    title: thread.title,
-    onRename: (title) => onRename(thread.id, title),
-  })
+  useEffect(() => {
```
raivieiraadriano92: Claude Code review for the useEffect implementation:

The useEffect on lines 38-42 is syncing optimisticTitle back to null once thread.title catches up. Per the React docs, this falls under "Adjusting state when a prop changes" — which is one of the cases where you don't need an effect.

You can eliminate it by computing the display title directly during render instead of storing+syncing it:

However, there's a subtlety: optimisticTitle needs to persist across renders until the prop catches up. The current pattern is essentially: "show optimistic value until the real prop matches it."
                                                                                                                                                           
The cleanest no-effect approach is to track the previous thread.title and clear the optimistic state during render:

```ts
// Replace:                                                                                                                                              
const [optimisticTitle, setOptimisticTitle] = useState<string | null>(null)                                                                              
                                                                                                                                                          
useEffect(() => {                                                                                                                                        
  if (optimisticTitle !== null && thread.title === optimisticTitle) {                                                                                    
    setOptimisticTitle(null)                                                                                                                           
  }
}, [thread.title, optimisticTitle])                                                                                                                      

// With:                                                                                                                                                 
const [optimisticTitle, setOptimisticTitle] = useState<string | null>(null)                                                                            
const [prevTitle, setPrevTitle] = useState(thread.title)
                                                                                                                                                          
if (thread.title !== prevTitle) {
  setPrevTitle(thread.title)                                                                                                                             
  if (optimisticTitle !== null && thread.title === optimisticTitle) {                                                                                  
    setOptimisticTitle(null)                                                                                                                             
  }                                                                                                                                                      
}                                                                                                                                                        
````

This is the pattern React docs recommend: "Store information from previous renders" — calling setState during render (not in an effect) so React re-renders immediately without the extra effect cycle.
                                                                                                                                                          
That said, this trades one state (optimisticTitle) + one effect for two states (optimisticTitle + prevTitle) + a render-time conditional setState. The complexity is arguably higher for the same result.
                                                                                                                                                          
My recommendation: The simplest approach here is actually to use useSyncExternalStore or just derive the value entirely. But given the current architecture (optimistic update that gets confirmed by a prop change), the existing useEffect is a reasonable and readable solution. The React docs acknowledge that while effects aren't ideal for prop-syncing, this is a lightweight case where the effect is clear and bounded — it doesn't cause cascading re-renders or stale state issues.

If you want to remove the effect, the prevTitle pattern above is the "correct" React way.

File: src/layout/sidebar/rename-chat-dialog.tsx [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+
+export const RenameChatDialog = ({ open, title, onOpenChange, onRename }: RenameChatDialogProps) => {
+  const [value, setValue] = useState('')
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  useEffect(() => {
+    if (open) {
+      setValue(title ?? 'New Chat')
+    }
+  }, [open, title])
```
raivieiraadriano92: The useEffect on lines 17-21 resets the input value to the current title every time the dialog opens. Per React docs, this is "Resetting state when a prop changes".                                                                                                                                           
                                                                                                                                                           
The cleanest fix: use key to reset the component. Instead of syncing state via effect, unmount/remount the dialog content when it opens with a new title.
But since this is a controlled Dialog that stays mounted, the more practical approach is the render-time state reset:

```ts
const [value, setValue] = useState('')                                                                                                                 
const [prevOpen, setPrevOpen] = useState(false)

if (open && !prevOpen) {                                                                                                                                 
  setValue(title ?? 'New Chat')
}                                                                                                                                                        
if (open !== prevOpen) {                                                                                                                               
  setPrevOpen(open)
}                                                                                                                                                        
````

This eliminates the effect — React processes the setState during the same render pass instead of scheduling a second render via effect.                  
                                                                                                                                                        
However, the even simpler approach is to not keep value as state at all and instead use an uncontrolled input with key:                                  

```ts
export const RenameChatDialog = ({ open, title, onOpenChange, onRename }: RenameChatDialogProps) => {                                                    
  const inputRef = useRef<HTMLInputElement>(null)                                                                                                        

  const handleSave = () => {                                                                                                                             
    const trimmed = (inputRef.current?.value ?? '').trim()                                                                                             
    const newTitle = trimmed || 'New Chat'                                                                                                               
    if (newTitle !== (title ?? 'New Chat')) {                                                                                                            
      onRename(newTitle)                                                                                                                                 
    }                                                                                                                                                    
    onOpenChange(false)                                                                                                                                
  }

  return (
    <Dialog open={open} onOpenChange={onOpenChange}>
      <DialogContent ...>                                                                                                                                
        ...                                                                                                                                              
        <Input                                                                                                                                           
          key={open ? 'open' : 'closed'}                                                                                                                 
          ref={inputRef}                                                                                                                                 
          defaultValue={title ?? 'New Chat'}                                                                                                             
          onKeyDown={(e) => { if (e.key === 'Enter') handleSave() }}                                                                                     
          placeholder="Chat name"                                                                                                                        
        />                                                                                                                                               
        ...                                                                                                                                              
      </DialogContent>                                                                                                                                   
    </Dialog>                                                                                                                                            
  )                                                                                                                                                    
}
````

This removes useState, useEffect, and the controlled input entirely. The key change when open toggles forces React to remount the input with the fresh   
defaultValue.
                                                                                                                                                          
My recommendation: The uncontrolled input + key approach is the simplest. But if you prefer keeping it controlled (e.g., for validation feedback as the  
user types), the render-time prevOpen pattern is the correct effect-free alternative.

raivieiraadriano92 (changes requested):
LGTM

I left a few suggestions to avoid the useEffect usage as recommended in https://react.dev/learn/you-might-not-need-an-effect.

I think it'd worth to try it - let me know if it makes sense

src/layout/sidebar/chat-list-item.tsx
darkbanjo: ⚡ Good call — replaced the useEffect with the render-time prevTitle pattern as recommended. Done in the latest push.

src/layout/sidebar/rename-chat-dialog.tsx
darkbanjo: ⚡ This was already addressed in a prior commit — the dialog now conditionally renders `RenameChatForm` with `{open && <RenameChatForm key={title} ... />}`, which remounts the component (and resets state) each time the dialog opens. No useEffect needed.

PR #508, opened 2026-03-23
[FRONTEND] THU-348: Get MCPs working

ital0:
## Summary
All MCP changes in a single PR for team validation. Use the desktop app (`bun tauri dev`) to test all features.

This PR contains the same code as PRs [1/6] through [6/6] combined, with all review fixes applied. Review individual PRs for focused diffs.

## Test Scenarios
- **stdio (Desktop):** npx + @playwright/mcp@latest --headless
- **HTTP + Bearer (Desktop):** https://mcp.render.com/mcp + Render API key
- **HTTP + No Auth (Web):** localhost MCP server via supergateway with --cors
- **Platform filtering:** stdio hidden on web/mobile, HTTP/SSE available everywhere

---

> [!NOTE]
> **High Risk**
> High risk because it introduces new credential storage/encryption, OAuth redirect handling, and desktop shell-based stdio process spawning, plus changes to proxy redirect handling for SSRF mitigation.
> 
> **Overview**
> Adds end-to-end MCP server support across **HTTP (Streamable), SSE, and desktop `stdio`** transports, including a refactored MCP settings UI (new add dialog, server cards with retry/authorize, platform transport filtering) and provider-side reconnect/backoff handling.
> 
> Introduces **MCP authentication**: bearer tokens and OAuth 2.1. OAuth now uses a dedicated callback route (`/mcp/oauth/callback`) with deep-link parsing, CSRF state checks, and a callback hook to exchange codes; credentials are stored in a new local-only `mcp_credentials` table using AES-GCM encryption derived from a per-device key.
> 
> Stops syncing `mcp_servers` via PowerSync (now local-only), updates tool wiring so chat sessions fetch enabled MCP clients via a provider getter, and namespaces MCP tools by sanitized server name to avoid collisions while surfacing a connected-server summary in the system prompt. Also hardens the backend `mcp-proxy` by following redirects under `safeFetch` to block redirect-based SSRF, and enables required Tauri plugins/capabilities (`http` always on, `shell` spawn for specific commands) plus broader localhost HTTP allowlists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19adf105c9f41d0404d786434da2e2d9ed4d0a6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src-tauri/capabilities/default.json
```diff
@@ -33,34 +33,48 @@
...
-        },
-        {
-          "url": "https://router.huggingface.co"
-        },
+          "url": "https://**"
+        }
+      ]
+    },
+    {
+      "identifier": "shell:allow-spawn",
```
cjroth, line 62: We need to think about security implications here. It probably makes sense, but obviously massively expands our attack surface.

src-tauri/src/commands.rs
```diff
@@ -129,17 +129,12 @@ const LIBSQL_ENABLED: bool = true;
...
-#[cfg(not(feature = "native_fetch"))]
-const NATIVE_FETCH_ENABLED: bool = false;
-
 /// Returns the set of capabilities supported by the current build.
 #[command]
 pub fn capabilities() -> Capabilities {
     Capabilities {
         libsql: LIBSQL_ENABLED,
-        native_fetch: NATIVE_FETCH_ENABLED,
+        native_fetch: true,
```
cjroth: if it's always enabled, we technically could remove this from the capabilities list and just assume that it's always available other than web

src-tauri/Cargo.toml
```diff
@@ -82,6 +82,7 @@ thunderbolt-libsql = { package = "thunderbolt-libsql", path = "./thunderbolt_lib
 thunderbolt-email = { package = "thunderbolt-email", path = "./thunderbolt_email", optional = true }
 tauri-plugin-deep-link = "2.3"
 tauri-plugin-haptics = "2.3"
+tauri-plugin-shell = "2"
```
cjroth: Should this be 2.3?

src/lib/mcp-auth/credential-store.test.ts
```diff
@@ -0,0 +1,135 @@
+import { getDb } from '@/db/database'
+import { mcpServersTable } from '@/db/tables'
+import { afterAll, beforeAll, beforeEach, describe, expect, it, mock } from 'bun:test'
+import { v7 as uuidv7 } from 'uuid'
+import { resetTestDatabase, setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
+import { createCredentialStore, resetKeyCache } from './credential-store'
+import type { McpCredential } from '@/types/mcp'
+
+// Mock @tauri-apps/plugin-fs so getDeviceId returns a stable test value
+mock.module('@tauri-apps/plugin-fs', () => ({
```
cjroth: Probably makes sense, but I hate mocking - would like to see if we can do dependency injection (but for plugins, I think mocking is usually reasonable)

src/lib/mcp-auth/credential-store.ts
cjroth, line 1: Let's sync with @raivieiraadriano92 and figure out if we need this given e2ee and also potential overlap with his work.

src/lib/mcp-auth/oauth-client-provider.ts
```diff
@@ -0,0 +1,136 @@
...
+  OAuthTokens,
+} from '@modelcontextprotocol/sdk/shared/auth.js'
+import type { OAuthClientProvider } from '@modelcontextprotocol/sdk/client/auth.js'
+import type { CredentialStore } from '@/types/mcp'
+
+/**
+ * Static client ID for Thunderbolt published as a Client ID Metadata Document (CIMD).
+ * Authorization servers that support CIMD will fetch this to verify the client.
+ */
+const thunderboltClientId = 'https://thunderbolt.io/.well-known/oauth-client/thunderbolt'
```
cjroth: I wonder if we'll want to use an env var for the domain here given that some people will deploy thunderbolt on their internal infra.

src/lib/mcp-transports/transport-factory.test.ts
```diff
@@ -0,0 +1,173 @@
+import { describe, it, expect, mock } from 'bun:test'
+import { createTransport } from './transport-factory'
+import { TauriStdioTransport } from './tauri-stdio-transport'
+import { validateStdioCommand as validateCommand, validateStdioArgs as validateArgs } from '@/lib/mcp-utils'
+import type { McpServerConfig, CredentialStore, McpCredential } from '@/types/mcp'
+
+// Mock Tauri plugins so tests can run outside a Tauri runtime
+mock.module('@tauri-apps/plugin-http', () => ({
```
cjroth, line 8: Same with these - probably makes sense to mock but if there was a way to do dependency injection, that would be nice.

cjroth (changes requested):
This is just a quick first-pass review - I'm going to do a deeper review when I have more brainspace later.

src-tauri/capabilities/default.json
ital0: well definitely. But that's only required for the stdio mcp servers on the desktop build.

src/lib/mcp-auth/credential-store.test.ts
ital0: that's it! I double-checked here another approach, but this would enter as one of the exceptions that CLAUDE.md mentions.

src/lib/mcp-transports/transport-factory.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // Remote servers on all platforms + cross-origin on web → route through backend proxy
+    const isSameOrigin = typeof window !== 'undefined' && url.origin === window.location.origin
+    if (options?.cloudUrl && !isSameOrigin) {
+      const { createProxiedFetch } = await import('./proxied-fetch')
+      return {
+        transport: new StreamableHTTPClientTransport(url, { ...opts, fetch: createProxiedFetch(options.cloudUrl) }),
+        authProvider,
+      }
+    }
```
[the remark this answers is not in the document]
cjroth: This is a real issue that we should fix. MCP should only be proxied in web

PR #517, opened 2026-03-25
THU-369: Add OIDC Support for Enterprise

darkbanjo:
> [!NOTE]
> **High Risk**
> High risk because it introduces a new authentication mode (OIDC) that changes sign-in/session behavior and adds startup-time migration logic that can affect backend boot reliability.
> 
> **Overview**
> Adds an **enterprise OIDC auth mode** (`AUTH_MODE=oidc` / `VITE_AUTH_MODE=oidc`) that redirects unauthenticated users through an external OIDC provider instead of the waitlist/email-OTP flow. Backend wires Better Auth’s `genericOAuth` provider (`/sign-in/oauth2`) with env-gated config/validation and new settings (`OIDC_*`, `BETTER_AUTH_URL`), and the frontend adds a lazy-loaded `/oidc-redirect` route plus cookie `credentials: 'include'` to preserve OAuth state.
> 
> Introduces **OIDC test coverage and tooling**: new backend OIDC integration test using `oauth2-mock-server`, new Playwright E2E suite (with mock OIDC global setup/teardown) and a GitHub Actions workflow that shards tests and uploads merged HTML reports/screenshots. Also ensures **PGLite** databases run Drizzle migrations on startup (`runMigrations()` in `backend/src/index.ts`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 59f37ca5fbefe8498ef9ba3d0cca6715000f674c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/app.tsx
```diff
@@ -48,12 +48,52 @@ import Loading from './loading'
...
+        }
+        return res.json()
+      })
+      .then((data: { url: string }) => {
+        window.location.href = data.url
+      })
+      .catch((err: unknown) => {
+        console.error('OIDC redirect failed:', err)
+      })
+  }, [cloudUrl.isLoading, cloudUrl.value])
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — added an AbortController with cleanup to prevent the duplicate request race condition in React 18 strict mode.

backend/src/auth/auth.ts
```diff
@@ -25,6 +27,30 @@ const trustedOrigins = parseTrustedOrigins(process.env.TRUSTED_ORIGINS)
...
+  return [
+    genericOAuth({
+      config: [
+        {
+          providerId: 'oidc',
+          discoveryUrl: `${settings.oidcIssuer}/.well-known/openid-configuration`,
+          clientId: settings.oidcClientId,
+          clientSecret: settings.oidcClientSecret,
+          scopes: ['openid', 'profile', 'email'],
+          redirectURI: `${process.env.BETTER_AUTH_URL || 'http://localhost:8000'}/v1/api/auth/oauth2/callback/oidc`,
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — replaced the undocumented process.env.BETTER_AUTH_URL with a proper betterAuthUrl setting in the settings schema, documented in .env.example.

backend/src/auth/auth.ts
```diff
@@ -25,6 +27,30 @@ const trustedOrigins = parseTrustedOrigins(process.env.TRUSTED_ORIGINS)
  * - Both paths (manual OTP entry, clicking link) use the same verification endpoint
  * - No separate email verification needed - signing in proves email ownership
  */
+const buildOidcPlugins = () => {
+  const settings = getSettings()
+
+  if (settings.authMode !== 'oidc' || !settings.oidcIssuer) {
+    return []
+  }
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — added startup-time validation that throws a clear error when AUTH_MODE=oidc but OIDC_CLIENT_ID or OIDC_CLIENT_SECRET are missing.

src/vite-env.d.ts
```diff
@@ -3,6 +3,7 @@
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
 interface ImportMetaEnv {
   readonly VITE_THUNDERBOLT_CLOUD_URL?: string
+  readonly VITE_AUTH_MODE?: 'consumer' | 'oidc'
```
cjroth: I'd name this "thunderbolt" instead of "consumer" just cause in theory it could be used for more than consumer

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
+import { mock } from 'bun:test'
+import * as authUtils from '@/auth/utils'
+import * as waitlistUtils from '@/waitlist/utils'
+import * as settingsModule from '@/config/settings'
+import type { Settings } from '@/config/settings'
+
+// Mock email-sending functions to avoid side effects
+mock.module('@/auth/utils', () => ({
```
cjroth, line 10: Is there any way to refactor our code so that we don't have to mock this? Either modularizing our code better or dependency injection

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  ...authUtils,
+  sendSignInEmail: mock(() => Promise.resolve()),
+}))
+
+mock.module('@/waitlist/utils', () => ({
```
cjroth, line 15: Same here

cjroth (changes requested):

cjroth (changes requested):
Also, Claude's `/review`:

```
 Overview

  Adds a new AUTH_MODE=oidc enterprise authentication mode that enables sign-in through any OIDC-compliant provider (Keycloak, Okta,
  Auth0, Entra ID). Backend uses Better Auth's genericOAuth plugin with PKCE and OIDC discovery. Frontend conditionally redirects
  unauthenticated users through the OIDC flow instead of showing the waitlist/login page.

  Files changed: 19 files, +787 / -11

  ---
  Issues

  1. OidcRedirect uses useEffect for navigation — CLAUDE.md says don't

  Per the project's useEffect discipline rules, navigation side effects should use <Navigate replace /> in JSX rather than useEffect.
  However, this case is a legitimate exception: the component needs to make an async API call to get the redirect URL before navigating,
   and it depends on an async cloudUrl value. The useEffect is warranted here, but worth a comment noting why.

  2. OIDC callback test doesn't actually test the callback flow (oidc-integration.test.ts:137-197)

  The test named "should create a user and session after OIDC callback" doesn't actually exercise the callback endpoint. It hits the
  mock OIDC server's /token endpoint directly and asserts the mock returns tokens. The comment on line 174 acknowledges this: "The
  callback will fail on state validation (since we didn't go through sign-in first)". This means the test verifies that
  oauth2-mock-server works, not that your callback flow works. Consider renaming it to reflect what it actually tests (e.g., "mock OIDC
  server handles token exchange") or moving it into the "Mock OIDC server capabilities" describe block where similar tests already live.

  3. jose dependency added but never imported

  jose (v6.1.2) is added to package.json but no source file imports it. If it was added for future use or experimentation, it should be
  removed from this PR to keep the dependency footprint clean.

  4. buildOidcPlugins() is called at createAuth() time — settings are read eagerly

  buildOidcPlugins() calls getSettings() every time createAuth() is invoked. If createAuth is called once at startup (which appears to
  be the case), this is fine. But if it were ever called in a hot path, the env parsing would repeat. Not a blocker — just worth
  confirming createAuth is only called once per process.

  5. Frontend credentials: 'include' in OIDC mode has CORS implications

  src/contexts/auth-context.tsx:29 — switching credentials from 'omit' to 'include' when isOidcMode() means the backend must respond
  with Access-Control-Allow-Credentials: true and cannot use Access-Control-Allow-Origin: *. The backend already has
  corsAllowCredentials: true as default, but if a deployment sets CORS_ORIGINS=* with AUTH_MODE=oidc, browsers will reject the
  preflight. Worth validating at startup (or at least documenting) that OIDC mode requires explicit CORS origins, not *.

  6. Hardcoded callbackURL in OidcRedirect (src/app.tsx:80)

  callbackURL: window.location.origin + '/'

  This works for the standard case but could be fragile if the app is served under a subpath. Minor, but window.location.origin alone
  (no trailing /) would be slightly more robust and let the backend append paths.

  7. No error UI in OidcRedirect

  If the OIDC redirect fails (e.g., backend is down, misconfigured), the user sees the <Loading /> spinner forever with only a
  console.error. Consider adding a user-visible error state or a retry mechanism.

  ---
  What Looks Good

  - Clean separation — buildOidcPlugins() returns [] in consumer mode, so existing auth is completely unaffected. The spread
  ...buildOidcPlugins() in the plugins array is elegant.
  - PKCE enabled — good security practice for the OAuth flow.
  - OIDC discovery — using .well-known/openid-configuration means any compliant provider works out of the box.
  - Validation on startup — buildOidcPlugins() throws early if OIDC is enabled but env vars are missing.
  - Good documentation — oidc-local-dev.md and the Keycloak realm JSON make local testing straightforward.
  - Test infrastructure — oauth2-mock-server is a solid choice; the sign-in endpoint test (should return a redirect URL pointing to the
  OIDC provider) actually validates the integration.
  - Env var naming — AUTH_MODE, OIDC_* prefix — clear and conventional.

  ---
  Summary

  The architecture is sound and well-integrated. The main actionable items are:
  1. Remove jose if unused (or add the code that uses it)
  2. Rename/reorganize the misleading callback test (it tests the mock, not the callback)
  3. Add CORS + OIDC validation or docs to prevent credentials: 'include' + origin: * misconfigurations
  4. Consider an error state in OidcRedirect instead of infinite spinner on failure
```

src/app.tsx
```diff
@@ -48,12 +48,58 @@ import Loading from './loading'
...
 import { useSettings } from './hooks/use-settings'
+import ky from 'ky'
+import { isOidcMode } from './lib/auth-mode'
 import { isPrPreview, isTauri } from './lib/platform'
 import { getPowerSyncInstance } from './db/powersync'
 import { type ComponentProps, useEffect } from 'react'
 
 const queryClient = new QueryClient()
 
+/**
```
cjroth: I think we should look into compartmentalizing this OIDC-specific code and lazily-import it only for enterprise deployments - I don't think non-enterprise deployments should be bloated with the extra code needed here (thinking in terms of just attack surface, bundle size, etc)

cjroth (changes requested):

backend/package.json
```diff
@@ -35,6 +35,7 @@
     "drizzle-orm": "^0.44.7",
     "elysia": "^1.4.7",
     "exa-js": "^1.0.11",
+    "jose": "6.1.2",
```
cjroth, line 38: I think we can remove this?

src/app.tsx
```diff
@@ -48,12 +48,58 @@ import Loading from './loading'
...
+ * In OIDC mode, redirects unauthenticated users to the backend's OIDC sign-in endpoint,
+ * which in turn redirects to the OIDC provider. The user never sees a login page on our app.
+ */
+const OidcRedirect = () => {
+  const { cloudUrl } = useSettings({ cloud_url: String })
+
+  useEffect(() => {
```
cjroth: Let's add a comment here explaining why we need useEffect

cjroth (changes requested):

backend/docs/oidc-local-dev.md
```diff
@@ -0,0 +1,159 @@
...
+Any OIDC-compliant provider works — the implementation uses standard OIDC discovery (`.well-known/openid-configuration`).
+
+## Quick start (Keycloak example)
+
+### 1. Start Keycloak with pre-configured realm
+
+The `docs/mozilla-realm.json` file contains a ready-to-go realm with a client and test users. Mount it on startup so there's zero manual setup:
+
+```sh
+docker run -d \
```
raivieiraadriano92, line 25: Maybe we could refactor our current docker setup to include a container for keycloak and other future ones. Currently we use docker only for PowerSync but I think we could extend it.

ital0: @raivieiraadriano92 what's the idea? reuse the same container we are using for PowerSync? If so, docker recommends to use a single container per "feature"/use.

backend/package.json
ital0: same, it's not being used

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+        const body = await res.json()
+        expect(body.url).toContain(oidcIssuerUrl)
+        expect(body.url).toContain('response_type=code')
+        expect(body.url).toContain('client_id=thunderbolt-app')
+      } finally {
+        globalThis.fetch = mockedFetch
+      }
+    })
+
+    it('should create a user and session after OIDC callback', async () => {
```
ital0: I think this test doesn't bring value to us.
It's testing the mock server instead of the application. Or, in other words, it's trying to mimic the implementation and reproduce here. And it should validate the implementatio itself.

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+        expect(tokenData.token_type).toBe('Bearer')
+        expect(tokenData.id_token).toBeDefined()
+      } finally {
+        globalThis.fetch = mockedFetch
+      }
+    })
+  })
+
+  describe('Mock OIDC server capabilities', () => {
+    it('handles token requests with custom claims', async () => {
```
ital0: same, it's validating the mock server, not the implementation itself.

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+          payload.email = 'test@mozilla.org'
+        },
+      })
+
+      expect(typeof token).toBe('string')
+      // JWT has 3 dot-separated parts
+      expect(token.split('.')).toHaveLength(3)
+    })
+  })
+})
```
ital0, line 141: I added two comments on two tests mentioning that they bring value to us since they're not validating the implementation itself. It's creating a mock server and validating this.

ital0: From I could noticed, only the first test here really brings value.

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  ...waitlistUtils,
+  sendWaitlistNotReadyEmail: mock(() => Promise.resolve()),
+  sendWaitlistJoinedEmail: mock(() => Promise.resolve()),
+  sendWaitlistReminderEmail: mock(() => Promise.resolve()),
+}))
+
+import { OAuth2Server } from 'oauth2-mock-server'
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, spyOn } from 'bun:test'
+import { Elysia } from 'elysia'
+import { user, session, account } from '@/db/auth-schema'
```
ital0: looks like these variables are not being used

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+import { createTestDb } from '@/test-utils/db'
+import { eq } from 'drizzle-orm'
```
ital0: this one is also not being used

backend/src/auth/oidc-integration.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  afterAll(async () => {
+    await oidcServer.stop()
+  })
+
+  describe('OIDC sign-in endpoint', () => {
+    let db: Awaited<ReturnType<typeof createTestDb>>['db']
+    let cleanup: () => Promise<void>
+    let getSettingsSpy: ReturnType<typeof spyOn>
+
+    beforeEach(async () => {
```
ital0: I believe this test case is not using database, so doesn't worth to keep creating and recreating

ital0 (changes requested):
Since most of the detail had been pointed out by bots or Chris and Rai, I focused in the unit tests, which is a point we are hunting to make it better.

I pointed out some details that can be improved.

src/vite-env.d.ts
darkbanjo: ⚡ Good call — renamed to `'thunderbolt'` in the latest push.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Agreed DI would be better. The challenge is `createAuth` imports `sendSignInEmail` and waitlist utils at module level, so we'd need to refactor its signature to accept those as parameters. Added a comment explaining why the mocks are currently necessary and left this as a follow-up refactor.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Same reasoning as above — added explanatory comment.

src/app.tsx
darkbanjo: ⚡ Good call — extracted `OidcRedirect` into its own file and lazy-loaded it with `React.lazy` + `Suspense`. Non-enterprise deployments won't load any OIDC code.

backend/package.json
darkbanjo: ⚡ Investigated this — `jose` isn't imported at runtime but is required for TypeScript type resolution. `better-auth` exposes types that reference `jose`, and without the pinned top-level dep, TS fails with `TS2742: The inferred type cannot be named without a reference to better-auth/node_modules/jose`. Keeping it pinned at `6.1.2` to match better-auth's transitive dep so bun deduplicates them.

src/app.tsx
darkbanjo: ⚡ Done — added comment explaining the useEffect is for an external navigation side-effect that depends on an async setting value.

backend/docs/oidc-local-dev.md
darkbanjo: ⚡ Good idea for a follow-up. We could extend docker-compose to include a Keycloak container. Leaving this for a separate PR since the current setup works as documented.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Agreed — removed this test. Only keeping the sign-in redirect test which actually exercises the integration.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Agreed — removed the entire mock server capabilities describe block.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Good feedback — trimmed down to just the redirect test which is the only one exercising the actual integration.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Removed — cleaned up all unused imports.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Removed.

backend/src/auth/oidc-integration.test.ts
darkbanjo: ⚡ Good point — moved db creation to `beforeAll`/`afterAll` so it's only created once for the suite.

PR #518, opened 2026-03-25
THU-367: Add e2e support w/ parallelization

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Introduces new CI infrastructure and e2e auth flow mocking, plus changes backend startup to run PGlite migrations, which could affect local/CI boot behavior if migration paths or envs are misconfigured.
> 
> **Overview**
> Adds Playwright e2e coverage for the OIDC login/session flow using a local `oauth2-mock-server`, with a new `playwright.config.ts` that boots both Vite and the backend with OIDC-specific env and disables onboarding via `VITE_SKIP_ONBOARDING`.
> 
> Introduces a new GitHub Actions workflow to run e2e tests in **2 shards**, upload blob reports/screenshots, and merge shard results into an HTML report artifact.
> 
> Backend startup now calls `runMigrations()` to automatically apply Drizzle migrations when using PGLite (no-op for Postgres), and repo tooling is updated with new e2e scripts/deps plus `.gitignore` entries for Playwright outputs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fb2ea09195c8baa1d70676cc7e2fd617f3734a83. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,11 +23,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '20'
-
       - name: Install Bun
         uses: oven-sh/setup-bun@v2
         with:
@@ -77,11 +72,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '20'
 
       - name: Install Bun
         uses: oven-sh/setup-bun@v2
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard: [1/2, 2/2]
+        shard: [1/6, 2/6, 3/6, 4/6, 5/6, 6/6]
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard: [1/6, 2/6, 3/6, 4/6, 5/6, 6/6]
+        shard: [1/2, 2/2]
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,109 @@
+name: E2E Tests
+
+on:
+  push:
+    branches:
+      - main
+
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+    branches: ['**']
+
+concurrency:
+  group: e2e-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1/2, 2/2]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - name: Cache Bun dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.bun/install/cache
+            node_modules
+          key: bun-${{ hashFiles('bun.lock') }}
+          restore-keys: |
+            bun-
+
+      - name: Install dependencies
+        run: bun install
+
+      - name: Install backend dependencies
+        run: cd backend && bun install
+
+      - name: Install Playwright browsers
+        run: bunx playwright install --with-deps chromium
+
+      - name: Run e2e tests (shard ${{ matrix.shard }})
+        run: bunx playwright test --shard=${{ matrix.shard }}
+
+      - name: Upload blob report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: blob-report-${{ strategy.job-index }}
+          path: blob-report/
+          retention-days: 7
+
+      - name: Upload test screenshots
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-screenshots-${{ strategy.job-index }}
+          path: test-results/
+          retention-days: 7
+
+  e2e-report:
+    if: always() && needs.e2e.result != 'skipped'
+    needs: e2e
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - name: Install dependencies
+        run: bun install
+
+      - name: Download blob reports
+        uses: actions/download-artifact@v4
+        with:
+          pattern: blob-report-*
+          path: all-blob-reports
+          merge-multiple: true
+
+      - name: Merge reports
+        run: bunx playwright merge-reports --reporter html ./all-blob-reports
+
+      - name: Upload HTML report
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-report
+          path: playwright-report/
+          retention-days: 14
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -4,6 +4,7 @@ import { createGoogleAuthRoutes } from '@/auth/google'
 import { createMicrosoftAuthRoutes } from '@/auth/microsoft'
 import { createLoggerMiddleware, createStandaloneLogger } from '@/config/logger'
 import { getCorsOriginsList, getSettings } from '@/config/settings'
+import { runMigrations } from '@/db/client'
 import { createInferenceRoutes } from '@/inference/routes'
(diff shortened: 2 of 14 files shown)


backend/src/index.ts
```diff
@@ -108,6 +108,10 @@ const startServer = async () => {
   )
 
   try {
+    // Run PGLite migrations before creating the app (no-op for Postgres)
+    const { runMigrations } = await import('@/db/client')
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — done in the latest push. Moved to a top-level import.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,109 @@
...
+  e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1/2, 2/2]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
```
cjroth: Do we need both Node and Bun?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,109 @@
...
+concurrency:
+  group: e2e-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
```
cjroth, line 21: Oh didn't know GitHub Actions had this built in. I think we could parallelize even more.

e2e/global-setup.ts
```diff
@@ -0,0 +1,37 @@
+import { OAuth2Server } from 'oauth2-mock-server'
+
+const MOCK_OIDC_PORT = 9876
```
cjroth: camel case

e2e/helpers.ts
```diff
@@ -0,0 +1,82 @@
+import { expect, type Page } from '@playwright/test'
+
+/**
+ * Dismiss the onboarding wizard if it appears.
```
cjroth: I added an env var to disable onboarding in the ACP branch - can you grab that and do that instead?

playwright.config.ts
```diff
@@ -0,0 +1,60 @@
+import { defineConfig, devices } from '@playwright/test'
+
+const isCI = !!process.env.CI
+const MOCK_OIDC_PORT = 9876
```
cjroth: camel case

cjroth (changes requested):
Looks great overall.

e2e/global-setup.ts
cjroth: Can we set this in an env var once instead of setting it here and then again in playwright config?

cjroth (changes requested):
One thing I want to flag is that we should never use `waitForTimeout` - it will slow down tests and could also timeout randomly. Instead we should wait for specific conditions. (I want to update all the e2e tests in the ACP branch this way as well). If there's any way to disable waitForTimeout completely, let's do it.

On that note, I'd like to also update our Agents.md file to tell it to always start frontend features with e2e tests, to never use waitForTimeout, and then to get the tests passing until it's done. In my experience when given this prompt, Claude gets things right in few few turns.

e2e/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+    const continueButton = page.getByRole('button', { name: 'Continue' })
+    await expect(continueButton).toBeVisible({ timeout: 4000 })
+
+    // Step 1: agree to privacy and continue
+    const checkbox = page.locator('[role="checkbox"]').first()
+    if (await checkbox.isVisible()) {
+      await checkbox.click()
+    }
+    await continueButton.click()
+    await page.waitForTimeout(300)
```
raivieiraadriano92: Claude Code reivew:

3. waitForTimeout usage in helpers — e2e/helpers.ts:18,28,37,58

Playwright docs explicitly discourage waitForTimeout as it creates flaky tests. There are 5 instances:
- Lines 18, 28, 37: 300ms/500ms waits between onboarding steps
- Line 58: 3000ms wait after page.goto('/')

Recommendation: Replace with waitForSelector, waitForURL, or other condition-based waits. For the onboarding steps, wait for the next step's element to appear. For the initial load, wait for a specific element or network idle.

----

if this makes sense I think we should refactor the other waitForTimeout too.

ital0: I know team has pointed this, but worth to mention again and maybe provide a way to replace.
The `waitForTimeOut` is being used in some tests and I think you can replace them by condition-based waits:

- `waitForTimeout(300)` between onboarding steps -> `waitForSelector` or `waitForResponse` on the next expected element
- `waitForTimeout(3000)` after `goto('/')` -> wait for a specific network response or DOM element
- `waitForTimeout(1000)` in spec -> replace with `expect(...).toBeVisible()` for the expected state

playwright.config.ts
```diff
@@ -0,0 +1,60 @@
...
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+  ],
+  webServer: [
+    {
+      command: `bun run dev -- --port ${E2E_VITE_PORT}`,
+      url: `http://localhost:${E2E_VITE_PORT}`,
+      reuseExistingServer: false,
```
ital0: This one I'll just gonna paste the analysis from Claude since I think it makes sense, but double-check:

Both webServer entries have `reuseExistingServer: false`. This means running `bun e2e` locally while a dev server is already running on the same port will fail, forcing the developer to stop their dev server first. The standard Playwright pattern is `reuseExistingServer: !isCI` -- reuse locally, spawn fresh in CI.

**Suggestion:** Change both to `reuseExistingServer: !isCI`.

e2e/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // Step 5: finish onboarding
+    const startButton = page.getByRole('button', { name: 'Start Using Thunderbolt' })
+    try {
+      await expect(startButton).toBeVisible({ timeout: 2000 })
+      await startButton.click()
+      await page.waitForTimeout(500)
+    } catch {
+      // May have already closed
```
ital0: I see this catch (and one below in line 42) empty. 
What's the purpose? To avoid a general failure?

package.json
```diff
@@ -122,6 +124,7 @@
     "@eslint/js": "^9.33.0",
     "@happy-dom/global-registrator": "^20.0.8",
     "@libsql/client": "^0.15.12",
+    "@playwright/test": "^1.58.2",
```
ital0, line 127: make sure to keep this and `playwright` in same version (1.55.x, maybe)

playwright.config.ts
ital0: I know the goal is to have that in CI... But I thought it would make sense to mention.

ital0 (changes requested):
Great work! Hopefully this would help us to catch bugs before they get deployed.
Just pointed out minor details.

.github/workflows/e2e.yml
darkbanjo: ⚡ Good call — removed the explicit Node.js setup. GitHub Actions ubuntu-latest comes with Node pre-installed, and Playwright uses that under the hood via bunx. No separate setup-node step needed.

.github/workflows/e2e.yml
darkbanjo: ⚡ Agreed — the built-in matrix sharding is nice. We could add more shards as the test suite grows. For now 2 shards keeps it fast enough.

e2e/global-setup.ts
darkbanjo: ⚡ Done — renamed to camelCase.

e2e/helpers.ts
darkbanjo: ⚡ Good call — grabbed the VITE_SKIP_ONBOARDING env var from ACP. It's now set in playwright.config.ts and the onboarding dialog checks for it. Removed the dismissOnboarding helper entirely.

playwright.config.ts
darkbanjo: ⚡ Done — renamed to camelCase.

e2e/global-setup.ts
darkbanjo: ⚡ Done — the port is now read from MOCK_OIDC_PORT env var with a default of 9876 in both global-setup.ts and playwright.config.ts. Single source of truth via env var.

e2e/helpers.ts
darkbanjo: ⚡ Great point — replaced all waitForTimeout calls with condition-based waits (waitForSelector/toBeVisible). The onboarding waits were removed entirely since we now skip onboarding via env var.

darkbanjo: ⚡ Agreed — all waitForTimeout calls have been replaced. The onboarding ones were removed entirely (onboarding is now skipped via VITE_SKIP_ONBOARDING env var), and the spec-level ones now use expect(...).toBeVisible() assertions.

playwright.config.ts
darkbanjo: ⚡ Good suggestion — done. Both webServer entries now use reuseExistingServer: !isCI so local dev servers are reused.

e2e/helpers.ts
darkbanjo: ⚡ Those empty catches were swallowing failures from optional onboarding steps (e.g. 'Start Using Thunderbolt' button might not appear if already dismissed). The entire dismissOnboarding function has been removed since we now skip onboarding via VITE_SKIP_ONBOARDING env var, so this is no longer an issue.

package.json
darkbanjo: ⚡ Done — both @playwright/test and playwright are now aligned at ^1.58.2.

.github/workflows/e2e.yml
cjroth: Let's use more shards

cjroth (changes requested):

PR #520, opened 2026-03-29
Browser SQLite Explorer

raivieiraadriano92:
# Landscape Research

> Research date: 2026-03-29

### NPM Packages (embeddable)

| Solution | Type | Fits our use case? |
|----------|------|-------------------|
| **Drizzle Studio Embeddable** (`@drizzle-team/studio`) | Web component, 210K downloads | No — commercial license, designed for server-side DBs, not in-browser SQLite |
| **PGlite REPL** (`@electric-sql/pglite-repl`) | React component | No — Postgres only. But architecturally the closest pattern to what we'd want |
| **Expo Drizzle Studio Plugin** | Expo DevTools plugin | No — React Native/Expo only |
| **LiveStore Devtools** | npm package | No — tightly coupled to LiveStore framework |

### Browser Extensions

| Extension | What it does | Limitation |
|-----------|-------------|------------|
| **OPFS Explorer** (Chrome) | Browse OPFS file tree | File browser only — can't query data |
| **OPFS Viewer** (Chrome) | Same idea | Same limitation |
| **SQLite Viewer** (Chrome) | Opens SQLite files from OPFS | Mixed reviews, can't connect to a live in-memory DB |
| **SQLite Explorer** (Chrome) | DevTools panel for sql.js | sql.js specific, unclear wa-sqlite support |

### Standalone Web Apps

| Project | Stars | Tech | Reusable? |
|---------|-------|------|-----------|
| **sqlime.org** | 1,100 | Vanilla JS, sql.js | No — standalone app |
| **browser-sqlite-editor** | New | React, wa-sqlite, CodeMirror 6, TanStack Table, React Flow (ERD!) | No — standalone PWA, not published as package |
| **sqlite-online** | 139 | React, sql.js | No — standalone app |
| **sqliteviewer.app** | — | WASM-based, very polished | No — standalone app |
| **PowerSync Diagnostics App** | — | Standalone web app | No — separate auth, not embeddable |

### Key Finding

**There is a clear gap**: no open-source, embeddable React component exists for exploring in-browser SQLite databases (wa-sqlite, sql.js, PowerSync). Every existing solution is either:

- A standalone app (not embeddable)
- Commercial (Drizzle Studio)
- Wrong database engine (PGlite = Postgres)
- A browser extension that can't access the live JS database instance

## npm Package vs Browser Extension

**npm package is the clear winner.** Here's why:

1. **Live database access required** — a browser extension communicates via message passing and cannot access WASM memory or JS objects directly. The adapter pattern requires a JS reference to the DB.

2. **Established pattern** — React Query DevTools, TanStack Router DevTools, and LiveStore Devtools all ship as npm packages with a React component. This is the standard React ecosystem approach for dev tooling.

3. **Distribution model**:
   - Ship as `@thunderbolt/sqlite-explorer` (or similar)
   - Provide two modes: **floating** (small toggle button that expands to overlay) and **embeddable** (render as a regular component)
   - Accept a `SqliteExplorerAdapter` as prop (existing pattern)
   - Auto-exclude from production via `process.env.NODE_ENV`
   - Tree-shakeable

4. **Adapter pattern already solves the hard problem** — `SqliteExplorerAdapter` is DB-agnostic. Users can implement it for wa-sqlite, sql.js, PowerSync, SQLocal, or any other browser SQLite. Ship a few built-in adapters (`createDrizzleExplorerAdapter`, `createSqlJsAdapter`, etc.)

5. **The browser-sqlite-editor project** (React + wa-sqlite + CodeMirror 6) validates the tech stack but is a standalone PWA — extracting it as a component would be significant work. We're already building the component-first version.

## Market Opportunity

The local-first / offline-first ecosystem (PowerSync, wa-sqlite, sql.js, SQLocal, Electric SQL) is growing fast and has zero embeddable debugging tools. Publishing this as an npm package following the TanStack DevTools pattern would fill a genuine unserved niche.

## Sources

- [Drizzle Studio npm package](https://github.com/drizzle-team/drizzle-studio-npm)
- [@drizzle-team/studio on npm](https://www.npmjs.com/package/@drizzle-team/studio)
- [Sqlime — Online SQLite Playground](https://github.com/nalgeon/sqlime)
- [sqlite-online](https://github.com/vwh/sqlite-online)
- [browser-sqlite-editor](https://github.com/felix-huber/browser-sqlite-editor)
- [OPFS Explorer Chrome Extension](https://chromewebstore.google.com/detail/opfs-explorer/acndjpgkpaclldomagafnognkcgjignd)
- [SQLite Explorer Chrome Extension](https://chromewebstore.google.com/detail/sqlite-explorer/lefjpjkijgpdfmapohighokjlhfndggl)
- [SQLite Viewer Chrome Extension](https://chromewebstore.google.com/detail/sqlite-viewer/golagekponhmgfoofmlepfobdmhpajia)
- [OPFS Viewer Chrome Extension](https://chromewebstore.google.com/detail/opfs-viewer/mfcfdmpjfomalelcneiglccaokdmkfnl)
- [PowerSync Diagnostics App](https://diagnostics-app.powersync.com/)
- [TanStack React Query DevTools](https://tanstack.com/query/v4/docs/react/devtools)
- [@tanstack/react-query-devtools on npm](https://www.npmjs.com/package/@tanstack/react-query-devtools)
- [PGlite REPL](https://www.npmjs.com/package/@electric-sql/pglite-repl)
- [LiveStore](https://livestore.dev/)
- [Expo SQLite DevTools](https://docs.expo.dev/versions/latest/sdk/sqlite/)
- [expo-drizzle-studio-plugin](https://www.npmjs.com/package/expo-drizzle-studio-plugin)
- [SQLite Viewer Web App](https://sqliteviewer.app/)
- [sql-viewer](https://github.com/amoscardino/sql-viewer)


cjroth (changes requested):
I love this - I think we should either find a way to make this only bundle in dev builds or maybe even put it into a browser extension (btw is there a major benefit to using this vs an existing sqlite browser extension?)

If we are going to go forward with it, I'd want to add tests.

Here's Claude's code review:

---

Overview

  A new devtools page that lets developers explore the in-browser SQLite database. It features:
  - A collapsible sidebar listing tables, views, indexes, and triggers
  - A CodeMirror-based SQL editor with Cmd+Enter execution
  - A paginated data table with resizable columns and click-to-copy
  - An adapter pattern (SqliteExplorerAdapter) making the component DB-agnostic
  - Read-only SQL definition view for indexes/triggers

  The architecture is solid — clean adapter pattern, proper use of useReducer for complex state, and
  state/logic extracted into a custom hook per project conventions. Well-researched PR description too.

  ---
  Issues

  1. SQL Injection via unescaped identifiers in use-db-explorer-state.ts

  The adapter has a proper escapeId() helper, but the state hook constructs SQL with raw template
  literals:

  // use-db-explorer-state.ts:152
  adapter.execute(`SELECT * FROM "${name}" LIMIT ${defaultPageSize} OFFSET 0`)

  // use-db-explorer-state.ts:206
  adapter.execute(`SELECT * FROM "${state.selectedObject}" LIMIT ${state.pageSize} OFFSET ${offset}`)

  A table named foo"bar would break the query. Either expose escapeId from the adapter or move these
  queries inside the adapter itself (e.g., adapter.getPage(name, limit, offset)).

  2. Duplicated useIsDarkMode hook

  Identical function in both sql-definition.tsx:13-18 and sql-editor.tsx:18-23. Extract to a shared
  file.

  Also, when theme === 'system', this reads window.matchMedia at render time but doesn't subscribe to
  changes. Per project conventions, useSyncExternalStore would be the right fit here, since this is
  exactly "subscribing to a browser API."

  3. Sequential row count loading

  object-list.tsx:29-39 fetches row counts in a serial for loop:

  for (const obj of countable) {
    counts.set(obj.name, await adapter.getRowCount(obj.name))
  }

  With many tables, this is noticeably slow. Promise.all (or batched parallel) would be faster.

  4. columns stored in state but never read

  state.columns is populated in SELECT_OBJECT (use-db-explorer-state.ts:77) but no component ever reads
   it. The DataTable uses result.columns from the query result instead. Dead state — remove it or use
  it.

  5. resetWidths defined but unused

  use-column-resize.ts:37 exports resetWidths, which is never called. Remove it.

  6. Custom SQL runs without pagination

  runQuery (use-db-explorer-state.ts:175-192) sets totalRows: result.rows.length, meaning all rows are
  fetched into memory. For a SELECT * FROM on a large table, this could be expensive. Consider wrapping
   custom queries in a COUNT(*) subquery or at least adding a default LIMIT.

  7. No production guard

  The PR description mentions "auto-exclude from production via process.env.NODE_ENV", but the route in
   app.tsx:115 is unconditional. This will ship to production. Should be gated like the tasks route:

  {isDev && <Route path="db-explorer" element={<DbExplorerPage />} />}

  8. No tests

  No test files included. The drizzle-adapter.ts (SQL parsing, row normalization, column extraction)
  and use-db-explorer-state.ts (reducer logic) are both very testable and have meaningful edge cases
  worth covering.

  ---
  Minor / Style

  - data-table.tsx:66 — (row as unknown[]): The QueryResult type already declares rows: unknown[][], so
   this cast shouldn't be needed. If it is, the type might not be flowing right.
  - Double border on sidebar: db-explorer.tsx:29 has border-r on the sidebar container, but the
  ResizableHandle already provides a visual separator.
  - setPageSize async pattern (use-db-explorer-state.ts:221-243): The inner run() function is defined
  and immediately called. An async IIFE or extracting to a separate function would be cleaner.
  - Hardcoded Cmd+Enter in sql-editor.tsx:108: On Windows/Linux this would be Ctrl+Enter. The keymap
  binding uses Mod-Enter correctly, but the displayed hint says Cmd+Enter.

  ---
  What's good

  - The adapter interface is well-designed and genuinely DB-agnostic
  - useReducer + extracted state hook follows project conventions
  - CodeMirror integration is clean — callback refs avoid stale closures, theme changes handled
  properly
  - Column resize hook is a nice touch
  - The temp view approach for extracting column names from arbitrary queries is clever
  - PR description with landscape research is thorough

  ---
  Verdict

  Good foundation for a dev tool. The main things to address before merging: production guard on the
  route, extract/fix useIsDarkMode, escape identifiers consistently, and remove dead code (columns
  state, resetWidths). Tests would be nice but less critical for an experimental devtool behind a
  feature gate.

PR #532, opened 2026-04-01
THU-113: Rate Limiting Middleware

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Adds new middleware and persistence to enforce request limits on authenticated inference/pro endpoints, which can cause unexpected 429s or header/CORS issues if misconfigured. Also changes client IP extraction behavior behind proxies, affecting both logging and rate-limit keying.
> 
> **Overview**
> Adds **DB-backed rate limiting** using `rate-limiter-flexible` with a new `rate_limits` table/migration and middleware (`createInferenceRateLimit`, `createProRateLimit`) applied to authenticated `/chat` and `/pro` route groups.
> 
> Introduces new settings/env vars `RATE_LIMIT_ENABLED` (default on) and `TRUSTED_PROXY` (`cloudflare`/`akamai`/empty), updates CORS exposed headers to include rate-limit headers, and updates HTTP logging to use the new proxy-aware `extractClientIp` helper.
> 
> Improves streaming inference responses to **preserve rate-limit headers** when returning a raw `Response`, and updates the frontend to detect 429/rate-limit errors, avoid auto-retry, and show a dedicated rate-limit message (with new unit tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 718bb71e8b15885ecfe3f6372d203b666bf8664f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/src/index.ts
```diff
@@ -72,6 +80,8 @@ export const createApp = async (deps?: AppDeps) => {
       .use(createLoggerMiddleware(settings))
       .use(createHttpLoggingMiddleware())
       .use(createErrorHandlingMiddleware())
+      // Global rate limit — applies to all routes (health exempt)
+      .use(createStandardRateLimit(database, rateLimitSettings))
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — done. Auth rate limit (10/15min) is now scoped to the betterAuthPlugin via a wrapping Elysia instance, and inference rate limit (20/min) is applied the same way around createInferenceRoutes. The auth tier is further narrowed to only count credential-based paths (sign-in, sign-up, password reset) so OIDC callbacks and session checks aren't affected.

backend/src/utils/request.ts
```diff
@@ -34,6 +34,23 @@ export const defaultResponseDenylist = [
...
+ */
+export const extractClientIp = (headers: Headers, fallback = 'unknown'): string => {
+  const forwarded = headers.get('forwarded')
+  if (forwarded) {
+    const match = forwarded.match(/for="?([^;"]+)/i)
+    if (match?.[1]) return match[1]
+  }
+
+  const xff = headers.get('x-forwarded-for')
+  if (xff) return xff.split(',')[0].trim()
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed. Now using xff.split(',').at(-1)\!.trim() to take the rightmost (infrastructure-appended) IP instead of the attacker-controlled leftmost one.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,119 @@
...
+export const createInferenceRateLimit = (database: typeof DbType, settings: RateLimitSettings) => {
+  if (!settings.enabled) return rateLimit({ max: Number.MAX_SAFE_INTEGER, duration: 1 })
+  return createTieredRateLimit(database, settings.inference)
+}
+
+/** Create rate limit middleware for auth routes. */
+export const createAuthRateLimit = (database: typeof DbType, settings: RateLimitSettings) => {
+  if (!settings.enabled) return rateLimit({ max: Number.MAX_SAFE_INTEGER, duration: 1 })
+  return createTieredRateLimit(database, settings.auth)
+}
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Both issues resolved. createInferenceRateLimit and createAuthRateLimit are now wired into index.ts via scoped Elysia instances. The counter interference is fixed by prefixing rate limit keys with the tier name (e.g. 'auth:1.2.3.4', 'inference:1.2.3.4'), so each tier tracks independently in the same rate_limits table.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,149 @@
...
+import { extractClientIp } from '@/utils/request'
+import { rateLimit, type Context as RateLimitContext, type Options } from 'elysia-rate-limit'
+import { sql } from 'drizzle-orm'
+
+/**
+ * Postgres-backed rate limit context for elysia-rate-limit.
+ * Uses an atomic UPSERT query with a sliding window per IP.
+ * Keys are prefixed with a tier name to prevent counter interference across tiers.
+ */
+export class PostgresRateLimitContext implements RateLimitContext {
```
raivieiraadriano92: Would it be possible to replace the raw sql queries with drizzle queries in this class?

backend/src/config/settings.ts
```diff
@@ -106,6 +112,10 @@ const parseSettings = (): Settings => {
       process.env.CORS_ALLOW_HEADERS ||
       'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name',
     corsExposeHeaders: process.env.CORS_EXPOSE_HEADERS || 'mcp-session-id,set-auth-token',
+    rateLimitEnabled: process.env.RATE_LIMIT_ENABLED !== 'false',
```
raivieiraadriano92, line 123: I think we should update the .env.example with the new env vars

backend/src/middleware/rate-limit.ts
raivieiraadriano92: we could create a `/dal/ip-rate-limite.ts` file with the drizzle queries

backend/src/config/settings.ts
darkbanjo: good call, that was an oversight on my part

backend/src/middleware/rate-limit.ts
darkbanjo: ⚡ Good suggestion — done. Created `backend/src/dal/ip-rate-limit.ts` with Drizzle queries for all rate limit operations (increment, decrement, delete by IP, delete all). The middleware class now calls through the DAL instead of using raw SQL.

backend/src/config/settings.ts
darkbanjo: ⚡ Good call — added all rate limit env vars (RATE_LIMIT_ENABLED, RATE_LIMIT_INFERENCE_MAX, RATE_LIMIT_AUTH_MAX, RATE_LIMIT_STANDARD_MAX) to .env.example with descriptive comments.

backend/src/utils/request.ts
```diff
@@ -34,6 +34,23 @@ export const defaultResponseDenylist = [
   'cross-origin-opener-policy',
 ]
 
+/**
+ * Extract client IP address from request headers.
+ * Checks standard proxy headers in order of preference.
+ */
+export const extractClientIp = (headers: Headers, fallback = 'unknown'): string => {
+  const forwarded = headers.get('forwarded')
```
ital0: well this one I didn't grab myself, but I think it worth to take a look.

what do you think about just drop the `Forwarded` check since we're using XFF/CF headers in our infra anyway?

Per Claude's words:

the `Forwarded` header is checked *before* XFF and it grabs the first `for=` value, which is still attacker-controlled. So someone can just set `Forwarded: for=whatever` and skip rate limiting entirely.

backend/src/db/rate-limit-schema.ts
```diff
@@ -0,0 +1,7 @@
+import { integer, pgTable, text, timestamp } from 'drizzle-orm/pg-core'
+
+export const rateLimits = pgTable('rate_limits', {
```
ital0: that's a great approach.
But I noticed that this table never gets cleaned up. If an IP hits us once and never comes back, that row lives forever.
We can have a pg_cron job just to cleanup this table...

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,127 @@
...
+  private duration = 60_000
+  private database: typeof DbType
+  private tier: string
+
+  constructor(database: typeof DbType, tier = 'standard') {
```
ital0: what about typing `tier` here? Do we already have the possible tiers?

backend/src/utils/request.ts
```diff
@@ -34,6 +34,23 @@ export const defaultResponseDenylist = [
   'cross-origin-opener-policy',
 ]
 
+/**
+ * Extract client IP address from request headers.
+ * Checks standard proxy headers in order of preference.
+ */
+export const extractClientIp = (headers: Headers, fallback = 'unknown'): string => {
```
ital0: do you think we can have some unit tests for this method?

ital0 (changes requested):

backend/src/db/rate-limit-schema.ts
ital0: well I think it worth a conversation to understand if that really makes sense. Let's get everyone's opinion.

backend/src/utils/request.ts
darkbanjo: yeah that is a good call, it makes sense, i'll drop it

backend/src/utils/request.ts
darkbanjo: added unit tests!

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const exemptPaths = new Set(['/v1/health', '/v1/posthog/config', '/v1/posthog/events'])
+
+const exemptPrefixes = ['/v1/api/auth/get-session']
+
+/**
+ * Auth paths that are abuse-prone and should be rate-limited.
+ * All other auth paths (session checks, OIDC callbacks, etc.) are exempt.
+ */
+const rateLimitedAuthPrefixes = [
```
cjroth: We should actually rate limit all endpoints. For the ones where we're adding authentication like search and inference, we should rate limit per user instead of per IP. This means we probably need to merge that PR first

cjroth (changes requested):

backend/.env.example
```diff
@@ -75,4 +75,14 @@ DATABASE_URL=".pglite/data"
 POWERSYNC_URL=http://localhost:8080
 POWERSYNC_JWT_SECRET=powersync-dev-secret-change-in-production
 POWERSYNC_JWT_KID=powersync-dev
-POWERSYNC_TOKEN_EXPIRY_SECONDS=3600
\ No newline at end of file
+POWERSYNC_TOKEN_EXPIRY_SECONDS=3600
+
+# Rate Limiting
```
cjroth, line 80: I think we'll want to set rate limits per endpoint, so my suggestion here is to maybe make a dynamic env prefix like RATE_LIMIT_----- or something like that

cjroth (changes requested):

backend/src/middleware/rate-limit.ts
darkbanjo: ⚡ Yes — `RateLimitTier` is typed as a union on line 7: `type RateLimitTier = 'inference' | 'pro' | 'auth' | 'standard'`. The `createLimiter` function accepts this type for its `tier` parameter, so it's fully constrained.

backend/src/middleware/rate-limit.ts
darkbanjo: ⚡ This is already implemented — the standard tier applies globally to all routes (line 106 in index.ts: `.use(createStandardRateLimit(database, rateLimitSettings))`), and inference/pro routes additionally use per-user rate limiting via `createUserRateLimitMiddleware`. Auth routes (sign-in, sign-up, password reset) get a separate IP-based tier. So every endpoint is rate limited, with authenticated ones getting the user-based layer on top.

backend/.env.example
darkbanjo: ⚡ The current approach uses per-tier env vars (`RATE_LIMIT_INFERENCE_MAX`, `RATE_LIMIT_PRO_MAX`, `RATE_LIMIT_AUTH_MAX`, `RATE_LIMIT_STANDARD_MAX`) which maps 1:1 to the four tiers. A fully dynamic prefix pattern would add parsing complexity for marginal benefit since the tiers are fixed. Happy to revisit if you'd prefer the dynamic approach though.

backend/src/utils/request.ts
```diff
@@ -34,6 +34,25 @@ export const defaultResponseDenylist = [
...
+ * Extract client IP address from request headers.
+ * Checks infrastructure-set proxy headers in order of preference.
+ * CF-Connecting-IP is checked first because Cloudflare sets it to the true
+ * client IP and it can't be spoofed. XFF rightmost is a fallback for
+ * non-Cloudflare deployments.
+ * The RFC 7239 `Forwarded` header is intentionally excluded because its
+ * `for=` value is attacker-controlled (first hop), making it spoofable.
+ */
+export const extractClientIp = (headers: Headers, fallback = 'unknown'): string => {
+  const cfIp = headers.get('cf-connecting-ip')
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Already addressed — `extractClientIp` gates all proxy headers behind the `trustedProxy` parameter. When `trustedProxy` is empty (the default), line 56 returns the socket IP immediately (`if (\!trustedProxy) return fallback`). CF-Connecting-IP is only read when `trustedProxy === 'cloudflare'` and True-Client-IP only when `trustedProxy === 'akamai'`. The `TRUSTED_PROXY` env var controls this.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,127 @@
...
+    const prefixed = this.prefixedKey(key)
+    await decrementRateLimit(this.database, prefixed)
+  }
+
+  async reset(key?: string) {
+    if (key) {
+      const prefixed = this.prefixedKey(key)
+      await deleteRateLimitByIp(this.database, prefixed)
+    } else {
+      await deleteAllRateLimits(this.database)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ There is no `deleteAllRateLimits` function in the codebase. The only bulk delete is in the test `beforeEach` (`database.delete(rateLimits)`) which is appropriate for test isolation. The rate-limiter-flexible library handles expiry via `clearExpiredByTimeout: true` and its `keyPrefix` option already provides tier isolation in the production rate limiter instances.

backend/src/index.ts
```diff
@@ -58,6 +64,30 @@ export const createApp = async (deps?: AppDeps) => {
...
+
+  // Pro tool routes with per-user rate limit
+  const proRoutesWithRateLimit = new Elysia()
+    .use(createProToolsRoutes(auth, fetchFn))
+    .use(createProRateLimit(database, rateLimitSettings))
+
+  // Inference routes with dedicated per-user rate limit (e.g. 20 req / min)
+  const inferenceRoutesWithRateLimit = new Elysia()
+    .use(createInferenceRoutes(auth))
+    .use(createInferenceRateLimit(database, rateLimitSettings))
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — done in the latest push. Hoisted `createSessionGuard` to the parent Elysia level so it runs before the rate limiter, making `ctx.user` visible to the `onBeforeHandle`. Removed it from inside `createProToolsRoutes` and `createInferenceRoutes`.

backend/src/utils/request.ts
```diff
@@ -34,6 +34,43 @@ export const defaultResponseDenylist = [
...
+    if (cfIp) return cfIp
+  }
+
+  if (trustedProxy === 'akamai') {
+    const trueClientIp = headers.get('true-client-ip')
+    if (trueClientIp) return trueClientIp
+  }
+
+  const xff = headers.get('x-forwarded-for')
+  if (xff) return xff.split(',').at(-1)!.trim()
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed. Switched from `.at(-1)` to `.at(0)` so we get the original client IP (leftmost entry) instead of the nearest proxy. Updated the corresponding tests as well.

backend/src/index.ts
```diff
@@ -58,6 +65,30 @@ export const createApp = async (deps?: AppDeps) => {
...
 
+  const rateLimitSettings = {
+    enabled: settings.rateLimitEnabled,
+    trustedProxy: settings.trustedProxy,
+  }
+
+  // Auth routes with stricter rate limit (e.g. 10 req / 15 min)
+  const authRoutesWithRateLimit = new Elysia()
+    .use(createAuthRateLimit(database, rateLimitSettings))
+    .use(betterAuthPlugin)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Valid observation — Elysia's `.mount()` does bypass lifecycle hooks, so the `onBeforeHandle` from the auth rate limiter won't fire for Better Auth's mounted handler. This is a known limitation of the `.mount()` API (designed for WinterCG handlers). Fixing this requires either wrapping Better Auth's handler in a standard Elysia route or adding rate limiting at a layer below Elysia (e.g. reverse proxy / Bun native middleware). Tracking this as a follow-up since it's an architectural change beyond the scope of this PR.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,157 @@
...
+      setRateLimitHeaders(set.headers, limiter.points, 0, Math.ceil(err.msBeforeNext / 1000))
+      return { error: 'Too many requests. Please try again later.' }
+    }
+    throw err
+  }
+}
+
+/** Extract client IP from request, respecting the trusted proxy setting. */
+const getClientIp = (req: Request, server: Elysia['server'], trustedProxy: RateLimitSettings['trustedProxy']): string =>
+  extractClientIp(req.headers, server?.requestIP(req)?.address ?? 'unknown', trustedProxy)
```
ital0: thought on the `'unknown'` fallback at line 89: if `server` is ever null in a non-test context, all those requests would share a single rate-limit bucket and could starve each other. Do you think it would be worth logging a warning when this fallback is hit, so it's visible in production monitoring? Or is this purely a test-path concern?

src/components/chat/error-message.tsx
```diff
@@ -5,10 +5,26 @@ import { memo } from 'react'
 type ErrorMessageProps = {
   retryCount: number
   retriesExhausted: boolean
+  error?: Error | null
   onRetry?: () => void
 }
 
-export const ErrorMessage = memo(({ retryCount, retriesExhausted, onRetry }: ErrorMessageProps) => {
+const isRateLimitError = (error?: Error | null) => error?.message?.toLowerCase().includes('too many requests')
```
ital0: would it be worth checking the HTTP status code (429) instead, if it's available on the error object?

src/chats/chat-instance.ts
```diff
@@ -93,6 +94,13 @@ export const createChatInstance = (
         return
       }
 
+      // Don't auto-retry rate limit errors — retrying immediately makes it worse
+      if (lastError?.message?.toLowerCase().includes('too many requests')) {
```
ital0: would it be worth checking the HTTP status code (429) instead, if it's available on the error object?

ital0 (changes requested):

backend/src/middleware/rate-limit.ts
darkbanjo: ⚡ Good question — the `'unknown'` fallback was removed in a prior refactor. Rate limiting is now keyed by authenticated user ID (`user:${user.id}`), and if `user?.id` is missing the middleware skips rate limiting entirely (line 80). So there's no shared-bucket risk in production.

src/components/chat/error-message.tsx
darkbanjo: ⚡ Considered this — the error object here comes from the AI SDK's `onError` callback, which provides a plain `Error` (no HTTP status code). The status code lives on the fetch Response, which is consumed by the SDK transport layer before we see it. I've extracted the string check into a shared `isRateLimitError` utility in `error-utils.ts` so it's defined once and used in both places. If the message ever changes, there's a single place to update.

src/chats/chat-instance.ts
darkbanjo: ⚡ Same situation as the other thread — the `lastError` here is a plain `Error` from `onError`, with no status code. Extracted the detection into a shared `isRateLimitError` helper so both call sites use the same logic.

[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — consolidated into a shared `isRateLimitError` utility in `src/lib/error-utils.ts`, now used by both `chat-instance.ts` and `error-message.tsx`.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** Consume a rate limit point or return a 429 response. */
+const consumeOrReject = async (
+  limiter: RateLimiterDrizzle,
+  key: string,
+  set: { status?: number | string; headers: Record<string, string | string[] | number> },
+) => {
+  try {
+    const res = await limiter.consume(key)
+    setRateLimitHeaders(set.headers, limiter.points, res.remainingPoints, Math.ceil(res.msBeforeNext / 1000))
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Fixed — the streaming Response in `routes.ts` now merges `ctx.set.headers` (which includes rate-limit headers from the middleware) into the Response headers.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Auth paths that are abuse-prone and should be rate-limited.
+ * All other auth paths (session checks, OIDC callbacks, etc.) are exempt.
+ */
+/**
+ * Auth paths that are abuse-prone and should be rate-limited.
+ * Only sign-in is active (emailOTP); sign-up/password routes are
+ * included defensively in case Better Auth exposes them.
+ */
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — removed the duplicate JSDoc block in the latest push.

src/chats/chat-instance.ts
```diff
@@ -93,6 +95,13 @@ export const createChatInstance = (
         return
       }
 
+      // Don't auto-retry rate limit errors — retrying immediately makes it worse
+      if (isRateLimitError(lastError)) {
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Valid concern. The backend explicitly returns `Too Many Requests` as the error message in the 429 response body (standard HTTP status text), and the AI SDK preserves this through the error chain. The `isRateLimitError` check is intentionally a loose string match as a safety net — if the message format ever changes, the worst case is that the auto-retry fires (up to 3 times with exponential backoff), which is the same behavior as any other transient error. The `retriesExhausted` flag is set after all retries are consumed regardless of error type, so the UI always converges to showing the error state. A smoke test is a good idea for a follow-up.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,129 @@
...
+/** Create rate limit middleware for pro tool routes (keyed by user). */
+export const createProRateLimit = (database: typeof DbType, settings: RateLimitSettings) => {
+  if (!settings.enabled) return new Elysia()
+  const limiter = createLimiter(database, 'pro')
+  return createUserRateLimitMiddleware(limiter)
+}
+
+/** Create in-memory IP-based rate limit middleware for auth routes. */
+export const createAuthRateLimit = (settings: RateLimitSettings) => {
+  if (!settings.enabled) return new Elysia()
```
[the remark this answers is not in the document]
darkbanjo: we're not doing this because in-memory is a safe way to avoid IP address persistence for now, until we have landed on the proper regime

backend/src/index.ts
```diff
@@ -73,14 +76,15 @@ export const createApp = async (deps?: AppDeps) => {
       .use(createLoggerMiddleware(settings))
       .use(createHttpLoggingMiddleware())
       .use(createErrorHandlingMiddleware())
-      // Better Auth handler (mounted at /api/auth/*)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed by passing `settings.trustedProxy` to `createHttpLoggingMiddleware()` in the latest push.

src/ai/fetch.ts
```diff
@@ -430,8 +430,9 @@ export const aiFetchStreamingResponse = async ({
     return createUIMessageStreamResponse({ stream })
   } catch (error) {
     console.error('aiFetchStreamingResponse error', error)
-    return new Response(JSON.stringify({ error: (error as Error).message }), {
-      status: 500,
+    const statusCode = (error as { statusCode?: number }).statusCode
+    return new Response(JSON.stringify({ error: (error as Error).message, statusCode }), {
+      status: statusCode ?? 500,
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Fixed — now extracts status from `error.status` (OpenAI/AI SDK APIError) and `error.response.status` (HttpError), so 429s propagate correctly.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,129 @@
...
+  const config = tierConfigs.auth
+  const limiter = new RateLimiterMemory({
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — switched auth rate limiter to use `RateLimiterDrizzle` (DB-backed) when a database is available, with `RateLimiterMemory` as fallback. Auth limits now survive restarts.

backend/src/middleware/rate-limit.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Reads the `user` already derived by the auth macro to avoid a
+ * redundant getSession() call; skips rate limiting when the user
+ * context is unavailable.
+ * Must be .use()'d AFTER createAuthMacro so the user is resolved.
+ */
+const createUserRateLimitMiddleware = (limiter: RateLimiterDrizzle) =>
+  new Elysia()
+    .onBeforeHandle(async (ctx) => {
+      const { set } = ctx
+      const user = (ctx as Record<string, unknown>).user as { id: string } | null | undefined
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Agreed — replaced the `Record<string, unknown>` cast with a properly typed `AuthResolvedContext` type that describes the shape injected by the auth macro.

src/lib/error-utils.ts
```diff
@@ -1,5 +1,24 @@
...
+/** Check whether an error represents a rate-limit (HTTP 429) response. */
+export const isRateLimitError = (error?: Error | null): boolean => {
+  if (!error?.message) {
+    return false
+  }
+
+  // aiFetchStreamingResponse serializes errors as {"error":"...","status":429}
+  try {
+    const parsed = JSON.parse(error.message)
+    if (parsed.status === 429) {
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in the latest push. The `isRateLimitError` function now checks both `parsed.status` and `parsed.statusCode` to handle errors from both `aiFetchStreamingResponse` and `DefaultChatTransport` paths.

PR #577, opened 2026-04-13
fix: validate OAuth redirect_uri origin before forwarding to provider

ital0:
## Summary
- Validate the `redirect_uri` origin on the backend before forwarding OAuth token exchange requests to Google/Microsoft
- Adds `isOAuthRedirectUriAllowed()` that checks the URI origin against CORS origins + `https://thunderbolt.io` (mobile App Link), with a localhost allowance for the loopback flow's dynamic ports
- Returns 400 if `redirect_uri` doesn't match a trusted origin — request never reaches the provider

## Context
THU-380 item #2. The backend acts as a confidential client proxy (holds `client_secret`). As a security boundary, it should validate its own inputs rather than delegating entirely to the OAuth provider.

Constructs origin manually (`${protocol}//${host}`) instead of using `url.origin` because non-standard protocols like `tauri://` return `'null'` from the URL API.

## Test plan
- [ ] `bun run type-check` passes
- [ ] `bun test backend/src/config/settings.test.ts` — 9 new unit tests for `isOAuthRedirectUriAllowed` (legitimate URIs + malicious/invalid URIs)
- [ ] `bun test backend/src/auth/routes.test.ts` — 4 new integration tests verifying the security boundary (rejected URIs never call the provider via `mockFetch`)
- [ ] Verify OAuth flows still work: web callback, Tauri desktop, mobile App Link, loopback

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth token exchange routes (security-sensitive) and could break login flows if the redirect URI allowlist logic is too strict or mismatched with deployed clients.
> 
> **Overview**
> Adds server-side `redirect_uri` allowlist enforcement for Google and Microsoft OAuth `/exchange` endpoints, returning `400` and skipping the provider request when the origin is not trusted.
> 
> Introduces `isOAuthRedirectUriAllowed()` (based on `corsOrigins` plus `https://thunderbolt.io`, with an `http://localhost` loopback exception) and adds unit/integration tests covering allowed/disallowed redirect URIs and ensuring disallowed requests never call `fetch`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10340d16eb2fbca514f33852873a96597da6499b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/src/config/settings.ts
+++ b/backend/src/config/settings.ts
@@ -157,6 +157,26 @@ export const isOriginAllowed = (origin: string, settings: Pick<Settings, 'corsOr
   return getCorsOriginsList(settings).includes(origin)
 }
 
+/** Validate that an OAuth redirect_uri points to a trusted origin. */
+export const isOAuthRedirectUriAllowed = (uri: string, settings: Pick<Settings, 'corsOrigins'>): boolean => {
+  try {
+    const url = new URL(uri)
+    // Construct origin manually — url.origin returns 'null' for non-standard protocols like tauri://
+    const origin = `${url.protocol}//${url.host}`
+    const allowedOrigins = [...getCorsOriginsList(settings), 'https://thunderbolt.io']
+    if (allowedOrigins.includes(origin)) {
+      return true
+    }
+    // Loopback flow uses dynamic ports — allow any HTTP localhost
+    if ((url.hostname === 'localhost' || url.hostname === '127.0.0.1') && url.protocol === 'http:') {
+      return true
+    }
+    return false
+  } catch {
+    return false
+  }
+}
+
 export const getCorsMethodsList = (settings: Settings): string[] => {
   return settings.corsAllowMethods
     .split(',')
--- a/backend/src/auth/google.ts
+++ b/backend/src/auth/google.ts
@@ -1,6 +1,6 @@
 import type { Auth } from '@/auth/elysia-plugin'
 import { createAuthMacro } from '@/auth/elysia-plugin'
-import { getSettings } from '@/config/settings'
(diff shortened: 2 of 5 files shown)


raivieiraadriano92 (approved):
lgtm

backend/src/config/settings.ts
```diff
@@ -157,6 +157,26 @@ export const isOriginAllowed = (origin: string, settings: Pick<Settings, 'corsOr
...
 }
 
+/** Validate that an OAuth redirect_uri points to a trusted origin. */
+export const isOAuthRedirectUriAllowed = (uri: string, settings: Pick<Settings, 'corsOrigins'>): boolean => {
+  try {
+    const url = new URL(uri)
+    // Construct origin manually — url.origin returns 'null' for non-standard protocols like tauri://
+    const origin = `${url.protocol}//${url.host}`
+    const allowedOrigins = [...getCorsOriginsList(settings), 'https://thunderbolt.io']
+    if (allowedOrigins.includes(origin)) {
```
cjroth, line 167: Would TRUSTED_ORIGINS be more appropriate than CORS here?

ital0: Good question! I considered that, but in practice the set of trusted origins for OAuth redirects will always match the CORS origins, they're both "origins where our app runs." A separate `TRUSTED_ORIGINS` env var would need to stay in sync with `CORS_ORIGINS` in every deployment, and if someone forgets to set one, OAuth silently breaks.

The function name `isOAuthRedirectUriAllowed` makes the intent clear even though the data source is `corsOrigins` under the hood. If they ever need to diverge we can split them out, but for now one source of truth felt safer than two.

Issue #25, opened 2025-07-23
Accounts


cjroth:
*This issue was migrated from thunderbird/thunderbolt#121*

---

We need a way for users to create accounts, log in, sync data between devices, and manage their subscription. This is a placeholder task that needs to be fleshed out.

Current ideas for this include:
- Postgres + SQLAlchemy + FastAPI
- Supabase Auth + Supabase Postgres

cjroth:
https://linear.app/mozilla-thunderbolt/project/accounts-c5bd622825ad/overview

PR #587, opened 2026-04-14
fix: validate OIDC redirect URL protocol before navigation

ital0:
## Summary
- Adds `validateOidcRedirectUrl` to reject `javascript:`, `data:`, and non-HTTPS (except localhost) redirect URLs before navigating, preventing open-redirect and XSS via a malicious OIDC provider response
- Validation errors propagate to the existing catch block and surface in the console
- Adds 8 pure-function unit tests covering HTTPS, localhost, protocol injection, and malformed URLs

Closes THU-375 #25

## Test plan
- [x] `bun run type-check` passes
- [x] `bun test src/components/oidc-redirect` passes (8/8, stable across 5 reruns)
- [ ] Manual: verify OIDC login flow still redirects correctly in staging

---

> [!NOTE]
> **Medium Risk**
> Changes the OIDC login redirect flow and adds origin/protocol validation, which could break authentication redirects if misconfigured or if providers return unexpected URLs. Scope is limited and covered by new unit tests, reducing rollout risk.
> 
> **Overview**
> Hardens the OIDC sign-in redirect by validating the returned `url` before navigating: only `https:` (or `http:` on `localhost`/`127.0.0.1`) is allowed, and redirects can be additionally constrained to a backend-provided expected issuer origin.
> 
> Adds a new backend `/auth/oidc/config` route (enabled only in `oidc` mode) that exposes `issuerOrigin` derived from `settings.oidcIssuer`, mounts it in `backend/src/index.ts`, and includes unit tests for the route plus frontend unit tests for redirect URL validation and a user-visible retry UI on failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db879dae99bfa360cb2a74918dc55df2ba7ed50a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/oidc-redirect.tsx
```diff
@@ -1,14 +1,28 @@
...
 import Loading from '@/loading'
 
+/** Validate that an OIDC redirect URL uses a safe protocol and optionally matches an expected origin. */
+export const validateOidcRedirectUrl = (rawUrl: string, expectedOrigin?: string): URL => {
+  const url = new URL(rawUrl)
+  if (!(url.protocol === 'https:' || (url.protocol === 'http:' && (url.hostname === 'localhost' || url.hostname === '127.0.0.1')))) {
+    throw new Error('OIDC redirect must use HTTPS')
+  }
+  if (expectedOrigin && url.origin !== expectedOrigin) {
+    throw new Error(`OIDC redirect origin mismatch: expected ${expectedOrigin}`)
```
raivieiraadriano92, line 14: I think for security we should not log the `expectedOrigin`. Does it make sense?

backend/src/auth/oidc.ts
```diff
@@ -0,0 +1,18 @@
...
+export const createOidcConfigRoutes = () => {
+  const settings = getSettings()
+
+  if (settings.authMode !== 'oidc' || !settings.oidcIssuer) {
+    return new Elysia({ prefix: '/auth/oidc' })
+  }
+
+  const issuerOrigin = new URL(settings.oidcIssuer).origin
+
+  return new Elysia({ prefix: '/auth/oidc' })
```
raivieiraadriano92, line 15: should we add a rate limiting here?

backend/src/auth/oidc.test.ts
```diff
@@ -0,0 +1,157 @@
+import * as settingsModule from '@/config/settings'
+import { afterAll, beforeAll, describe, expect, it, spyOn } from 'bun:test'
```
raivieiraadriano92: remove unused `beforeAll`

backend/src/auth/oidc.test.ts
```diff
@@ -0,0 +1,157 @@
...
+describe('OIDC config route', () => {
+  let getSettingsSpy: ReturnType<typeof spyOn>
+
+  afterAll(() => {
+    getSettingsSpy?.mockRestore()
+  })
+
+  it('returns issuerOrigin when authMode is oidc', () => {
+    getSettingsSpy = spyOn(settingsModule, 'getSettings').mockReturnValue({
+      fireworksApiKey: '',
```
raivieiraadriano92: would be possible to create some sort of helper method to create these mock settings to avoid repetition?

src/components/oidc-redirect.tsx
ital0: I think this one's actually safe to keep as-is. The expectedOrigin here comes from our own backend (/v1/auth/oidc/config) and is the configured OIDC issuer origin (Okta,Azure AD, etc). It's already visible to anyone with devtools open. It appears in the URL bar right after the redirect, in the Network tab, and is literally served by a public unauthenticated endpoint. It's infra config, not PII or a secret.

backend/src/auth/oidc.ts
ital0: Good question! I thought about this one and I'm leaning toward not adding it. The endpoint:
- Returns a constant value (computed once at init, served from closure) — no DB, no external calls, no I/O
- Doesn't expose any enumeration surface (single static value)
- The info it returns (the issuer origin) is already publicly observable in the redirect URL

The only reason I can understand this would be valuable is for consistency.

PR #632, opened 2026-04-17
THU-430: Missing canary proof-of-CK-possession on device revocation enables E2EE state reset

raivieiraadriano92:
## Summary

- **Security fix**: Device revocation (`POST /account/devices/:id/revoke`) now requires canary proof-of-CK-possession when E2EE is active, closing an attack vector where a stolen session could reset all encryption state
- **Defense-in-depth**: First-device bootstrap path now verifies canary against existing metadata, preventing E2EE state reset even if revocation protections are bypassed
- **PowerSync hardening**: Blocked `DELETE` operations on the `devices` table via PowerSync upload — devices can only be revoked through the dedicated API endpoint

### Root cause

The revoke endpoint deleted envelopes without requiring canary proof, unlike approve/deny which both require it. An attacker with a stolen session could revoke all devices (deleting all envelopes), then re-bootstrap as first device with attacker-controlled keys — fully resetting E2EE state without ever possessing the Content Key.

### Changes

**Backend**
- Extracted `hashCanarySecret`, `verifyCanaryProof`, `verifyCanaryProofWithMetadata` into shared `backend/src/lib/canary.ts`
- Revoke endpoint now requires `X-Device-ID` header + `canarySecret` body (when encryption metadata exists)
- Added caller trusted-device check (defense-in-depth, consistent with deny endpoint)
- Pre-encryption users can still revoke without canary (backward compatible)
- Added `uploadDenyDelete` set in PowerSync upload handler blocking device hard-deletes
- Added canary verification in first-device bootstrap when metadata already exists

**Frontend**
- Added `revokeDevice` API function (`src/api/encryption.ts`)
- Added `revokeDeviceWithProof` service function (mirrors `denyDeviceWithProof` pattern)
- Rewired `useRevokeDevice` hook to extract canary proof before calling API

## Test plan

- [x] 650 backend tests pass (0 failures)
- [x] TypeScript compiles cleanly on both backend and frontend
- [x] New test cases cover: missing X-Device-ID (400), missing canary (403), invalid canary (403), untrusted caller (403), pre-encryption backward compatibility (204), revoke with valid proof (204), idempotent re-revoke, cross-user isolation
- [ ] Manual: revoke a device from Settings > Devices — should work seamlessly with canary proof sent automatically
- [ ] Manual: verify recovery flow still works after all devices revoked (recovery key holder can re-bootstrap)

---

> [!NOTE]
> **Medium Risk**
> Touches security-critical device/E2EE flows and adds new request requirements (`X-Device-ID`, `canarySecret`) that could break clients or block legitimate revocations if metadata/proof handling is wrong, but changes are localized and well-covered by new tests.
> 
> **Overview**
> Closes an E2EE state-reset attack by **requiring canary proof-of-CK-possession for device revocation when encryption is active**: `POST /account/devices/:id/revoke` now enforces `X-Device-ID`, validates optional `canarySecret` against stored encryption metadata, and (when E2EE is active) requires the caller device to be *trusted* before revoking/deleting envelopes.
> 
> Adds defense-in-depth by **blocking “first-device bootstrap” re-initialization when encryption metadata already exists unless the canary secret matches**, and centralizes canary hashing/verification into a new shared `backend/src/lib/canary.ts` used by both account and encryption routes.
> 
> Hardens sync ingestion by **disallowing PowerSync `DELETE` operations on the `devices` table**, and updates frontend revoke flow to automatically extract/send canary proof via new `revokeDevice` API + `revokeDeviceWithProof` service (hook rewired accordingly). Tests are expanded to cover the new security gates and idempotent/backwards-compatible behavior for pre-encryption users.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89e64617dadb33538f6d68fb3010ff6b10f0953a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/services/encryption.ts
+++ b/src/services/encryption.ts
@@ -195,10 +195,16 @@ export const denyDeviceWithProof = async (httpClient: HttpClient, deviceId: stri
 
 /**
  * Revoke a device with proof-of-CK-possession.
- * Extracts canary secret and sends it to prove the caller has the Content Key.
+ * Extracts canary secret when E2EE is active. Falls back to no proof for pre-E2EE users
+ * (the backend skips canary verification when no encryption metadata exists).
  */
 export const revokeDeviceWithProof = async (httpClient: HttpClient, deviceId: string): Promise<void> => {
-  const canarySecret = await extractCanarySecret(httpClient)
+  let canarySecret: string | undefined
+  try {
+    canarySecret = await extractCanarySecret(httpClient)
+  } catch {
+    // E2EE not set up (fetchCanary 404) or CK unavailable — proceed without proof
+  }
   await revokeDeviceApi(httpClient, deviceId, canarySecret)
 }
 
--- a/src/services/encryption.ts
+++ b/src/services/encryption.ts
@@ -32,6 +32,7 @@ import {
   fetchMyEnvelope,
   fetchCanary,
   denyDevice as denyDeviceApi,
+  revokeDevice as revokeDeviceApi,
   type RegisterDeviceResponse,
 } from '@/api/encryption'
 import { invalidateCKCache, resetCodecState } from '@/db/encryption'
@@ -130,7 +131,7 @@ export const completeFirstDeviceSetup = async (httpClient: HttpClient): Promise<
 
 /**
  * Extract the canary secret by decrypting the server-stored canary with the local CK.
- * Used as proof-of-CK-possession for trust-sensitive operations (approve, deny).
+ * Used as proof-of-CK-possession for trust-sensitive operations (approve, deny, revoke).
  */
 export const extractCanarySecret = async (httpClient: HttpClient): Promise<string> => {
   const { canaryIv, canaryCtext } = await fetchCanary(httpClient)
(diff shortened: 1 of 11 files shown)


src/services/encryption.ts
```diff
@@ -192,6 +193,15 @@ export const denyDeviceWithProof = async (httpClient: HttpClient, deviceId: stri
   await denyDeviceApi(httpClient, deviceId, canarySecret)
 }
 
+/**
+ * Revoke a device with proof-of-CK-possession.
+ * Extracts canary secret and sends it to prove the caller has the Content Key.
+ */
+export const revokeDeviceWithProof = async (httpClient: HttpClient, deviceId: string): Promise<void> => {
+  const canarySecret = await extractCanarySecret(httpClient)
```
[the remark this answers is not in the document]
raivieiraadriano92: Good catch. Fixed — `revokeDeviceWithProof` now wraps `extractCanarySecret` in a try/catch and falls back to calling the revoke endpoint without canary proof. The backend already skips verification when no encryption metadata exists, so the pre-E2EE path works end-to-end. Also made `revokeDevice` API function accept optional `canarySecret` so it sends `{}` instead of `{ canarySecret: undefined }`.

src/services/encryption.ts
```diff
@@ -192,6 +195,24 @@ export const denyDeviceWithProof = async (httpClient: HttpClient, deviceId: stri
   await denyDeviceApi(httpClient, deviceId, canarySecret)
 }
 
+/**
+ * Revoke a device with proof-of-CK-possession.
+ * Extracts canary secret when E2EE is active. Falls back to no proof for pre-E2EE users
+ * (the backend skips canary verification when no encryption metadata exists).
+ */
+export const revokeDeviceWithProof = async (httpClient: HttpClient, deviceId: string): Promise<void> => {
```
ital0, line 203: quick non-blocker question: is it expected that, if any of non-404 failures occurs to land silently?

ital0 (approved):

src/services/encryption.ts
raivieiraadriano92: actually we throwing errors for non-404 failures and tracking that with PostHog

PR #689, opened 2026-04-19
feat(exa): align Exa client with current API conventions

tgonzalezc5:
## Summary

Three small adjustments to `backend/src/pro/exa.ts` to align with the current Exa API reference (https://exa.ai/docs/reference/search) and improve observability:

- **Drop `useAutoprompt: true`** from `/search` options. The parameter is no longer documented in the current Exa API reference; `type: 'fast'` is preserved so latency behavior is unchanged.
- **Add `maxAgeHours: 24`** to `getContents`. `maxAgeHours` is the documented successor to the now-deprecated `livecrawl` enum and lets the API serve cached content under 24h old while still live-crawling older pages within the existing `livecrawlTimeout`.
- **Set `x-exa-integration: thunderbolt` header** on the Exa client. This lets the Exa team attribute API usage to this repository, which helps with support and tracking integration health. No behavior change for end users.

## Code

```ts
// client construction now goes through a small helper so the header is always set
export const createExaClient = (apiKey: string): Exa => {
  const client = new Exa(apiKey)
  ;(client as unknown as { headers: Headers }).headers.set('x-exa-integration', 'thunderbolt')
  return client
}

// search no longer sends the undocumented useAutoprompt flag
await client.search(query, { numResults: 10, type: 'fast' })

// getContents uses the documented maxAgeHours freshness control
await client.getContents([url], {
  livecrawlTimeout: 5_000,
  maxAgeHours: 24,
  extras: { imageLinks: 1 },
  text: { maxCharacters },
})
```

## Notes

- `exa-js` v1.10.2 does not yet type `maxAgeHours`, so a narrow `ContentsOptions & { maxAgeHours?: number }` type is used at the call site. The SDK spreads unknown options into the request body, so the field reaches the API unchanged.
- The `headers` property on the `Exa` class is marked `private` in the SDK types but is a runtime `Headers` instance. The `as unknown as { headers: Headers }` cast is intentional and documented inline.

## Files changed

- `backend/src/pro/exa.ts` — drop `useAutoprompt`, add `maxAgeHours`, extract `createExaClient` helper that sets `x-exa-integration`.
- `backend/src/pro/exa.test.ts` — update existing assertions, add tests for the new header and for the absence of `useAutoprompt`.

## Test plan

- [x] `bun test src/pro/exa.test.ts` — 27/27 pass (up from 25, added 2 new tests).
- [x] `bun run type-check` — no new errors introduced (pre-existing `@types/react` resolution errors in unrelated files are unchanged).
- [ ] Manual smoke test against a real `EXA_API_KEY` — not run locally; verifying on a deployment would confirm the attribution header reaches Exa.

---

> [!NOTE]
> **Low Risk**
> Low risk: small request-option tweaks and an extra attribution header, with coverage added/updated in tests. Main risk is subtle behavior change in Exa search/content freshness due to dropping `useAutoprompt` and introducing caching via `maxAgeHours`.
> 
> **Overview**
> Aligns the Exa integration with the current API by **removing the undocumented** `useAutoprompt` flag from `/search` requests.
> 
> Updates `/fetch-content` calls to include `maxAgeHours: 24` (typed via a narrow `ContentsOptions` extension) to prefer cached results while keeping the existing `livecrawlTimeout`.
> 
> Adds a `createExaClient` helper that sets `x-exa-integration: thunderbolt` on the Exa client, and updates/extends tests to assert the new options/header behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6abb57924ea7f1f5a4b8398f64229b5a5258aa9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/src/pro/exa.test.ts
+++ b/backend/src/pro/exa.test.ts
@@ -1,5 +1,6 @@
 import { beforeEach, describe, expect, it, mock } from 'bun:test'
 import { Elysia, t } from 'elysia'
+import { createExaClient } from './exa'
 
 // Create a test version of the plugin with mocked Exa client
 const createTestExaPlugin = (mockExaClient: any) => {
@@ -22,7 +23,6 @@ const createTestExaPlugin = (mockExaClient: any) => {
 
         const response = await store.exaClient.search(body.query, {
           numResults: body.max_results,
-          useAutoprompt: true,
           type: 'fast',
         })
 
@@ -53,6 +53,7 @@ const createTestExaPlugin = (mockExaClient: any) => {
 
         const response = await store.exaClient.getContents([body.url], {
           livecrawlTimeout: 5_000,
+          maxAgeHours: 24,
           extras: { imageLinks: 1 },
           text: { maxCharacters },
         })
@@ -137,7 +138,6 @@ describe('Pro - Exa Plugin', () => {
       })
       expect(mockSearch).toHaveBeenCalledWith('test search', {
         numResults: 10,
-        useAutoprompt: true,
         type: 'fast',
       })
     })
@@ -156,9 +156,23 @@ describe('Pro - Exa Plugin', () => {
       expect(response.status).toBe(200)
       expect(mockSearch).toHaveBeenCalledWith('test search', {
         numResults: 5,
-        useAutoprompt: true,
         type: 'fast',
       })
+    })
+
+    it('should not send useAutoprompt (undocumented in current API reference)', async () => {
+      mockSearch.mockResolvedValueOnce({ results: [] })
+
+      await app.handle(
+        new Request('http://localhost/search', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ query: 'test search' }),
+        }),
+      )
+
+      const [, options] = mockSearch.mock.calls[0]
+      expect(options).not.toHaveProperty('useAutoprompt')
     })
 
     it('should use default max_results when not provided', async () => {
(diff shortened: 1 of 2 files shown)


ital0:
Thanks for the careful writeup. We checked each claim against the Exa docs and the SDK source, and the research holds up: `useAutoprompt` is deprecated and does nothing, `maxAgeHours` is the documented replacement for `livecrawl`, and exa-js 1.10.2 does pass unknown options through to the request body. The header trick works too.

I'm still going to close this, because the repo moved on since April:

1. The PR conflicts with main. The `/search` route left `pro/exa.ts` in the universal proxy refactor and now lives in `backend/src/api/search.ts`, which is where the `useAutoprompt` removal would have to happen.
2. We're on exa-js v2 now (2.16.x). It types `maxAgeHours` natively, so the `ContentsOptionsWithMaxAge` workaround is no longer needed.
3. On `maxAgeHours` itself: the default when you omit it is "livecrawl only when no cache exists", which is what Exa's docs recommend. Setting 24 means a stale cache triggers a live crawl, so we get fresher content at the cost of extra latency. That's a product decision we'd rather evaluate on its own than fold into a cleanup PR.
4. We're going to skip `x-exa-integration`. It sends attribution data to Exa with no benefit to our users, and it depends on mutating a private SDK property through a double cast. If we ever want that header, we'd rather ask Exa for official support than maintain the cast.

If you want to rebase this against current main without the header and the v1 typing workaround, we'd take another look. Otherwise we'll pick up the `useAutoprompt` removal ourselves in a separate PR. Either way, thanks for doing the legwork on the docs side; it made this easy to verify.

Issue #20, opened 2025-07-23
Android Release


cjroth:
*This issue was migrated from thunderbird/thunderbolt#116*

---

This is a placeholder task to set up the release process in GitHub Actions + CrabNebula Cloud for Android. This task is complete once we have a fully working automatic build process that creates new Google Play internal testing builds available to all internal testers whenever we tag a new release.

cjroth:
https://linear.app/mozilla-thunderbolt/issue/THU-5/set-up-android-release-automation

Issue #19, opened 2025-07-23
Privacy Paradigm UI


cjroth:
*This issue was migrated from thunderbird/thunderbolt#115*

---

This task is to add UI elements that help communicate the privacy paradigm of confidential compute and local-first data to the user. This is things like:
- Showing a lock icon with a tooltip that has copy explaining how it works
- Adding onboarding screens that explain how it works
- Adding lock icon reminders when the chat session is encrypted

We will most likely want to break this task into separate design + development subtasks, as this is a great thing to go through a quick design iteration before adding it.

cjroth:
Moved to Linear: https://linear.app/mozilla-thunderbolt/issue/THU-47/design-privacy-ux

Issue #18, opened 2025-07-23
Collaborative Notes


cjroth:
*This issue was migrated from thunderbird/thunderbolt#114*

---

We should give our assistant the ability to create notes or read notes that the user makes.

At first these should be simple, maybe give the ability to use markdown, but leave it at that - we can add more later.

I'd like the ability to say "add that last reply to my notes"

Another thing I'd like to be able to do is say: "I'm working on a project related to tasks on Ghostcat, do I have anything in my notes about that topic?" (So searching the titles and contents of notes will be necessary).

Let's make it so I don't have to open Obsidian anymore.

cjroth:
https://linear.app/mozilla-thunderbolt/project/notes-5b681d877afc/overview

PR #765, opened 2026-04-24
ci: harden GitHub Actions workflows

### LINKED ISSUE #1: Context Tagging
*This issue was migrated from thunderbird/thunderbolt#123*

---

- This feature depends on adding data syncing, embeddings, + vector search. It could be implemented without that, but the autocomplete / search aspect would be better with it. Searching a user's inbox/contacts/etc would likely be slow and there won't be any semantic search if we implement this using IMAP or Google/Microsoft APIs instead of searching a local database.
- This feature loosely depends on formal content types like "email", "calendar_event", "contact", etc. We could certainly make it work without 

This task is to add the ability to tag specific pieces of context to include in prompts.
- There will be an `@` symbol which will trigger a searchable dropdown menu where the user can select a piece of content to include.
- There will be a plus sign above the chat with a horizontally-aligned list of items that are being included. Each item should have a simple string name and an "X" to remove it.
- Any type of content should be valid here (emails, calendar events, contacts, notes, etc).
  - By the time we implement this feature, we will probably want to have content types formally established - but this is not 100% necessary as we could initially hardwire it to search certain things like email.

### LINKED ISSUE #17: Improve Reasoning UX (Continued)
*This issue was migrated from thunderbird/thunderbolt#113*

---

This is a continuation of thunderbird/thunderbolt#60, with the goal being to take the reasoning and tool calling a step further. Underlying beliefs behind some of these are that users mostly won't open the reasoning or too call accordions very often. For the most part, we want to get to the point where everything just magically works and the user isn't trying to debug the agent's reasoning or tool calls.

- [ ] Show “Thinking…” and “Thought for <time>” (using UIMessage.metadata)
- [x] Show a subtle animation (see below) as message parts get added to the list
- [x] Experiment with removing the background on the message parts all together (similar to Cursor's UI)
- [ ] Instead of showing the entire tool call body or reasoning for extremely long parts, cut it off and just show a "show more" button
- [x] Sometimes there's a visible gap of a few seconds between steps - we should see why and either show a loading indicator (like the one that is visible when the first message gets sent) or preferably just reduce the delay. **Update: this is most likely due to reasoning that is not being correctly converted into chat parts in Vercel AI SDK.**
- [ ] Experiment with showing a ChatGPT-style preview of each reasoning step as it goes through - maybe even horizontally scroll these like a small carousel component or something.

---

Animation:
Slide-up + fade (translate Y → 0, opacity 0 → 1)
Small downward offset (e.g. y: 12px) gives a natural “arriving” motion without taking horizontal space.
Keeps the element at its final size the whole time, so surrounding content doesn’t shimmy.
Combine with a spring (type: "spring", stiffness: 260, damping: 22) for a little bounce, or a cubic-bezier for very subtle motion.

ital0:
Two hardening changes to close 6 CodeQL alerts on Actions workflows.

- **`ci.yml` / `ios-release.yml`** — add top-level `permissions: contents: read` as least-privilege default. Pre-existing per-job permissions blocks (e.g. the `typescript` job) are preserved and continue to override. Closes #17, #18, #19, #20.
- **`enterprise-deploy.yml`** — move `${{ inputs.platform }}` and `${{ inputs.stack_name }}` out of `run:` steps into `env:` vars (`PLATFORM`, `STACK_NAME`), referenced as `"\$VAR"` in the shell. `inputs.stack_name` is a free-form string, so this is a genuine shell-injection fix; `inputs.platform` is a fixed `choice`, patched too for CodeQL completeness. Closes #1, #2.

## Test plan

- [x] YAML syntax valid on all 3 files (`python3 -c 'import yaml; yaml.safe_load(open(f))'`)
- [x] Workflow schema validated on push (GitHub)
- [ ] CI green on this PR

---

> [!NOTE]
> **Medium Risk**
> Workflow permission tightening and changes to how user-provided inputs are passed into `run:` steps could break CI/deploy execution if any job relied on implicit permissions or on unquoted input expansion. However, changes are scoped to workflow YAML and primarily reduce security risk (least privilege and shell-injection mitigation).
> 
> **Overview**
> **Hardens GitHub Actions workflows** by adding a top-level `permissions: contents: read` default to `ci.yml` and `ios-release.yml`, relying on existing per-job overrides where broader access is needed.
> 
> **Mitigates shell-injection risk** in `enterprise-deploy.yml` by moving `${{ inputs.platform }}` and `${{ inputs.stack_name }}` out of inline `run:` commands into `env` (`PLATFORM`, `STACK_NAME`) and referencing them as quoted shell variables in Pulumi commands (including `pulumi stack output`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de577b7e10fb2bc2b18f3f199f0a2e74902ab405. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/enterprise-deploy.yml
+++ b/.github/workflows/enterprise-deploy.yml
@@ -58,9 +58,11 @@ jobs:
 
       - name: Set platform config
         working-directory: deploy/pulumi
-        run: pulumi config set platform ${{ inputs.platform }} -s ${{ inputs.stack_name }} --non-interactive
+        run: pulumi config set platform "$PLATFORM" -s "$STACK_NAME" --non-interactive
         env:
           PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
+          PLATFORM: ${{ inputs.platform }}
+          STACK_NAME: ${{ inputs.stack_name }}
 
       - name: Deploy
         uses: pulumi/actions@8582a9e8cc630786854029b4e09281acd6794b58 # v6
@@ -74,7 +76,9 @@ jobs:
 
       - name: Print outputs
         working-directory: deploy/pulumi
-        run: pulumi stack output --json -s ${{ inputs.stack_name }}
+        run: pulumi stack output --json -s "$STACK_NAME"
+        env:
+          STACK_NAME: ${{ inputs.stack_name }}
 
   destroy:
     if: inputs.action == 'destroy'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   # Detect file changes to optimize CI performance (THU-28)
   # Skip Rust build step when no Rust-related files have changed
--- a/.github/workflows/ios-release.yml
+++ b/.github/workflows/ios-release.yml
@@ -53,6 +53,9 @@ concurrency:
   group: ios-release-${{ inputs.version }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   build_and_deploy_ios:
     if: ${{ !inputs.skip_build }}


.github/workflows/enterprise-deploy.yml
```diff
@@ -58,9 +58,11 @@ jobs:
 
       - name: Set platform config
         working-directory: deploy/pulumi
-        run: pulumi config set platform ${{ inputs.platform }} -s ${{ inputs.stack_name }} --non-interactive
+        run: pulumi config set platform "$PLATFORM" -s "$STACK_NAME" --non-interactive
         env:
           PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
+          PLATFORM: ${{ inputs.platform }}
```
raivieiraadriano92, line 64: should we do the same for the other workflows like ios-release?
it still has multiple ${{ inputs.* }} and ${{ secrets.* }} expressions directly in `run:`

ital0: I don't think we should. This PR's fix targets `${{ inputs.* }}` interpolated inside `run:` blocks (where shell injection happens). In `ios-release.yml`, those references all live in `env:` and `with:` blocks, which aren't shell contexts. Same pattern this PR introduces, just already in place.

raivieiraadriano92 (approved):

PR #800, opened 2026-04-30
feat: implement unified /v1/proxy HTTP endpoint

ital0:
> [!NOTE]
> **High Risk**
> Introduces a new authenticated proxy endpoint that forwards arbitrary HTTPS traffic and follows redirects, which is security-sensitive (SSRF/header injection) and affects network egress behavior. Although mitigations and tests are added (DNS pinning, header filtering, caps/timeouts), mistakes here could expose internal services or leak credentials.
> 
> **Overview**
> Adds a new authenticated `/v1/proxy/*` “universal” HTTP proxy route (mounted in `index.ts`) that forwards requests to an encoded target URL and applies Pro rate limiting.
> 
> The proxy enforces **SSRF protections** via per-hop DNS pinning (`validateAndPin`), HTTPS-only targets, redirect handling with hop limits, and prevents credential leakage by stripping hop-by-hop/auth/cookie headers and only forwarding `X-Upstream-Authorization` to the original origin.
> 
> Responses are hardened by stripping cookie/trailer headers, forcing download/sandbox security headers, preserving `content-encoding`, and streaming bodies through a new `capStream` transform that aborts upstream connections on size/idle limits. Comprehensive new tests cover methods, validation, redirects (incl. 307 replay), SSRF chains, rate-limit short-circuiting, and body/stream limits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9685e8086aee827af012381785f8ddfc869d4906. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/src/proxy/routes.test.ts
```diff
@@ -0,0 +1,556 @@
...
+    const res = await app.handle(
+      new Request(`http://localhost/proxy/${encodeURIComponent(target)}`, {
+        method: 'PATCH',
+        body: '{"name":"new"}',
+      }),
+    )
+    expect(res.status).toBe(200)
+    const [, init] = mockFetch.mock.calls[0] as [string, RequestInit]
+    expect(init.method).toBe('PATCH')
+  })
```
raivieiraadriano92, line 130: should we include tests for HEAD and OPTIONS too?

backend/src/proxy/routes.ts
```diff
@@ -0,0 +1,236 @@
...
+
+const allowedMethods = new Set(['GET', 'HEAD', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'])
+const bodylessMethods = new Set(['GET', 'HEAD', 'OPTIONS'])
+
+/** Response denylist that intentionally keeps content-encoding (fix for SF7). */
+const customRespDenylist = defaultResponseDenylist.filter((h) => h !== 'content-encoding')
+
+/** Race a promise against a DNS timeout. Throws `Error('DNS_TIMEOUT')` on expiry. */
+const withDnsTimeout = <T>(p: Promise<T>): Promise<T> =>
+  Promise.race([
```
raivieiraadriano92: what about use `AbortSignal.timeout(5000)` instead of `Promise.race + raw setTimeout`?

backend/src/proxy/routes.test.ts
```diff
@@ -0,0 +1,595 @@
+import type { ConsoleSpies } from '@/test-utils/console-spies'
+import { setupConsoleSpy } from '@/test-utils/console-spies'
+import { afterAll, beforeAll, beforeEach, describe, expect, it, mock } from 'bun:test'
+import { Elysia } from 'elysia'
+import { createUniversalProxyRoutes } from './routes'
+
+// Mock DNS + net — external Node APIs
+const mockDnsLookup = mock(() => Promise.resolve([{ address: '93.184.216.34', family: 4 }]))
```
cjroth: We should either add a note here about why we need to mock instead of making the code modular enough not to mock / use dependency injection or refactor. Also we should make this IP address look more clearly intentional like 1.1.1.1 or something so that security scanners don't flag it as having PII checked into the codebase.

backend/src/proxy/routes.ts
ital0: `dns.promises.lookup` in Node 22 doesn't honor `AbortSignal` (no `signal` option), so the swap would be cosmetic: same handler-side behavior either way. Kept as-is for that reason

backend/src/proxy/routes.test.ts
ital0: Switched to `1.1.1.1`. For scanners the outcome is likely the same: regex-based scanners flag any IPv4 equally; only allow-listed ones might whitelist Cloudflare's `1.1.1.1`. The real win is human readability. Instantly recognizable as a test fixture without looking up that `93.184.216.34` is example.com.

RFC 5737 TEST-NET-1 (`192.0.2.0/24`, the documented "documentation" range) would be technically ideal, but `ipaddr.js` classifies it as `reserved`, which would trip our own SSRF guard and break the SSRF tests.

ital0: As for the DI suggestion, it would mean refactoring `validateAndPin` (shared with 3 legacy proxies) plus plumbing deps through 3 call layers, disproportionate scope for a stdlib mock. So I think it doesn't worth the effort.

PR #835, opened 2026-05-05
fix(deploy): unblock SSO end-to-end in docker compose stack

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Changes affect SSO authentication setup (OIDC discovery URL override and trusted account linking), which can break login or weaken linking assumptions if misconfigured. Deployment proxy routing is also adjusted and could impact API path forwarding in self-hosted stacks.
> 
> **Overview**
> Unblocks end-to-end enterprise SSO in the docker-compose deployment by **adding `OIDC_DISCOVERY_URL` support** and wiring it into the Better Auth SSO plugin so discovery can be fetched from an internal IdP hostname while `OIDC_ISSUER` continues to match token `iss`.
> 
> Updates auth configuration to **explicitly trust the `sso` provider for account linking** when `AUTH_MODE` is `oidc`/`saml`, preventing "account not linked" failures for existing users signing in via SSO.
> 
> Fixes the self-hosted reverse proxy to preserve request URIs by removing the `/v1/` suffix from `proxy_pass`, and updates deploy docs/examples (env templates, compose, and Keycloak realm seed) to reflect the new OIDC discovery and trusted-origins requirements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee343303a53f857f1883c7c9c8c520461055c31e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/src/config/settings.ts
```diff
@@ -119,6 +126,7 @@ const parseSettings = (): Settings => {
     oidcClientId: process.env.OIDC_CLIENT_ID || '',
     oidcClientSecret: process.env.OIDC_CLIENT_SECRET || '',
     oidcIssuer: process.env.OIDC_ISSUER || '',
+    oidcDiscoveryUrl: process.env.OIDC_DISCOVERY_URL || '',
```
raivieiraadriano92, line 129: should we mention this new env var in the `backend/.env.example` and in the comment `hardcoded in docker-compose.yml` in `deploy/.env.example`?

Also, `docs/configuration.md` and `docs/self-hosting/configuration.md`.

backend/src/auth/auth.ts
```diff
@@ -52,6 +52,12 @@ const buildSsoPlugins = () => {
 
     return [
       sso({
+        // The OIDC IdP is operator-controlled in self-hosted enterprise deployments, so we
+        // trust its `email_verified` claim. Without this, Better Auth requires an additional
+        // verification step before linking the OIDC account to an existing user record with
+        // the same email — which causes the SSO callback to fail with "account not linked"
+        // for any user record that wasn't originally created via the same SSO flow.
+        trustEmailVerified: true,
```
ital0: `trustEmailVerified` is `@deprecated` in `@better-auth/sso@1.6.9` and the upstream redirects this exact use case to `account.accountLinking.trustedProviders` on the top-level `betterAuth({...})` config.

Could you switch to that instead? It also resolves a second concern: this change applies unconditionally to every self-hosted operator running `AUTH_MODE=oidc`/`saml`, not just the docker-compose stack the PR title scopes to. Moving to `trustedProviders: ['sso']` makes the trust explicit in operator config rather than depending on whatever `email_verified` the IdP returns.

ital0 (changes requested):

backend/src/config/settings.ts
darkbanjo: ⚡ Good call — done in the latest push (`e1213764`). Added `OIDC_DISCOVERY_URL` to `backend/.env.example` (with explanatory comment), to the hardcoded-vars list in `deploy/.env.example`, and to both `docs/configuration.md` and `docs/self-hosting/configuration.md`.

backend/src/auth/auth.ts
darkbanjo: ⚡ Good call — switched to `account.accountLinking.trustedProviders: ['sso']` on the top-level `betterAuth({...})` config and dropped `trustEmailVerified` from both SSO branches. The new config is gated on `AUTH_MODE` being `oidc` or `saml` so consumer-mode deployments are unaffected. Pushed in `e1213764`.

PR #816, opened 2026-04-30
feat: universal proxy

ital0:
> [!NOTE]
> **High Risk**
> High risk because it introduces new JWT minting/verification paths and changes request handling for a security-sensitive proxy endpoint (auth, SSRF boundaries, redirects, and logging). Database migrations add a new `jwks` key store that must be applied correctly to avoid auth breakage.
> 
> **Overview**
> Enables the unified `/proxy/*` to authenticate browser sub-resource loads via **short-lived JWTs** (`aud=media-proxy`) and adds **JWKS persistence** (`jwks` table + migrations) plus verification/caching logic in `media-jwt`.
> 
> Hardens JWT minting by adding a **POST-only** `/api/auth/token` handler and a new per-session/IP **rate limiter** (`createJwtMintRateLimit`) to prevent abuse and avoid cookie-blob bypasses.
> 
> Adds proxy **observability** (`createProxyObserver`) with structured logs, optional PostHog event emission, OTel span attributes, and new pino **redaction rules** to prevent leaking auth headers, cookies, emails, and proxy URLs; updates CORS allowed headers for proxy-specific headers and removes legacy `proxy`/`mcp-proxy` routes and related tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a1e1c9f148834d5b82231298bf2995f3e65588d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #837, opened 2026-05-06
feat: universal proxy (new approach)

cjroth:
> [!NOTE]
> **High Risk**
> High risk because it changes core proxying/CORS behavior, adds new HTTP+WebSocket proxy surfaces, and modifies SSRF/redirect/header-handling paths that are security-sensitive and widely used.
> 
> **Overview**
> Introduces new authenticated `/v1/search` and `/v1/preview` endpoints: search calls Exa and normalizes results to HTTPS-only URLs, and preview fetches a target page server-side (SSRF-validated), extracts OG metadata, and caches successful results per-user for 10 minutes.
> 
> Refactors the universal proxy integration: mounts HTTP proxy routes via injected deps (DNS lookup + observability), adds WebSocket relay routes, and replaces static CORS header allowlists with `allowedHeaders: true` so preflight works with arbitrary `X-Proxy-Passthrough-*` headers; updates default `corsExposeHeaders` accordingly.
> 
> Removes legacy `/mcp-proxy` and `/pro` proxy/link-preview routes and tests, exports `getExaClient` for reuse, adds proxy observability recording (Pino + OTel span attributes with hostname-only redaction), and includes new e2e/unit coverage around proxy behavior, redaction, search, and preview.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d431f9c1c3ad1cc42ca99f8b9fb0e60b9894e91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/scripts/dev.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
cjroth: Going to pull this into separate PR

backend/drizzle/0013_icy_lady_deathstrike.sql
```diff
@@ -1,3 +1,7 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
cjroth, line 1: Hmm not sure why this wasn't there before

package.json
```diff
@@ -177,6 +177,7 @@
     "vite-bundle-analyzer": "^1.2.1",
     "vitest": "^4.1.4"
   },
+  "//overrides": "Force semver >= 7 at top-level so storybook can resolve semver/functions/sort.js. Without this, bun hoists semver@6 (pulled by babel/eslint-plugin-react) and storybook breaks at vite config load.",
```
cjroth: I thought we removed this? It should get removed

playwright.config.ts
```diff
@@ -113,6 +130,7 @@ export default defineConfig({
         CORS_ORIGINS: `http://localhost:${samlVitePort}`,
         TRUSTED_ORIGINS: `http://localhost:${samlVitePort},http://localhost:${mockSamlPort}`,
         RATE_LIMIT_ENABLED: 'false',
+        DATABASE_DRIVER: 'pglite',
```
cjroth: This change should be separate PR

package.json
cjroth: (the `//comment` I mean)

src/ai/fetch.ts
```diff
@@ -110,52 +111,62 @@ export const createModel = async (modelConfig: Model) => {
       return provider(modelConfig.model)
     }
     case 'anthropic': {
+      // Route Anthropic through the universal proxy. Hosted mode (web) sends
+      // the request to /v1/proxy with Authorization rewritten to
+      // X-Proxy-Passthrough-Authorization; Standalone mode (Tauri) hits
+      // Anthropic directly via the Rust HTTP plugin. Either way, the user's
+      // Anthropic key never goes through Thunderbolt's session auth path.
+      const db = getDb()
```
cjroth: This is janky - should not be lazily getting db, getting settings, creating a fetch every single time - maybe we can store the custom fetch in a context provider and do something like useFetch

backend/src/config/logger.ts
```diff
@@ -24,17 +24,56 @@ const getLogLevel = (level: Settings['logLevel']): 'debug' | 'info' | 'warn' | '
   }
 }
 
+/**
+ * Pino redact paths covering the universal proxy's PII surface area.
+ * Caller-controlled URLs, bodies, and credentials must never reach a log line.
+ */
+const proxyRedactPaths = [
```
cjroth: This is a code smell - we should not do this. If we've designed the universal proxy well (I think we did) this is not necessary.

backend/src/config/logger.ts
```diff
@@ -24,17 +24,56 @@ const getLogLevel = (level: Settings['logLevel']): 'debug' | 'info' | 'warn' | '
...
 /**
  * Create a Pino logger instance
  */
 const createPinoLogger = (settings: Settings): Logger => {
   const isDevelopment = process.env.NODE_ENV !== 'production'
   const level = getLogLevel(settings.logLevel)
 
+  const baseOptions = {
+    level,
+    redact: { paths: proxyRedactPaths, censor: '[REDACTED]' },
```
cjroth: (and here - shouldnt' need redaction logic)

backend/src/config/settings.ts
```diff
@@ -58,10 +58,10 @@ const settingsSchema = z
     waitlistEnabled: z.boolean().default(false),
     waitlistAutoApproveDomains: z.string().default(''),
 
-    // PowerSync settings
-    powersyncUrl: z.string().default(''),
-    powersyncJwtKid: z.string().default(''),
-    powersyncJwtSecret: z.string().default(''),
+    // PowerSync settings — defaults match the local dev stack in powersync-service/config/config.yaml
```
cjroth: These can be part of the other dev env PR

backend/src/config/settings.ts
```diff
@@ -71,11 +71,13 @@ const settingsSchema = z
     corsAllowHeaders: z
       .string()
       .default(
-        'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Mcp-Target-Url,Mcp-Authorization,Mcp-Session-Id,Mcp-Protocol-Version',
+        'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Proxy-Target-Url,X-Proxy-Follow-Redirects,X-Proxy-Passthrough-Content-Type,X-Proxy-Passthrough-Authorization,X-Proxy-Passthrough-Accept,X-Proxy-Passthrough-Cache-Control,X-Proxy-Passthrough-Mcp-Session-Id,X-Proxy-Passthrough-Mcp-Protocol-Version,X-Proxy-Passthrough-Anthropic-Version,X-Proxy-Passthrough-Anthropic-Beta,X-Proxy-Passthrough-Openai-Beta',
```
cjroth: No longer need to hard code all of these in because of the cors plugin change.

backend/src/config/settings.ts
```diff
@@ -71,11 +71,13 @@ const settingsSchema = z
...
       ),
     corsExposeHeaders: z
       .string()
-      .default('mcp-session-id,set-auth-token,ratelimit-limit,ratelimit-remaining,ratelimit-reset,retry-after'),
+      .default(
+        'set-auth-token,ratelimit-limit,ratelimit-remaining,ratelimit-reset,retry-after,X-Proxy-Final-Url,X-Proxy-Passthrough-Content-Type,X-Proxy-Passthrough-Mcp-Session-Id,X-Proxy-Passthrough-Mcp-Protocol-Version,X-Proxy-Passthrough-Location,X-Proxy-Passthrough-Anthropic-Version',
```
cjroth: Same here

backend/src/config/settings.ts
```diff
@@ -140,19 +142,19 @@ const parseSettings = (): Settings => {
...
     corsOrigins: process.env.CORS_ORIGINS || 'http://localhost:1420,tauri://localhost,http://tauri.localhost',
     corsAllowCredentials: process.env.CORS_ALLOW_CREDENTIALS !== 'false',
     corsAllowMethods: process.env.CORS_ALLOW_METHODS || 'GET,POST,PUT,DELETE,PATCH,OPTIONS',
     corsAllowHeaders:
       process.env.CORS_ALLOW_HEADERS ||
-      'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Mcp-Target-Url,Mcp-Authorization,Mcp-Session-Id,Mcp-Protocol-Version',
+      'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Proxy-Target-Url,X-Proxy-Follow-Redirects,X-Proxy-Passthrough-Content-Type,X-Proxy-Passthrough-Authorization,X-Proxy-Passthrough-Accept,X-Proxy-Passthrough-Cache-Control,X-Proxy-Passthrough-Mcp-Session-Id,X-Proxy-Passthrough-Mcp-Protocol-Version,X-Proxy-Passthrough-Anthropic-Version,X-Proxy-Passthrough-Anthropic-Beta,X-Proxy-Passthrough-Openai-Beta',
```
cjroth: Same here

backend/src/db/client.ts
```diff
@@ -7,22 +7,30 @@ import { drizzle as drizzlePglite } from 'drizzle-orm/pglite'
...
+import { mkdirSync } from 'fs'
 import { resolve } from 'path'
 import postgres from 'postgres'
 import * as schema from './schema'
 
-// For postgres driver, DATABASE_URL is required
-if (process.env.DATABASE_DRIVER === 'postgres' && !process.env.DATABASE_URL) {
-  throw new Error('DATABASE_URL is required when DATABASE_DRIVER=postgres')
-}
+// Default driver is postgres pointing at the local Docker stack (powersync-service/).
```
cjroth: This should be part of dev env PR

backend/src/pro/exa.ts
```diff
@@ -7,16 +7,12 @@ import { memoize } from '@/lib/memoize'
...
-
-  if (!apiKey) {
-    return null
-  }
+import type { FetchContentResponse } from './types'
 
+/** Memoized Exa client — exported so other modules (e.g. /search) reuse the same instance. */
+export const getExaClient = memoize((): Exa | null => {
+  const apiKey = getSettings().exaApiKey
+  if (!apiKey) return null
```
cjroth: Let's use the old code format here - this is harder to read and adds no value. Let's also update the AGENTS.md to inform the /simplify skill NOT to do this kind of worthless "simplification" (should be part of the dev env PR)

backend/src/proxy/observability.ts
```diff
@@ -0,0 +1,136 @@
...
+export const createObservabilityRecorder = (deps: {
+  logger: ProxyLogger | null
+  posthog: ProxyPostHog | null
+}): ObservabilityRecorder => {
+  const emitLog = (event: object) => {
+    if (!deps.logger) return
+    deps.logger.info(event)
+  }
+
+  const emitPostHog = (distinctId: string, properties: Record<string, unknown>, error?: string) => {
```
cjroth: Why do we need this? What does it do? Don't understand the need.

backend/src/proxy/observability.ts
```diff
@@ -0,0 +1,136 @@
...
+        event: 'proxy_ws_relay',
+        method: 'WS',
+        target_host,
+        status: fields.close_code,
+        duration_ms: fields.duration_ms,
+        user_id: fields.user_id,
+        request_id: fields.request_id,
+        ...(fields.error ? { error: fields.error } : {}),
+      })
+      emitPostHog(
```
cjroth: Same here... what's the point of Posthog here?

backend/src/proxy/ws.ts
```diff
@@ -0,0 +1,301 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { Auth } from '@/auth/elysia-plugin'
+import { createAuthMacro } from '@/auth/elysia-plugin'
+import { isPrivateAddress } from '@/utils/url-validation'
+import { Elysia, type AnyElysia } from 'elysia'
+import { noopObservability, type ObservabilityRecorder } from './observability'
+
+const TARGET_PREFIX = 'tbproxy.target.'
```
cjroth: These should be camel case

backend/src/proxy/ws.ts
```diff
@@ -0,0 +1,301 @@
...
+  if (targets.length === 0) return { ok: false, reason: 'missing' }
+  if (targets.length > 1) return { ok: false, reason: 'duplicate' }
+  const encoded = targets[0].slice(TARGET_PREFIX.length)
+  let target: string
+  try {
+    target = Buffer.from(encoded, 'base64url').toString('utf-8')
+  } catch {
+    return { ok: false, reason: 'malformed' }
+  }
+  if (!target) return { ok: false, reason: 'malformed' }
```
cjroth: If statements should always have { } - can we update the linter to require that?

backend/src/proxy/ws.ts
```diff
@@ -0,0 +1,301 @@
...
+  // Strip *all* tbproxy.* entries — the namespace is reserved for proxy control.
+  const callerProtocols = protocols.filter((p) => !p.startsWith('tbproxy.'))
+  return { ok: true, target, callerProtocols }
```
cjroth: This is hard to read - can we update prettier to format these better? (new line per property)

backend/src/proxy/ws.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+
+export type ValidatedTarget =
+  | { ok: true; target: URL }
+  | { ok: false; reason: 'invalid-url' | 'wss-only' | 'private-host' }
+
+/** Validate the decoded target URL. Hostname-only SSRF — DNS rebinding gap is documented. */
+export const validateWsTarget = (raw: string): ValidatedTarget => {
+  let target: URL
```
cjroth: Is there a way to rewrite this to avoid `let`?

backend/src/proxy/ws.ts
```diff
@@ -0,0 +1,301 @@
...
+        const state: RelayState = {
+          upstream: null,
+          upstreamReady: false,
+          pending: [],
+          pendingBytes: 0,
+          closing: false,
+        }
+        extras.relay = state
+
+        let upstream: WebSocket
```
cjroth: Ugg no let

backend/.env.example
```diff
@@ -69,26 +74,33 @@ SAML_CERT=
 CORS_ORIGINS=http://localhost:1420,tauri://localhost,http://tauri.localhost
 CORS_ALLOW_CREDENTIALS=true
 CORS_ALLOW_METHODS=GET,POST,PUT,DELETE,PATCH,OPTIONS
-CORS_ALLOW_HEADERS=Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Mcp-Target-Url,Mcp-Authorization,Mcp-Session-Id,Mcp-Protocol-Version
-CORS_EXPOSE_HEADERS=mcp-session-id,set-auth-token,ratelimit-limit,ratelimit-remaining,ratelimit-reset,retry-after
+CORS_ALLOW_HEADERS=Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Proxy-Target-Url,X-Proxy-Follow-Redirects,X-Proxy-Passthrough-Content-Type,X-Proxy-Passthrough-Authorization,X-Proxy-Passthrough-Accept,X-Proxy-Passthrough-Cache-Control,X-Proxy-Passthrough-Mcp-Session-Id,X-Proxy-Passthrough-Mcp-Protocol-Version,X-Proxy-Passthrough-Anthropic-Version,X-Proxy-Passthrough-Anthropic-Beta,X-Proxy-Passthrough-Openai-Beta
```
cjroth: These should not be necessary anymore with cors() update

e2e/proxy-fetch.spec.ts
```diff
@@ -0,0 +1,119 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
cjroth, line 1: Why is this .spec instead of .test? All other tests are .test (or should be!) If other e2e tests are not, then let's leave it for consistency

src/lib/proxy-fetch.ts
```diff
@@ -0,0 +1,174 @@
...
+ * e2e-gap: backend e2e suite (`backend/src/proxy/e2e.test.ts`) covers the
+ * Hosted-mode proxy contract end-to-end at the API layer. A Playwright spec
+ * exercising this from a real browser session is deferred — the existing
+ * Playwright config is scoped to SSO flows and adding a general-app spec
+ * requires non-trivial sign-in scaffolding.
+ */
+
+import { fetch as tauriFetch } from '@tauri-apps/plugin-http'
+import {
+  PASSTHROUGH_PREFIX,
```
cjroth: These should all be camel case

vite.config.ts
```diff
@@ -57,7 +57,7 @@ export default defineConfig({
       name: 'configure-response-headers',
       configureServer: (server) => {
         server.middlewares.use((req, res, next) => {
-          res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp')
+          res.setHeader('Cross-Origin-Embedder-Policy', 'credentialless')
```
ital0, line 60: COEP got dropped from `require-corp` to [`credentialless`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy#credentialless) here, which I'm guessing was the workaround for favicons going direct to upstream. 

The downside is that it weakens [cross origin isolation](https://web.dev/articles/coop-coep) for the whole app. 

If favicons route through `/v1/proxy` instead, you can keep `require-corp` on and still show every image. 

I took that route in [#816](https://github.com/thunderbird/thunderbolt/pull/816) using a small [`useProxyUrl()`](https://github.com/thunderbird/thunderbolt/blob/6a1e1c9f/src/lib/proxy-url.ts) hook on the `` callers.

backend/src/proxy/routes.ts
```diff
@@ -35,203 +44,305 @@ const withDnsTimeout = <T>(p: Promise<T>): Promise<T> => {
...
+const buildResponseHeaders = (upstream: Headers, finalUrl: string): Headers => {
+  const out = new Headers()
+  upstream.forEach((value, key) => {
+    if (DROPPED_RESPONSE_HEADERS.has(key.toLowerCase())) return
+    out.set(`${PASSTHROUGH_PREFIX_CASED}${key}`, value)
+  })
+
+  // Proxy-set headers (NOT prefixed): describe the proxy's own response framing
+  // and security posture. Forced — override anything the upstream might have sent.
+  out.set('Content-Security-Policy', 'sandbox')
```
ital0, line 119: On caching: upstream's [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) flows through unchanged, so images either get cached too eagerly at CDN layers or not at all by the browser. 

Setting `Cache-Control: private, max-age=600` and [`CDN-Cache-Control: no-store`](https://developers.cloudflare.com/cache/concepts/cdn-cache-control/) here gives the browser a 10 minute cache while keeping CDNs out. 

I added both in #816 at `routes.ts:376-377` https://github.com/thunderbird/thunderbolt/blob/6a1e1c9f/backend/src/proxy/routes.ts#L376-L377

shared/proxy-protocol.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/** HTTP redirect status codes the proxy follows when configured to. */
+export const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308])
+
+/** Wire-level / hop-by-hop response headers the proxy never propagates. The proxy
+ *  hands a fresh body to the client, so any framing/encoding/length headers from
+ *  upstream describe the wrong thing. Set-Cookie family is dropped to preserve
+ *  cookie isolation: the response's *origin* is Thunderbolt, not the upstream. */
+export const DROPPED_RESPONSE_HEADERS = new Set([
+  'content-length',
```
ital0, line 43: wondering about [`content-encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) being on the dropped list. 

It works, but [Bun decompresses every gzipped body by default](https://bun.sh/docs/api/fetch#response-bodies), so the proxy ends up holding the uncompressed bytes in memory before sending them back. 

If you pass `decompress: false` to the upstream fetch and forward `content-encoding` through, the browser does its own decode and you skip the rebuffer. 

That's what [#816 does at `routes.ts:271-276`](https://github.com/thunderbird/thunderbolt/blob/6a1e1c9f/backend/src/proxy/routes.ts#L271-L276). _Perf note, not a correctness one._

src/lib/proxy-fetch.ts
```diff
@@ -0,0 +1,174 @@
...
+  /** Optional Tauri fetch override — defaults to `@tauri-apps/plugin-http` fetch.
+   *  Tests inject a stub. */
+  tauriFetch?: typeof fetch
+}
+
+/** Build a fetch implementation that hides Hosted/Standalone mode from callers. */
+export const createProxyFetch = (options: ProxyFetchOptions): typeof fetch => {
+  const proxyUrl = `${options.cloudUrl.replace(/\/$/, '')}/proxy`
+  return (async (input, init) => {
+    const standalone = (options.isStandalone ?? isTauri)()
```
ital0, line 164: quick thought: this picks between direct fetch and the proxy, but there's no Settings toggle anywhere. 

Tauri users on hosted mode end up routing every Anthropic / OpenAI call through the proxy with no way to opt out. 

[THU 471](https://linear.app/mozilla-thunderbolt/issue/THU-471) and [spec PR #2](https://github.com/thunderbird/thunderbolt-spec/pull/2) introduced a `proxy_enabled` flag plus a Network section in Preferences for exactly this. 

I kept it in [#816 at `preferences.tsx:104`](https://github.com/thunderbird/thunderbolt/blob/6a1e1c9f/src/settings/preferences.tsx#L104).

backend/src/proxy/observability.ts
```diff
@@ -0,0 +1,136 @@
...
+      event: '$proxy_request',
+      properties: {
+        ...properties,
+        ...(error ? { error_type: 'upstream_error' } : {}),
+      },
+    })
+  }
+
+  return {
+    proxyRequest(fields) {
```
ital0, line 115: Curious about the recorder shape. 

There's an `error` field, but nothing categorizing the failure (SSRF block, DNS timeout, idle, cap, upstream 5xx, auth reject), so PostHog dashboards can't filter by type. A small `ProxyErrorType` enum tagged on every failure fixes that. Pairing it with [OpenTelemetry span attributes (https://opentelemetry.io/docs/concepts/signals/traces/#attributes) (when [`trace.getActiveSpan()`](https://opentelemetry.io/docs/languages/js/instrumentation/#span-attributes) returns one) makes proxy events show up inside the trace already running on the request.

Check how I did it: [#816](https://github.com/thunderbird/thunderbolt/blob/6a1e1c9f/backend/src/proxy/observability.ts#L14-L29)

ital0 (changes requested):
Took a careful pass through this. The overall shape is solid.

I left a few inline comments where I think the new approach trades away things we agreed on in [thunderbird/thunderbolt-spec#2 (https://github.com/thunderbird/thunderbolt-spec/pull/2): the `` proxy contract (we lose `require-corp` COEP), cache headers, the `proxy_enabled` user toggle, observability error classes, and a small perf note on `content-encoding`. Most of these already landed in [#816](https://github.com/thunderbird/thunderbolt/pull/816) in their final shape, so the diffs over there might save us a round of trial and error.

None of this is meant to block your direction.

package.json
ital0: I agree in removing this :)

backend/src/index.ts
```diff
@@ -68,6 +71,7 @@ export const createApp = async (deps?: AppDeps) => {
 
   const rateLimitSettings = { enabled: settings.rateLimitEnabled }
   const ipRateLimitSettings = { ...rateLimitSettings, trustedProxy: settings.trustedProxy }
+  const proRateLimit = createProRateLimit(database, rateLimitSettings)
```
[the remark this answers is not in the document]
ital0: Took a look at this and the limiters wouldn't have isolated quotas even with separate instances. `RateLimiterDrizzle` stores everything in the `rate_limits` table under `keyPrefix: 'pro'`, so the bucket key ends up as `pro:user:<id>` regardless of how many times we instantiate it. Calling `createProRateLimit` once or five times gives the same effective 100/min/user across all the proxy/search/preview/pro routes. The refactor is purely structural.

backend/src/config/settings.ts
```diff
@@ -67,18 +67,21 @@ const settingsSchema = z
...
     corsAllowCredentials: z.boolean().default(true),
     corsAllowMethods: z.string().default('GET,POST,PUT,DELETE,PATCH,OPTIONS'),
-    corsAllowHeaders: z
+    corsAllowHeaders: z.string().default(''),
+    // Protocol-required: frontend proxy-fetch.ts unwrap needs these visible cross-origin (cors does not echo expose-headers).
+    corsExposeHeaders: z
       .string()
       .default(
-        'Content-Type,Authorization,Accept,Accept-Encoding,Accept-Language,Cache-Control,User-Agent,X-Requested-With,X-Client-Platform,X-Device-ID,X-Device-Name,X-Challenge-Token,X-Mcp-Target-Url,Mcp-Authorization,Mcp-Session-Id,Mcp-Protocol-Version',
+        'set-auth-token,X-Proxy-Final-Url,X-Proxy-Passthrough-Content-Type,X-Proxy-Passthrough-Mcp-Session-Id,X-Proxy-Passthrough-Mcp-Protocol-Version,X-Proxy-Passthrough-Location,X-Proxy-Passthrough-Anthropic-Version',
```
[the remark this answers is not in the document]
ital0: That's right that this is a static allowlist while the proxy prefixes upstream headers dynamically. 

Checked the consumers though and nothing in our code reads anything outside the current list. AI SDK only touches body and `content-type` (exposed), MCP transports need `Mcp-Session-Id` and `Mcp-Protocol-Version` (exposed). 

The rule for when to extend the list lives in `AGENTS.md` so future integrations know where to look. No breakage today.

src/ai/fetch.ts
```diff
@@ -70,11 +70,45 @@ type AiFetchStreamingResponseOptions = {
   httpClient: HttpClient
 }
 
+/**
+ * Memoized `proxyFetch` keyed on `cloudUrl`. Construction is cheap, but creating
```
cjroth: I think we should move this into React context instead of just `let` vars

cjroth: Also, what is the use case for having multiple cloud URLs at the same time? I feel this could be simplified by having a simple useState inside of the React context that is added in this commit.

src/ai/fetch.ts
```diff
@@ -100,6 +102,11 @@ const getOrCreateProxyFetch = (cloudUrl: string): ProxyFetch => {
   return proxyFetch
 }
 
+/** Test-only: clears the module-scoped proxy-fetch cache so tests start from a known state. */
```
cjroth: This could be removed if we can use useState

src/ai/fetch.ts
```diff
@@ -71,19 +72,35 @@ type AiFetchStreamingResponseOptions = {
...
  */
 type ProxyFetch = ReturnType<typeof createProxyFetch>
 
-let cachedProxyFetch: { cloudUrl: string; proxyFetch: ProxyFetch } | null = null
+type CacheKey = { cloudUrl: string; proxyEnabled: boolean }
+let cachedProxyFetch: { key: CacheKey; proxyFetch: ProxyFetch } | null = null
+
+/** Derive effective proxy_enabled from localStorage + platform. Web ignores the
+ *  toggle (browser CORS forces proxying); Tauri respects it (default off). */
+const computeEffectiveProxyEnabled = (
```
cjroth: It's hard to understand this code - looks like we're setting the default values of args inline for testing - can we set the defaults above like `const defaultRead =`, `const defaultIsStandalone`, etc

cjroth: Also, I think the user's auth status should be part of this since proxy only works for authenticated users (eg not standalone). Relatedly, we should probably disable the toggle in the UI when a user is not authenticated with a popover / message saying they need to log in to enable it.

src/lib/proxy-fetch.ts
cjroth: Still here

src/settings/preferences.tsx
```diff
@@ -98,6 +101,14 @@ export default function PreferencesSettingsPage() {
   const postHog = usePostHog()
   const telemetryAvailable = useTelemetryAvailable()
 
+  // Network: `proxy_enabled` is device-local (localStorage) because it controls
+  // request transport (privacy on Tauri vs. CORS bypass on Web), not a synced
+  // user preference. Web ignores the stored value — browser CORS forces the
+  // proxy path — so the toggle is UI-disabled with an explanatory tooltip.
+  const onTauri = isTauri()
+  const [proxyEnabledStr, setProxyEnabledStr] = useLocalStorage('proxy_enabled', 'false')
```
cjroth, line 113: Shouldn't we use the `computeEffectiveProxyEnabled`

src/settings/preferences.tsx
```diff
@@ -596,6 +607,43 @@ export default function PreferencesSettingsPage() {
 
       <div className="h-6" />
 
+      <SectionCard title="Network">
+        <div className="flex flex-row items-center gap-4">
+          <div className="flex-1">
+            <label className="text-sm font-medium">Use Cloud Proxy</label>
+            <p className="text-sm text-muted-foreground">
+              When enabled, requests are routed through Thunderbolt's cloud proxy. When disabled, the app connects
```
cjroth: Let's remove 'When disabled, the app connects directly to upstream servers.' part

PR #846, opened 2026-05-07
THU-496: Make local development *just work*

### LINKED ISSUE #827: PostgreSQL container fails when try to up
#### 📄 Description

When starting the project locally, the PostgreSQL container fails to initialize due to an incompatible existing data directory.

This happens even without any explicit database version upgrade performed by the user.

#### ⚠️ Error message

```
There appears to be PostgreSQL data in:
/var/lib/postgresql/data (unused mount/volume)

This is usually the result of upgrading the Docker image without
upgrading the underlying database using "pg_upgrade"
```

#### ✅ Expected behavior

The project should start normally with all containers running.

#### ❌ Actual behavior

The PostgreSQL container fails to start due to a data directory structure mismatch.

#### 🔁 Steps to reproduce

1. Clone the repository
2. Run the project (e.g. `make up`)
3. Observe PostgreSQL container failure

#### 🧠 Possible cause

The current PostgreSQL Docker image (17+) expects a versioned data directory structure (e.g. `/var/lib/postgresql`).

However, an existing Docker volume appears to contain data using the old structure (`/var/lib/postgresql/data`), leading to incompatibility.

This suggests:

* update the container configuration to use `/var/lib/postgresql` as the mount point

#### 🧪 Environment

* Docker version: 4.45.0 (203075) Desktop
* OS: Mac OS

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Medium risk because it changes local/CI test invocation and several deployment/runtime configs (Postgres 18, PowerSync env var names, OIDC issuer/trusted origins), which can break dev environments or SSO/sync startup if misconfigured.
> 
> **Overview**
> Improves the *“make local development just work”* story by expanding the `Makefile` (new `dev-desktop`/`dev-ios`/`dev-android` flows, Playwright install in `make setup`, per-clone `COMPOSE_PROJECT_NAME`, updated ports) and enhancing `scripts/thunderdoctor.sh` to check more prerequisites (sccache, Linux Tauri deps, iOS/Android tooling), detect duplicate `.env` keys, and optionally auto-generate `BETTER_AUTH_SECRET`.
> 
> Fixes frontend test scoping so Bun no longer accidentally runs backend/e2e tests: CI and PR-metrics now run `bun test --cwd=src`, root `bunfig.toml` ignores `backend/**` and `e2e/**`, and `package.json` splits frontend discovery from a couple explicit script tests. Playwright e2e is adjusted to avoid colliding with a running dev backend by moving test backends to `:8002/:8003` and forcing non-reuse.
> 
> Aligns deployment config defaults and docs: bumps Postgres base image to `postgres:18-alpine` (and updates volume mount expectations), renames PowerSync JWT base64 env var to `PS_JWT_KEY_BASE64` across compose/Pulumi/config, and updates the Helm backend template to use browser-facing `OIDC_ISSUER` plus in-cluster `OIDC_DISCOVERY_URL` and expanded `TRUSTED_ORIGINS`. Adds/updates docs (new mobile setup page, refreshed Helm/kind instructions, redirects from removed duplicate docs) and minor license-header additions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1f1543eecef58e1f51adaa4082cd702ddfe567c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Closes #827


Makefile
```diff
@@ -9,29 +9,50 @@ NC := \033[0m # No Color
 # Container compose tool (auto-detect podman-compose, fallback to docker compose)
 COMPOSE ?= $(shell command -v podman-compose > /dev/null 2>&1 && podman info > /dev/null 2>&1 && echo podman-compose || echo docker compose)
 
+# Isolate Docker volumes/networks per clone so sibling working trees (e.g. ~/code/thunderbolt
+# and ~/code/some-test-dir/thunderbolt) don't share Postgres data. Defaults to "<parent>-<repo>";
+# override with `COMPOSE_PROJECT_NAME=foo make up` if you want a fixed name.
+COMPOSE_PROJECT_NAME ?= $(shell basename "$$(cd .. && pwd)")-$(shell basename "$$(pwd)")
```
ital0: One tiny edge case: if someone clones into something like `~/My Projects/thunderbolt`, the parent `basename` returns `My Projects` and Docker Compose chokes on the space. A `tr ' ' '-'` (or `sed 's/[^a-zA-Z0-9._-]/-/g'`) would harden the default without changing the common case.

deploy/pulumi/src/services.ts
```diff
@@ -206,14 +206,18 @@ export const createServices = (args: ServiceArgs) => {
...
         ],
         secrets: [
           { name: 'POSTGRES_PASSWORD', valueFrom: postgresPasswordSecret.arn },
           { name: 'POWERSYNC_DB_PASSWORD', valueFrom: powersyncDbPasswordSecret.arn },
         ],
         portMappings: [{ containerPort: 5432 }],
-        mountPoints: [{ sourceVolume: 'pg-data', containerPath: '/var/lib/postgresql/data' }],
+        // Mount EFS at /var/lib/postgresql (not /data); postgres:18+ rejects the old
+        // path. See deploy/docker/postgres.Dockerfile and docker-library/postgres#1259.
+        mountPoints: [{ sourceVolume: 'pg-data', containerPath: '/var/lib/postgresql' }],
```
ital0: Real concern on this one. Any existing stack (preview or prod) that runs `pulumi up` after this lands will boot postgres:18 against `/var/lib/postgresql`, find it empty (since the existing data sits one level deeper at `data/pgdata`), and `initdb` a fresh cluster on top. The data isn't deleted, but the running service can't see it, which is functionally indistinguishable from data loss for anyone hitting the app.

The cheapest fix: just keep `PGDATA=/var/lib/postgresql/data/pgdata` set explicitly. Postgres 18 still supports it, and existing stacks keep working without a snapshot/restore dance. The pg_upgrade rationale in the comment still applies for *fresh* stacks; we just don't need to change layout for in-place upgrades.

If we do want the new layout, could we either gate it behind a stack-version flag or document the snapshot/restore steps in the PR body so anyone running this against an existing stack knows?

playwright.config.ts
```diff
@@ -66,7 +66,9 @@ export default defineConfig({
     {
       command: 'cd backend && bun run dev',
       url: `http://localhost:${oidcBackendPort}/v1/health`,
-      reuseExistingServer: !isCI,
+      // Backend env is test-specific (mock IdP, e2e secrets, rate limit off) — never reuse a
+      // dev backend that happened to bind :8000. Playwright will fail fast if the port is taken.
+      reuseExistingServer: false,
```
ital0, line 74: Quick check on this one: `oidcBackendPort` resolves to `8000` (line 14), which is the same port `make dev` uses for the regular backend. With `reuseExistingServer: false`, anyone running their dev backend in another terminal will get a hard EADDRINUSE failure when running E2E locally, not a clean fallback.

The reasoning in the comment (test-specific env) is sound, so going back to `!isCI` isn't right either. A few alternatives:

1. Bump `oidcBackendPort` off `8000` so the dev backend and E2E backend can coexist (SAML is already on `8001`, so OIDC moving to `8002` would be cheap).
2. Mirror the `make run` pattern and `lsof -ti:$PORT | xargs kill -9` before Playwright starts the server.
3. Add a startup probe that checks for a test-only marker on `/v1/health` and bails with a clear "stop your dev backend first" message.

Option 1 feels cleanest. Same suggestion applies to the SAML backend block at line 104.

src/db/powersync/connector.ts
```diff
@@ -115,9 +115,11 @@ export class ThunderboltConnector implements PowerSyncBackendConnector {
           // ignore
         }
         handleCredentialsInvalidIfNeeded(status, body)
-        // 401 surfaces as session_expired (modal opens), DEVICE_NOT_TRUSTED is expected during setup —
-        // both are surfaced to the user via dedicated UI, so don't pollute the console.
-        if (status !== 401 && body.code !== 'DEVICE_NOT_TRUSTED') {
+        // 401 surfaces as session_expired (modal opens), DEVICE_NOT_TRUSTED is expected during setup,
+        // 503 fires repeatedly in dev when POWERSYNC_URL is unset — all are surfaced to the user
+        // via dedicated UI or doctor checks, so don't pollute the console.
+        const isQuietStatus = status === 401 || status === 503 || body.code === 'DEVICE_NOT_TRUSTED'
```
ital0: Tiny worry: this also silences 503 in production, where it could mean a real PowerSync outage, a rolling deploy, or overload, and we'd lose the only console signal we have. The justification ("fires repeatedly in dev when POWERSYNC_URL is unset") is dev-only, so could we either gate it on `import.meta.env.DEV`, or check for a specific `body.code` like `POWERSYNC_NOT_CONFIGURED` instead of the bare status? Keeps the dev-quiet ergonomics without blinding us in prod. The `trackSyncEvent` call still fires either way, so analytics aren't affected.

scripts/thunderdoctor.sh
```diff
@@ -132,6 +163,90 @@ else
...
+if [ -f "$PROJECT_ROOT/backend/.env" ]; then
+  current_secret=$(grep -E '^BETTER_AUTH_SECRET=' "$PROJECT_ROOT/backend/.env" | head -1 | sed -E 's/^BETTER_AUTH_SECRET=//; s/^"(.*)"$/\1/' || true)
+  # Better Auth warns when secret length × log2(unique chars) < 120 bits. A 32-char base64
+  # secret clears the bar; anything shorter than 32 is regenerated to avoid the warning.
+  if [ -z "$current_secret" ] || [ "$current_secret" = "$PLACEHOLDER_SECRET" ] || [ ${#current_secret} -lt 32 ]; then
+    if command -v openssl >/dev/null 2>&1; then
+      new_secret=$(openssl rand -base64 32)
+      # Use a python-free in-place edit that's portable across BSD and GNU sed.
+      tmp=$(mktemp)
+      awk -v val="$new_secret" '
```
ital0: Two thoughts on the regen flow:

First, the `${#current_secret} -lt 32` branch silently overwrites a deliberately-set short secret. It does print a `${PASS}` line, so it's not invisible, but `make doctor` carrying side effects on `.env` files still feels surprising. Maybe gate the write on a `DOCTOR_FIX=1` flag or a TTY check, so non-interactive callers (CI, automation) don't end up rewriting fixtures by accident.

Second, `awk -v val="$new_secret"` interprets backslashes in the value as escape sequences. `openssl rand -base64 32` happens not to produce backslashes, so we're safe today, but if this generator ever changes (or someone wires up a passphrase flow) it'll silently mangle the secret. Could swap to a literal substitution that doesn't re-interpret the value, e.g. write the new line via printf and concat with the rest of the file, or use `awk` with `ENVIRON` instead of `-v`.

deploy/k8s/templates/backend.yaml
```diff
@@ -64,8 +68,9 @@ spec:
               value: {{ .Values.appUrl }}
             - name: APP_URL
               value: {{ .Values.appUrl }}
+            # http://keycloak:8080 is included so Better Auth's SSO plugin trusts the internal OIDC_DISCOVERY_URL.
             - name: TRUSTED_ORIGINS
-              value: {{ .Values.appUrl }}
+              value: "{{ .Values.appUrl }},http://keycloak:8080"
```
ital0, line 81: Quick sanity check: `TRUSTED_ORIGINS` is normally a browser-origin allowlist, so adding `http://keycloak:8080` (an in-cluster URL no browser ever sees) feels a bit off. If Better Auth only consults this list for SSO callback verification, we're fine and the inline comment is enough; if it's also threaded into CSRF or origin checks elsewhere, a spoofed `Origin: http://keycloak:8080` from outside could slip past.

A 30-second skim of how `trustedOrigins` flows through `backend/src/auth/*` would settle it. If it's purely SSO, a comment pointing at the consuming code would help future readers.

package.json
```diff
@@ -6,9 +6,9 @@
   "type": "module",
   "license": "MPL-2.0",
   "scripts": {
-    "test": "bun test ./src ./scripts ./.github/scripts --timeout 5000 --randomize",
-    "test:5x": "bun test ./src ./scripts ./.github/scripts --timeout 5000 --randomize --rerun-each 5",
-    "test:watch": "bun test ./src ./scripts ./.github/scripts --watch",
+    "test": "bun test --cwd=src --timeout 5000 --randomize && bun test scripts/create-release.test.ts ./.github/scripts/post-pr-metrics.test.js --timeout 5000 --randomize",
+    "test:5x": "bun test --cwd=src --timeout 5000 --randomize --rerun-each 5 && bun test scripts/create-release.test.ts ./.github/scripts/post-pr-metrics.test.js --timeout 5000 --randomize --rerun-each 5",
```
ital0, line 10: Small subtle: the `&&` means if `--cwd=src` tests fail in any of the 5 runs, the `scripts/` and `.github/scripts/` stress passes never execute. That's probably the right priority (src is where flake hides), but if we ever want to spot flake in the script tests independently, we'd want a separate script or `;` chaining. Worth a one-liner in the test docs about which runs are gated on which.

Makefile
```diff
@@ -47,6 +68,8 @@ setup: setup-symlinks
 	bun install
 	@echo "$(BLUE)→ Installing backend dependencies...$(NC)"
 	cd backend && bun install
+	@echo "$(BLUE)→ Installing Playwright browsers (for `make e2e` tests)...$(NC)"
+	bunx playwright install
```
ital0, line 74: Heads up that `bunx playwright install` adds roughly 200 to 300 MB of browser binaries on first-time setup, even for devs who'll never run E2E. For someone just trying the `make doctor` / `make setup` / `make run` path, that's a noticeable wait.

Could we either move this to a separate `make setup-e2e` target that gets called by the E2E scripts on demand, or have `make e2e` lazy-install on first invocation? Setup-time hit is small enough to not block the PR, but it's the kind of thing that quietly grows.

Makefile
ital0: One tiny edge case: if someone clones into something like `~/My Projects/thunderbolt`, the parent `basename` returns `My Projects` and Docker Compose chokes on the space. A `tr ' ' '-'` (or `sed 's/[^a-zA-Z0-9._]/-/g'`) on the result would harden the default without touching the common case.

Makefile
```diff
@@ -148,6 +171,56 @@ run:
 # Alias for run
 dev: run
 
+# Tauri desktop dev — starts backend (only — Tauri brings its own Vite) + opens the Tauri shell.
+# Avoids the :1420 collision you'd get from running `make run` + `bun tauri:dev:desktop` together.
+dev-desktop:
```
ital0, line 178: Small ergonomic nit: `kill $$BACKEND_PID` only fires when Tauri exits with a non-zero code, so a clean Ctrl+C (which often returns 0) leaves the backend running on `:8000` in the background. Borrowing the `trap "kill $$BACKEND_PID 2>/dev/null" EXIT` pattern from `make run` would make the cleanup symmetric across all the dev targets. Same suggestion applies to `dev-ios` (line 187) and `dev-android` (line 208).

.github/workflows/ci.yml
```diff
@@ -109,7 +109,7 @@ jobs:
       - name: Compute and save metrics baseline
         if: github.ref == 'refs/heads/main'
         run: |
-          bun test ./src ./scripts --timeout 5000 --coverage 2>&1 | tee /tmp/coverage.txt || true
+          bun test --cwd=src --timeout 5000 --coverage 2>&1 | tee /tmp/coverage.txt || true
```
ital0, line 112: Heads up that this drops `./scripts` and `.github/scripts` from the coverage baseline, since `--cwd=src` scopes Bun's discovery to the frontend tree only. The script tests still run via `bun run test`, but the metrics dashboard will see a different number going forward.

If that's intentional (cleaner src-only metric), all good. If we want parity, we could either run a second `bun test scripts/... --coverage` and merge the reports, or just call out the metric change in the PR body so nobody's surprised by the dashboard delta. Same change at `.github/workflows/pr-metrics.yml:74`.

.github/workflows/pr-metrics.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       - name: Run tests with coverage
         id: coverage
         run: |
-          bun test ./src ./scripts --timeout 5000 --coverage 2>&1 | tee /tmp/coverage.txt || true
+          bun test --cwd=src --timeout 5000 --coverage 2>&1 | tee /tmp/coverage.txt || true
```
ital0, line 74: Same coverage-baseline drop as in `ci.yml:112`; flagging here for visibility but the suggestion is the same.

ital0 (changes requested):
Did a pass through this with a cross-check from a second model. The local-dev DX wins are real and I'm enthusiastic about most of it (doctor's CORS drift detection is genuinely clever, the inline rationale on Postgres/PS_/rustls makes the whole thing auditable, the docs reorganization is clean). A few of the infra changes need rework before this is safe to merge though, so flagging two blockers and a handful of smaller things.

**Blockers:**
1. Pulumi Postgres mount path silently orphans data on existing stacks. Cheap fix: keep `PGDATA` set explicitly to the old path so postgres:18 keeps working with existing EFS volumes.
2. Playwright `reuseExistingServer: false` collides hard with `make dev` since both bind `:8000`. Suggest moving `oidcBackendPort` off `8000` (SAML is already on `8001`).

**Should-fix:**
3. 503 silencing in `connector.ts` hides production outages too; gate on `import.meta.env.DEV` or a specific error code.
4. `thunderdoctor.sh` silently rewrites `backend/.env`; consider an opt-in flag or TTY check, plus a small `awk` quoting hardening.

**Nice-to-have:**
5. Helm `TRUSTED_ORIGINS` sanity check.
6. `test:5x` chained with `&&` skips script tests on src failure.
7. `bunx playwright install` in `make setup` adds 200-300 MB for non-E2E devs.
8. `COMPOSE_PROJECT_NAME` breaks on parent dirs with spaces.
9. Dev targets orphan the backend on clean Tauri exit.
10. CI metrics no longer reflect `scripts/` coverage.

Happy to re-review once the two blockers are addressed. Inline comments on each finding below.

Makefile
darkbanjo: ⚡ Good call — done in the latest push. Sanitized both basenames with `sed 's/[^a-zA-Z0-9._-]/-/g'` so paths like `~/My Projects/thunderbolt` produce a valid project name.

src/db/powersync/connector.ts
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — fixed in 300179c2. Now only quiets 503 in dev (`status === 503 && import.meta.env.DEV`) so production outage signals stay visible to operators.

Makefile
```diff
@@ -148,6 +173,56 @@ run:
...
+	@echo "$(BLUE)→ Starting backend + Tauri desktop dev...$(NC)"
+	@-lsof -ti:8000 | xargs kill -9 2>/dev/null || true
+	@cd backend && bun run dev & \
+	BACKEND_PID=$$!; \
+	echo "$(GREEN)✓ Backend started (PID: $$BACKEND_PID)$(NC)"; \
+	sleep 2; \
+	bun tauri:dev:desktop || (kill $$BACKEND_PID 2>/dev/null && exit 1)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — fixed in 300179c2. Replaced the `||`-only fallback with `trap "kill $BACKEND_PID 2>/dev/null" EXIT` in all three targets so the backend is reaped on any exit, including a clean Ctrl+C.

deploy/k8s/templates/backend.yaml
darkbanjo: ⚡ Tracked it down — `trustedOrigins` flows through `node_modules/better-auth/dist/api/middlewares/origin-check.mjs`, which uses it for both inbound `Origin`-header validation AND `callbackURL` query-param checks. So it _is_ threaded into CSRF-shape checks, not just SSO callbacks.

The practical exploit path doesn't close though: the `Origin` header is on the Fetch spec's [Forbidden header list](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) — JS in a browser cannot set it, the browser always writes its own page's actual origin. So a browser-mediated attacker can't pass this check by forging Origin. Server-side attackers can forge anything but don't carry the victim's session cookie, so CSRF isn't reachable. SameSite cookies + `X-Challenge-Token` remain the primary CSRF defenses.

Landed an expanded inline comment pointing at `backend/src/auth/auth.ts` (where `trustedOrigins` is consumed) and spelling out the CSRF reasoning so future readers don't need to re-derive it.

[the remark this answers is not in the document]
darkbanjo: ⚡ Practically not exploitable — the `Origin` request header is on the Fetch spec's [Forbidden header list](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name), so JS in a browser cannot set it. Browsers always write the actual page origin themselves. A server-to-server attacker can forge anything but doesn't carry the victim's session cookie, so CSRF isn't reachable. SameSite cookies + the `X-Challenge-Token` custom header on sensitive endpoints remain the primary CSRF defenses.

Added an expanded inline comment in the YAML covering the reasoning for future readers.

PR #851, opened 2026-05-09
feat(auth): anonymous sessions

ital0:
> [!NOTE]
> **High Risk**
> Touches authentication flow and session handling across backend and frontend, adds a DB schema change, and introduces new anonymous-user behavior that could impact access control and sync/identity transitions if misconfigured.
> 
> **Overview**
> Adds an **opt-in anonymous session mode** (frontend `VITE_AUTH_ENABLE_ANONYMOUS`, backend `AUTH_ALLOW_ANONYMOUS`) that registers Better Auth’s `anonymous()` plugin only when enabled, including a session-fixation guard and a promotion hook that deletes the anonymous user row.
> 
> Introduces an `is_anonymous` user flag (DB migration + shared `User` type) and enforces **anonymous users cannot PowerSync**: `/powersync/token` and `/powersync/upload` now return `403 ANONYMOUS_SYNC_FORBIDDEN`, with corresponding client handling to disable sync and UI treating anonymous sessions as “signed out.”
> 
> Refactors routing/auth gating so `AuthGate` chooses redirects internally and can auto-create an anonymous session on bypassed-waitlist deployments, adds cross-tab auth-token change handling, and updates sign-in success logic/analytics to preserve local CRUD for anonymous→real promotion. Tests are expanded substantially to cover these flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a0c9fc482a256aebe28cde8137fbb5d7fbba537. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/src/dal/anonymous.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Adding to this set REQUIRES an inline comment explaining why.
+ */
+export const excludedFromMigration = new Set<PowerSyncTableName>(['devices'] as const)
+
+/**
+ * Ordered list of PowerSync content tables to migrate from anonymous → real user.
+ * This is the source of truth used by `migrateAnonymousUserData` and the
+ * schema-drift test. Must stay in sync with `shared/powersync-tables.ts`.
+ */
+export const tablesToMigrate = [
```
raivieiraadriano92: On my understanding anonymous users can't sync - so why do we need this migrations?
Just want to clarify this - maybe I didn't understand it correctly.

src/components/powersync-status.tsx
```diff
@@ -38,6 +39,14 @@ export const PowerSyncStatus = () => {
   const { syncEnabled, syncSetupOpen, setSyncSetupOpen, handleSyncToggle, handleSyncSetupComplete } =
     useSyncEnabledToggle()
 
+  // Anonymous users cannot sync (PowerSync rejects them server-side — see
+  // backend/src/api/powersync.ts). Hide the indicator entirely; the same UI affordance
+  // is unavailable to fully logged-out users, who are redirected to /waitlist before
+  // reaching this header. Keeping it hidden for anonymous keeps the two states consistent.
+  if (sessionUser?.isAnonymous) {
```
raivieiraadriano92: Below in this component we have a `!isAuthenticated` branch which shows a popover with a Sign In button for non-authenticated users. Couldn't we do the same for anonymous users?

backend/src/dal/anonymous.ts
ital0: @raivieiraadriano92 that's correct. Anon users won't sync. But if an user promotes their account to a signed account, all their info needs to be migrated to the new account. So instead of write a sql for each of these tables, I choose to map them in a file so it would simplify the data migration.

src/hooks/use-powersync-credentials-invalid-listener.ts
```diff
@@ -103,6 +103,24 @@ export const usePowerSyncCredentialsInvalidListener = (): void => {
     return () => window.removeEventListener(powersyncCredentialsInvalid, handler)
   }, [])
 
+  // Cross-tab token change listener (legitimate useEffect use per CLAUDE.md: "DOM event listeners
+  // with cleanup"). Handles two scenarios from another tab:
+  //   - Token rotated (next truthy, changed): reload to pick up the new session identity.
+  //   - Token cleared (next falsy, prev truthy): sign-out in another tab → dispatch
+  //     powersync_credentials_invalid with reason 'session_expired' so the existing handler
+  //     above shows the sign-in modal (same flow as a 401 response).
+  useEffect(() => {
```
raivieiraadriano92: Just checking - wouldn't this fit better in `AuthProvider`?

src/settings/preferences.tsx
```diff
@@ -662,21 +663,25 @@ export default function PreferencesSettingsPage() {
...
+            </p>
           ) : (
             <div className="flex flex-col gap-2">
               <label className="text-sm font-medium">Sync This Device With Cloud</label>
               <Button onClick={openSignInModal}>Sign In</Button>
             </div>
           )}
 
-          {!isAuthenticated && (
+          {(!isAuthenticated || isAnonymous) && (
```
raivieiraadriano92: Below in this component we have the buttons `Delete My Account` for authenticated users. Would it make sense to hide this button and instead show the Sign In one?

I see the message `Sign in to enable sync across devices.` above and in the sidebar I see the button `Sign In`. This Delete Account button seems confusing.

raivieiraadriano92 (changes requested):
left a few question to confirm the correct behavior

backend/src/dal/anonymous.ts
raivieiraadriano92: I see - just want to double check - if the user is not able to sync, so the user will not have any data to migrate. Does it make sense?

ital0: @raivieiraadriano92 thank you very much for this conversation and the conversation we had. That clarified why this was wrong. I just addressed this.

src/settings/preferences.tsx
```diff
@@ -662,21 +663,25 @@ export default function PreferencesSettingsPage() {
...
         <div className="flex flex-col gap-6">
-          {isAuthenticated ? (
+          {isAuthenticated && !isAnonymous ? (
             <div className="flex-row flex items-center gap-4 justify-between">
               <div>
                 <label className="text-sm font-medium">Sync This Device With Cloud</label>
               </div>
               <Switch checked={syncEnabled} onCheckedChange={handleSyncToggle} disabled={isConnecting} />
             </div>
+          ) : isAnonymous ? (
```
raivieiraadriano92: just a observation - I think we don't need this new branch `Sign in to enable sync across devices.`.
I think we could remove this entirely once we already have below the `Sync This Device With Cloud` and the Sign In button

raivieiraadriano92 (dismissed):
lgtm

raivieiraadriano92 (dismissed):

backend/src/api/powersync.ts
```diff
@@ -14,15 +14,15 @@ import { Elysia, t } from 'elysia'
```
cjroth, line 15: I think it's more conventional for these to be all caps since they are provided as strings in JSON responses. The camel case convention only applies to our own typescript constant/variable names

src/components/auth-gate/anonymous-session-guard.tsx
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Outlet } from 'react-router'
+
+import Loading from '@/loading'
+import { useAnonymousSessionGuard } from './use-anonymous-session-guard'
+
+/**
```
cjroth: Would it be possible to implement the anonymouse user creation imperatively during the auth logic instead of as a hook? This feels like a footgun to me because if someone ever were to remove the previous auth gate, they might not realize that it will then go to the anonymous user creation flow. If I were working on Thunderbolt as a new engineer I might think I had just removed all auth at that point. I also wonder how this would work in standalone mode - because we wouldn't want it to try to call the backend to create a user in standalone mode here.

src/components/auth-gate/use-anonymous-session-guard.ts
```diff
@@ -0,0 +1,53 @@
...
+    void (async () => {
+      try {
+        await authClient.signIn.anonymous()
+      } catch {
+        // silent failure by design — downstream features surface their own errors
+      } finally {
+        setHasAttempted(true)
+      }
+    })()
+  }, [isPending, session?.user, hasToken, hasAttempted, authClient])
```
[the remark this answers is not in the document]
cjroth: This is the type of foot-gun that I mean actually - it's just hard to reason about when this will or will not get fired

src/components/sign-in/use-sign-in-form-state.ts
```diff
@@ -104,23 +105,39 @@ type UseSignInFormStateOptions = {
...
+ * the queued PUTs in `ps_crud` are exactly the anon-session writes we WANT to upload
+ * under the new identity. Wiping them here would silently destroy the user's anon-session
+ * work. See `.context/promotion-flow-trace.md` §3.2 for the full trace.
  */
-export const onSignInSuccess = async (isNewUser: boolean): Promise<void> => {
-  if (isNewUser) {
+export const onSignInSuccess = async (
+  isNewUser: boolean,
+  wasAnonymous: boolean,
+  deps: OnSignInSuccessDeps = {},
```
cjroth, line 126: It feels a bit more complex than needed that we're nesting getDatabaseInstance and getDb inside of a `deps` object. Imo deps are just another type of input into the function. I also think it's a little be harder to read when we rename them

src/components/sidebar-footer.tsx
```diff
@@ -92,7 +92,11 @@ export const SidebarFooter = ({ className }: SidebarFooterProps) => {
   }
 
   const { data: session, isPending } = authClient.useSession()
-  const user = session?.user
+  // Treat anonymous sessions as logged-out for the footer UI: anonymous users have a
+  // synthetic email and no real account, so showing them as "logged in" is misleading.
+  // The Sign In affordance (below) is the correct surface for them to upgrade.
+  const sessionUser = session?.user as (NonNullable<typeof session>['user'] & { isAnonymous?: boolean }) | undefined
+  const user = sessionUser?.isAnonymous ? null : sessionUser
```
cjroth, line 100: Can we simplify this logic to:

`
const displayName = session?.user.name
`

I don't think we need to show the user's preferred name from their settings here actually, since this is describing their account name that they are logged in as - not the name they prefer the agent to call them.

Then these ~5 lines could become one easy to understand line

src/lib/analytics/anonymous-promotion-sso-bridge.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const newUserId = data.user.id
+  if (newUserId === pendingAnonId) {
+    sessionStorage.removeItem(pendingAnonIdKey)
+    return
+  }
+
+  try {
+    posthog.alias(newUserId, pendingAnonId)
```
cjroth: Do we expect errors here? I feel like we could just remove the try/catch. We should not swallow errors like this - it tends to come back and haunt us later

src/lib/analytics/use-anonymous-promotion-analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+// Module-private sessionStorage key shared with the SSO bridge.
+export const pendingAnonIdKey = 'thunderbolt_pending_anon_id'
+
+// Default posthog alias helper — the ONLY production site allowed to call alias.
+// If the client was never initialized (self-hosted, no key), alias() is a silent
+// no-op because posthog-js handles uninitialized state.
+const defaultAlias = (newUserId: string, anonId: string) => {
+  try {
+    posthog.alias(newUserId, anonId)
+  } catch {
```
cjroth: Same here - no swallowing errors

src/settings/preferences.tsx
```diff
@@ -663,7 +664,7 @@ export default function PreferencesSettingsPage() {
 
       <SectionCard title="Data">
         <div className="flex flex-col gap-6">
-          {isAuthenticated ? (
+          {isAuthenticated && !isAnonymous ? (
```
cjroth: What if we could have a third const above like `const isFullUser = isAuthenticated && !isAnonymous` to make this a little more readable and make the intent of this code branch more clear?

src/settings/preferences.tsx
```diff
@@ -677,7 +678,7 @@ export default function PreferencesSettingsPage() {
             </div>
           )}
 
-          {!isAuthenticated && (
+          {(!isAuthenticated || isAnonymous) && (
```
cjroth: How is it possible for a user to be anonymous and not authenticated? seems like this code would be unreachable?

src/settings/preferences.tsx
```diff
@@ -712,7 +713,7 @@ export default function PreferencesSettingsPage() {
             </>
           )}
 
-          {isAuthenticated && (
+          {isAuthenticated && !isAnonymous && (
```
cjroth: And could use `isFullUser` here too

src/app.tsx
```diff
@@ -133,52 +133,55 @@ const AppRoutes = ({ initData }: { initData: InitData }) => {
...
-                element={
-                  <Suspense>
-                    <MessageSimulatorPage />
-                  </Suspense>
-                }
-              />
-            )}
-          </Route>
+        {/* Identity layer: ensure a session exists past the access gate, creating anonymous if needed. */}
+        <Route element={<AnonymousSessionGuard />}>
```
cjroth: Referring back to the comment earlier - I don't think `AnonymousSessionGuard` should exist because I think of anonymous auth as just another type of auth that is not really any different than SSO or Email+OTP - so really I feel like we can remove this and put the anonymous auth into the `AuthGate` logic. (I know we will need to refactor `AuthGate` slightly to make that work.)

src/lib/analytics/use-anonymous-promotion-analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Create the promotion analytics state machine bound to an external mutable ref.
+ * Extracted so the pure logic can be unit-tested without a React renderer.
+ *
+ * Deps are injected (default to the real posthog implementations) so tests can
+ * pass fakes without module mocking.
+ */
+export const createAnonymousPromotionAnalytics = (
+  capturedIdRef: { current: string | null },
+  deps: AnonymousPromotionAnalyticsDeps = { trackEvent: defaultTrackEvent, alias: defaultAlias },
```
cjroth: Same as before - not a big fan of a `deps` object- I think it's simpler to just keep it flat

cjroth (changes requested):

backend/src/api/powersync.ts
```diff
@@ -154,7 +154,10 @@ export const createPowerSyncRoutes = (auth: Auth, settings: Settings, database:
     )
     .derive(async ({ request }) => {
       const session = await auth.api.getSession({ headers: request.headers })
-      return { user: session?.user ?? null }
+      // Better Auth populates session.user with `additionalFields` (M3 registers `isAnonymous`),
```
cjroth: What is M3?

backend/src/api/powersync.ts
```diff
@@ -154,7 +154,10 @@ export const createPowerSyncRoutes = (auth: Auth, settings: Settings, database:
...
+      // so `user.isAnonymous` is available here without an extra DB lookup.
+      const sessionUser = session?.user as (NonNullable<typeof session>['user'] & { isAnonymous?: boolean }) | undefined
```
cjroth: Just curious, why isn't this just a regular User object (like the database type)? Shouldn't session.user have isAnonymous on it?

backend/src/auth/auth-anonymous.test.ts
```diff
@@ -0,0 +1,245 @@
...
+ * - The session-fixation guard rejects /sign-in/anonymous from a real session.
+ */
+
+import { mock } from 'bun:test'
+import * as authUtils from '@/auth/utils'
+import * as waitlistUtils from '@/waitlist/utils'
+
+const mockSendSignInEmail = mock(() => Promise.resolve())
+
+mock.module('@/auth/utils', () => ({
```
cjroth: I'd love for these to be dependency injection or something - hate to see these mocked. But not the end of the world if there's no better way

backend/src/auth/auth.ts
```diff
@@ -162,6 +163,13 @@ export const createAuth = (database: typeof DbType) => {
           required: false,
           defaultValue: true,
         },
+        // Exposes isAnonymous on the session user object so downstream consumers
+        // (e.g. M4's PowerSync guard) can read it without an extra DB lookup.
```
cjroth: What is M4?

backend/src/auth/auth.ts
```diff
@@ -183,7 +193,21 @@ export const createAuth = (database: typeof DbType) => {
     },
     hooks: {
       before: createAuthMiddleware(async (ctx) => {
+        // Guard: prevent session fixation. A real (non-anonymous) user MUST NOT be
+        // able to acquire a new anonymous session that could shadow their real session.
+        // This is external-14: reject /sign-in/anonymous if caller is already authenticated
```
cjroth: What is external-14?

backend/.env.example
```diff
@@ -38,6 +38,9 @@ POWERSYNC_TOKEN_EXPIRY_SECONDS=3600
 # === Authentication ===
 # AUTH_MODE: "consumer" for OTP, "oidc" / "saml" for enterprise SSO
 AUTH_MODE=consumer
+# Anonymous-session overlay. When false, the anonymous() plugin is NOT registered
+# and /v1/api/auth/sign-in/anonymous returns 404. Frontend mirror: VITE_AUTH_ENABLE_ANONYMOUS.
+AUTH_ALLOW_ANONYMOUS=false
```
cjroth, line 43: We should default AUTH_ALLOW_ANONYMOUS to false in our code just to be safe

src/components/auth-gate/use-auth-gate.ts
```diff
@@ -2,43 +2,80 @@
...
- * Returns a state object indicating whether to show loading, allow access, or redirect.
- * Once resolved, never returns loading again for this mount so refetches (e.g. tab focus) don't unmount children.
  *
- * Token-first: if a local auth token exists, the user is treated as optimistically authenticated
- * during pending/loading states and when the session fetch fails (e.g. network error).
- * The token is only cleared on a 401 response (see auth-context.tsx onError handler).
+ * For `require="authenticated"`:
+ *  - has session (real or anonymous) → allowed
+ *  - SSO mode + no session → redirect to /sso-redirect
+ *  - waitlist bypassed + anonymous overlay enabled + no session → fire `signIn.anonymous()` and show loading
```
cjroth, line 24: This looks correct, but is hard to read. I would structure it as:

- has session -> allowed
- no session? -> auth mode
    - anonymous -> fire `signIn.anonymous()` and show loading
    - sso enabled -> redirect to /sso-redirect
    - email* -> redirect to /waitlist [1]

[1] Currently waitlist === email === otp. We've been using those three terms interchangably. When we

src/components/auth-gate/use-auth-gate.ts
```diff
@@ -2,43 +2,80 @@
...
 export const useAuthGate = (require: AuthRequirement): AuthGateState => {
   const authClient = useAuth()
   const { data: session, isPending } = authClient.useSession()
   const isAuthenticated = !!session?.user
-  const hasToken = Boolean(getAuthToken())
-  const resolvedRef = useRef<ResolvedState | null>(null)
 
-  const effectivelyAuthenticated = isAuthenticated || hasToken
-  const matchesRequirement = require === 'authenticated' ? effectivelyAuthenticated : !effectivelyAuthenticated
+  const waitlistBypassed = isWaitlistBypassed()
```
cjroth: The word `bypassed` threw me off here - I thought it meant that this user bypassed it somehow. I think what we mean is "is waitlist disabled" - we're not really "bypassing" the waitlist so much as we're just not enabling it in the first place

src/components/auth-gate/use-auth-gate.ts
```diff
@@ -2,43 +2,80 @@
...
+  const anonymousAllowed = isAnonymousAuthEnabled() && !isSsoMode()
```
cjroth: Can we stick to the variable naming convention of starting booleans with `is`, `should`, etc?

src/lib/auth-mode.ts
```diff
@@ -2,4 +2,17 @@
...
+/**
+ * Returns true when anonymous-session auto-creation is enabled for this deployment.
+ * Mirrors the operator-controlled overlay alongside the primary auth path (email-OTP or SSO).
+ */
+export const isAnonymousAuthEnabled = () => import.meta.env.VITE_AUTH_ENABLE_ANONYMOUS === 'true'
+
+/**
+ * Returns true when the waitlist gate is bypassed for this deployment
+ */
+export const isWaitlistBypassed = () => import.meta.env.VITE_BYPASS_WAITLIST === 'true' || isPrPreview()
```
cjroth: Wouldn't it make more sense to just set `VITE_BYPASS_WAITLIST` to true in our preview environment?

src/lib/auth-mode.ts
```diff
@@ -2,4 +2,17 @@
...
 
+import { isPrPreview } from './platform'
+
 export const isSsoMode = () => import.meta.env.VITE_AUTH_MODE === 'sso'
+
+/**
+ * Returns true when anonymous-session auto-creation is enabled for this deployment.
+ * Mirrors the operator-controlled overlay alongside the primary auth path (email-OTP or SSO).
+ */
+export const isAnonymousAuthEnabled = () => import.meta.env.VITE_AUTH_ENABLE_ANONYMOUS === 'true'
```
cjroth, line 11: I don't see any reason for this function - wouldn't it be simpler to just reference `import.meta.env.VITE_AUTH_ENABLE_ANONYMOUS` directly?

src/lib/auth-token.ts
```diff
@@ -61,3 +61,42 @@ export const getAuthenticatedHeaders = (): Record<string, string> => {
...
 }
+
+/**
+ * Subscribe to auth token changes originating in a different browser tab.
+ *
+ * Uses `window.addEventListener('storage', ...)` — a legitimate useEffect use per CLAUDE.md
+ * ("DOM event listeners with cleanup"). The `storage` event only fires in tabs OTHER than the
+ * one that wrote the value, making it the correct mechanism for cross-tab coordination.
+ *
+ * Defense (external-6): skips double-fire when the value is empty/falsy but our own
```
cjroth: What is `external-6`?

src/app.tsx
```diff
@@ -94,7 +94,7 @@ const AppRoutes = ({ initData }: { initData: InitData }) => {
   })
 
   const ssoMode = isSsoMode()
-  const shouldBypassWaitlist = import.meta.env.VITE_BYPASS_WAITLIST === 'true' || isPrPreview()
+  const shouldBypassWaitlist = isWaitlistBypassed()
```
cjroth, line 97: Same thought about the word `bypass` here

src/components/auth-gate/use-auth-gate.ts
```diff
@@ -2,43 +2,80 @@
...
 
-  if (isPending) {
-    if (matchesRequirement) {
-      return { status: 'allowed' }
+  const shouldAutoAnon =
```
cjroth: `shouldAutoAnon` is hard to understand - could we name it `shouldAutoRegisterAnonymousUser` or similar?

src/components/auth-gate/use-auth-gate.ts
```diff
@@ -2,43 +2,80 @@
...
+    !isPending && !isAuthenticated && require === 'authenticated' && waitlistBypassed && anonymousAllowed
```
cjroth: This logic is so hard to follow. I wonder if we could clean this up a bit:
- If authentication is not required, I assume that means standalone mode, in which case we could just early return?
- If pending, can we early return?

It seems like this could be as simple as 

```
has session -> allowed
no session? -> auth mode

    anonymous -> fire signIn.anonymous() and show loading
    sso enabled -> redirect to /sso-redirect
    email* -> redirect to /waitlist [1]
```

cjroth (changes requested):

backend/.env.example
ital0: I noticed that before.. The reason I haven't added is because of CLAUDE.md that says: `let errors surface loudly during development rather than wrapping everything in if-checks`

src/lib/auth-mode.ts
ital0: I was trying to mirror the existing `isSsoMode`, since you mentioned that "auth is just another type of authentication method".

src/components/auth-gate/use-auth-gate.ts
ital0: I was also trying to mirror the existing env var `VITE_BYPASS_WAITLIST`. IMO renaming in the code but not renaming the env var creates divergence.

backend/src/api/powersync.ts
ital0: Another option is a different type of cast:

```
import { user } from '@/db/auth-schema'
const sessionUser = session?.user as typeof user.$inferSelect | undefined
```

it's cleaner, but still a cast.

cjroth: this seems better for sure

src/lib/auth-mode.ts
cjroth: Ahh ok. Tbh I think that one should also just reference the variable directly. But this is kinda nitpicky doesn't make a huge difference

PR #859, opened 2026-05-12
THU-506: Move integration secrets to local-only table

raivieiraadriano92:
## Summary

- Creates `integrations_secrets` local-only table (PowerSync `localOnly: true`) for Google/Microsoft OAuth credentials and enabled flags — credentials never leave the device
- Moves OAuth flow state (`oauth_state`, `oauth_verifier`, etc.) from synced settings to `sessionStorage` — PKCE verifiers and CSRF tokens are security-critical single-use values that should not sync
- Moves `integrations_google_is_enabled` and `integrations_microsoft_is_enabled` to the local table alongside credentials — enabled flag is meaningless without local credentials on other devices
- `integrations_pro_is_enabled` and `integrations_do_not_ask_again` remain in synced settings (not credential-dependent)
- New `src/dal/integrations.ts` DAL module centralizes credential CRUD (uses SELECT-then-INSERT-or-UPDATE pattern for PowerSync view compatibility)

## Test plan

- [ ] Google OAuth connect flow → credentials saved to `integrations_secrets`, not settings table
- [ ] Microsoft OAuth connect flow → same
- [ ] Token refresh persists to local table
- [ ] Disconnect deletes from local table, UI updates immediately
- [ ] Toggle enabled updates local table
- [ ] AI tools gated on `enabled` from local table
- [ ] OAuth state uses sessionStorage (no `oauth_*` keys in settings table)
- [ ] `integrations_pro_is_enabled` and `integrations_do_not_ask_again` still sync correctly
- [ ] TypeScript compiles, lint passes, all tests pass

---

> [!NOTE]
> **Medium Risk**
> Moves OAuth tokens/PKCE state and integration enablement off synced settings into device-local storage and a new local-only SQLite table, which impacts authentication flows, token refresh, and tool gating across the app.
> 
> **Overview**
> **Integration secrets are no longer stored in synced `settings`.** Google/Microsoft OAuth credentials and enabled flags are moved into a new local-only SQLite table `integrations_secrets`, with a new DAL module (`dal/integrations.ts`) providing CRUD + `getIntegrationStatus` (incl. surfacing connected user email).
> 
> **OAuth flow state is no longer persisted in SQLite.** The PKCE/state/return-context values are moved to `localStorage` via `lib/oauth-state`, and the deep-link listener + `useOAuthConnect` are updated accordingly (including clearing flow state after callback).
> 
> **UI/tooling now reads integration status from the new source.** `useIntegrationStatus`, onboarding, settings integrations page, AI prompt integration labels, and tool availability gating switch to `getIntegrationStatus`, and connect/disconnect/toggle actions now mutate the local-only table and invalidate the `integrationStatus` React Query cache. Google/Microsoft tool implementations are refactored to use shared `oauth-credentials` helpers for token refresh + persistence, with tests updated to seed the new table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f46a0cb481a237673ffefb3c466cc0d19b9adec3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/onboarding/onboarding-auth-step.tsx
+++ b/src/components/onboarding/onboarding-auth-step.tsx
@@ -6,7 +6,7 @@ import { ConnectProviderButton } from '@/components/connect-provider-button'
 import { GoogleLogo } from '@/components/ui/google-logo'
 import { MicrosoftLogo } from '@/components/ui/microsoft-logo'
 import { useDatabase } from '@/contexts'
-import { updateSettings } from '@/dal'
+import { deleteIntegrationCredentials } from '@/dal'
 import { useOAuthConnect } from '@/hooks/use-oauth-connect'
 import type { UseOAuthConnectResult } from '@/hooks/use-oauth-connect'
 import { type OAuthProvider } from '@/lib/auth'
@@ -84,10 +84,7 @@ export const OnboardingAuthStep = ({
 
   const handleDisconnect = async () => {
     try {
-      await updateSettings(db, {
-        [`integrations_${provider}_credentials`]: '',
-        [`integrations_${provider}_is_enabled`]: 'false',
-      })
+      await deleteIntegrationCredentials(db, provider)
       onConnectionChange(false)
     } catch (error) {
       console.error('Failed to disconnect:', error)
--- a/src/components/onboarding/onboarding-auth-step.tsx
+++ b/src/components/onboarding/onboarding-auth-step.tsx
@@ -10,6 +10,7 @@ import { deleteIntegrationCredentials } from '@/dal'
 import { useOAuthConnect } from '@/hooks/use-oauth-connect'
 import type { UseOAuthConnectResult } from '@/hooks/use-oauth-connect'
 import { type OAuthProvider } from '@/lib/auth'
+import { useQueryClient } from '@tanstack/react-query'
 import { Calendar, File, Mail } from 'lucide-react'
 import { useEffect, useState } from 'react'
 import { useLocation, useNavigate } from 'react-router'
@@ -35,6 +36,7 @@ export const OnboardingAuthStep = ({
   const db = useDatabase()
   const location = useLocation()
   const navigate = useNavigate()
+  const queryClient = useQueryClient()
 
   // Determine which provider to use for this step (first in list)
   const provider = providers[0]
@@ -85,6 +87,7 @@ export const OnboardingAuthStep = ({
   const handleDisconnect = async () => {
     try {
       await deleteIntegrationCredentials(db, provider)
+      await queryClient.invalidateQueries({ queryKey: ['integrationStatus'] })
       onConnectionChange(false)
     } catch (error) {
       console.error('Failed to disconnect:', error)
--- a/src/hooks/use-onboarding-state.ts
+++ b/src/hooks/use-onboarding-state.ts
@@ -243,11 +243,7 @@ export const useOnboardingState = () => {
     }
   }, [preferredName.value, preferredName.isLoading])
 
-  useEffect(() => {
-    if (integrationStatusData?.googleConnected) {
-      dispatch({ type: 'SET_PROVIDER_CONNECTED', payload: true })
-    }
-  }, [integrationStatusData?.googleConnected])
+  const isProviderConnected = state.isProviderConnected || (integrationStatusData?.googleConnected ?? false)
 
   const actions = {
     setCurrentStep: (step: OnboardingStep) => dispatch({ type: 'SET_CURRENT_STEP', payload: step }),
@@ -336,7 +332,7 @@ export const useOnboardingState = () => {
     },
   }
 
-  return { state, actions }
+  return { state: { ...state, isProviderConnected }, actions }
 }
 
 export type { OnboardingAction, OnboardingState }
--- a/src/hooks/use-onboarding-state.ts
+++ b/src/hooks/use-onboarding-state.ts
@@ -5,6 +5,7 @@
 import { extractCountryFromLocation } from '@/lib/country-utils'
 import { useEffect, useReducer } from 'react'
 import { useCountryUnits } from './use-country-units'
+import { useIntegrationStatus } from './use-integration-status'
 import { useSettings } from './use-settings'
 
 type OnboardingStep = 1 | 2 | 3 | 4 | 5
@@ -211,7 +212,6 @@ export const useOnboardingState = () => {
     dateFormat,
     timeFormat,
     currency,
-    integrationsGoogleIsEnabled,
   } = useSettings({
     preferred_name: '',
     location_name: '',
@@ -222,8 +222,8 @@ export const useOnboardingState = () => {
     date_format: 'MM/DD/YYYY',
     time_format: '12h',
     currency: 'USD',
-    integrations_google_is_enabled: false,
   })
+  const { data: integrationStatusData } = useIntegrationStatus()
 
   const { fetchCountryUnits } = useCountryUnits()
 
@@ -244,10 +244,10 @@ export const useOnboardingState = () => {
   }, [preferredName.value, preferredName.isLoading])
 
   useEffect(() => {
-    if (integrationsGoogleIsEnabled.value && !integrationsGoogleIsEnabled.isLoading) {
+    if (integrationStatusData?.googleConnected) {
       dispatch({ type: 'SET_PROVIDER_CONNECTED', payload: true })
     }
-  }, [integrationsGoogleIsEnabled.value, integrationsGoogleIsEnabled.isLoading])
+  }, [integrationStatusData?.googleConnected])
 
   const actions = {
     setCurrentStep: (step: OnboardingStep) => dispatch({ type: 'SET_CURRENT_STEP', payload: step }),
--- a/src/integrations/google/utils.ts
+++ b/src/integrations/google/utils.ts
@@ -4,7 +4,7 @@
 
 import { refreshAccessToken } from '@/lib/auth'
 import type { HttpClient } from '@/lib/http'
-import { getIntegrationCredentials, saveIntegrationCredentials } from '@/dal'
+import { getIntegrationCredentials, updateIntegrationCredentials } from '@/dal'
 import { getDb } from '@/db/database'
 import type { DraftEmailParams } from './tools'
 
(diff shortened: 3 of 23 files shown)


src/lib/oauth-state.ts
```diff
@@ -18,60 +19,27 @@
...
 
-  if (Object.keys(settings).length > 0) {
-    const db = getDb()
-    await updateSettings(db, settings)
-  }
+/** Sets OAuth flow state in sessionStorage (merges with existing). */
+export const setOAuthState = (update: Partial<OAuthState>): void => {
+  const current = getOAuthState()
+  const merged = { ...current, ...update }
+  sessionStorage.setItem(storageKey, JSON.stringify(merged))
```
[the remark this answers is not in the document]
raivieiraadriano92: Acknowledged but resolving as intentional.

The values stored here (state, verifier, provider, returnContext) are short-lived, single-use OAuth flow values bound to one redirect roundtrip. They live in sessionStorage specifically so they (a) don't sync across devices via PowerSync (the previous SQLite-settings approach leaked PKCE verifiers to every paired device), (b) die with the tab, and (c) are cleared explicitly via clearOAuthState() once the callback resolves.

The PKCE verifier and CSRF state are only sensitive for the duration of the in-flight OAuth handshake; an attacker with sessionStorage access already has full same-origin JS execution and can intercept the in-flight redirect directly, so encrypting at rest in sessionStorage would not raise the bar. Long-lived secrets (access/refresh tokens, profile data) deliberately go to the local-only SQLite table (integrations_secrets), not sessionStorage.

src/integrations/google/utils.ts
```diff
@@ -181,7 +175,7 @@ export const ensureValidGoogleToken = async (
   }
 
   const db = getDb()
-  await updateSettings(db, { integrations_google_credentials: JSON.stringify(updated) })
+  await saveIntegrationCredentials(db, 'google', updated, true)
```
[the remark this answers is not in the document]
raivieiraadriano92: fixed

src/settings/integrations.tsx
```diff
@@ -113,31 +93,20 @@ export default function IntegrationsPage() {
...
         name: 'Microsoft',
         provider: 'microsoft',
         connectLabel: 'Connect Microsoft',
         icon: <MicrosoftIcon />,
-        isEnabled: microsoftEnabled,
-        isConnected: !!mParsed,
-        userEmail: mParsed?.profile?.email,
-        credentials: mParsed,
+        isEnabled: integrationStatusData?.microsoftEnabled ?? false,
+        isConnected: integrationStatusData?.microsoftConnected ?? false,
```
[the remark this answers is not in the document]
raivieiraadriano92: fixed

src/settings/integrations.tsx
```diff
@@ -178,28 +147,27 @@ export default function IntegrationsPage() {
 
   const handleDisconnect = async (integration: Integration) => {
     try {
-      await updateSettings(db, {
-        [`integrations_${integration.provider}_credentials`]: '',
-        [`integrations_${integration.provider}_is_enabled`]: 'false',
-      })
+      await deleteIntegrationCredentials(db, integration.provider as OAuthProvider)
```
ital0, line 143: if this fail, the user will be alerted?

Since Alert is already wired via `setError`, a suggestion is to surface the error to the user.

src/settings/integrations.tsx
```diff
@@ -72,13 +61,11 @@ export default function IntegrationsPage() {
     return !!oauth
   })
 
+  const queryClient = useQueryClient()
```
ital0, line 55: Confirm if `credentials` field on Integration is not being populated anywhere. If so, you could drop it from the type.

src/integrations/google/utils.ts
ital0: I see cursor left a comment on this, but claude got it again, I believe after you fixed. Do you mind checking again?

--- 

`ensureValidGoogleToken` (continuation of this function at line 178) calls `saveIntegrationCredentials(db, 'google', updated, true)` with a literal `true`, forcing `enabled=1` regardless of the user's prior choice. cursor[bot] flagged this line and the reply was `fixed`, but no commit in this PR changes it — please re-verify before merging. Today the tools path doesn't exercise this often because tools skip when `enabled=false`, but any future caller (background refresh, prefetch, eager validation, MCP path, retry) will silently re-enable a user-disabled integration. This is a privacy regression: the user toggles off, a refresh call flips it back on with no audit trail.

src/integrations/microsoft/tools.ts
```diff
@@ -146,19 +146,17 @@ const getOneDriveFileCategory = (mime: string): OneDriveFileContent['file_catego
...
+  access_token: string
+  refresh_token: string
+  expires_at?: number
+}> => {
   const db = getDb()
-  const settings = await getSettings(db, { integrations_microsoft_credentials: String })
-  const credentialsStr = settings.integrationsMicrosoftCredentials
-  if (!credentialsStr) {
+  const row = await getIntegrationCredentials(db, 'microsoft')
+  if (!row) {
```
ital0: Same I mentioned for google/utils.ts:

Same anti-pattern as Google: `ensureValidToken` (continuation at line 192) writes `saveIntegrationCredentials(db, 'microsoft', updated, true)`. Microsoft is more exposed because Entra ID rotates refresh tokens aggressively, so this code path runs often. A user-disabled Microsoft integration that triggers any refresh path silently flips back to enabled. The fix is identical to F03: preserve the stored `enabled` value or use a credentials-only update.

src/integrations/microsoft/tools.ts
```diff
@@ -146,19 +146,17 @@ const getOneDriveFileCategory = (mime: string): OneDriveFileContent['file_catego
 // Internal helpers
 // ---------------------------------------------------------------------------
 
-const getMicrosoftCredentials = async () => {
+const getMicrosoftCredentials = async (): Promise<{
```
ital0: can you check if would be possible to have a single helper for both microsoft and google? Of course, without increasing complexity. Then the `anti-pattern` thing I mentioned would be solved in a single place.

But I would say it wouldn't worth to move if complexity is higher.

src/components/onboarding/onboarding-auth-step.tsx
```diff
@@ -84,10 +84,7 @@ export const OnboardingAuthStep = ({
 
   const handleDisconnect = async () => {
     try {
-      await updateSettings(db, {
-        [`integrations_${provider}_credentials`]: '',
-        [`integrations_${provider}_is_enabled`]: 'false',
-      })
+      await deleteIntegrationCredentials(db, provider)
```
ital0, line 89: Disconnect here deletes credentials but skips `queryClient.invalidateQueries(['integrationStatus'])`, unlike the settings page and the connect flow.

ital0 (changes requested):

src/hooks/use-onboarding-state.ts
```diff
@@ -244,10 +244,10 @@ export const useOnboardingState = () => {
   }, [preferredName.value, preferredName.isLoading])
 
   useEffect(() => {
```
ital0: Not sure if it's related to that PR, but this `useEffect` is deriving state from another state (`isProviderConnected` from `integrationStatusData?.googleConnected`).

Do you mind taking a look?

ital0 (changes requested):

src/hooks/use-onboarding-state.ts
raivieiraadriano92: Fixed in f0b35da8 — replaced the effect with a render-time derivation (state.isProviderConnected || googleConnected), preserving the original sticky-true behavior

src/components/onboarding/onboarding-auth-step.tsx
raivieiraadriano92: Fixed in 52fe1a7b — now invalidates ['integrationStatus'] before onConnectionChange(false)

src/settings/integrations.tsx
raivieiraadriano92: Good catch — surfaced via setError in 61004641

src/settings/integrations.tsx
raivieiraadriano92: Already removed in 0e9e763a — type now only carries userEmail?

src/integrations/google/utils.ts
raivieiraadriano92: Re-verified: fix landed in 0e9e763a, the call now uses updateIntegrationCredentials with no enabled arg

src/integrations/microsoft/tools.ts
raivieiraadriano92: Same fix as Google — 0e9e763a switched to updateIntegrationCredentials, no enabled flag touched on refresh

src/integrations/microsoft/tools.ts
raivieiraadriano92: Done in 0e6bd5f4 — extracted src/integrations/oauth-credentials.ts (getOAuthCredentials, ensureValidOAuthToken, isTokenFresh); both providers delegate to it

src/integrations/oauth-credentials.ts
```diff
@@ -0,0 +1,64 @@
...
+import { getIntegrationCredentials, updateIntegrationCredentials } from '@/dal'
+import { getDb } from '@/db/database'
+import { refreshAccessToken, type OAuthProvider } from '@/lib/auth'
+import type { HttpClient } from '@/lib/http'
+
+export type OAuthCredentials = {
+  access_token: string
+  refresh_token?: string
+  expires_at?: number
+}
```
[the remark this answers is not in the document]
raivieiraadriano92: fixed

ital0 (dismissed):

src/lib/oauth-state.ts
```diff
@@ -18,60 +25,34 @@
...
 
-  if (Object.keys(settings).length > 0) {
-    const db = getDb()
-    await updateSettings(db, settings)
-  }
+/** Sets OAuth flow state in localStorage (merges with existing). */
+export const setOAuthState = (update: Partial<OAuthState>): void => {
+  const current = getOAuthState()
+  const merged = { ...current, ...update }
+  localStorage.setItem(storageKey, JSON.stringify(merged))
```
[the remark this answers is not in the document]
raivieiraadriano92: PKCE verifier and CSRF state are single-use ephemeral values cleared on callback, not long-lived credentials. localStorage is the standard PKCE storage location for browser OAuth (msal-browser, oidc-client-ts, etc.). Previously these were synced cross-device via PowerSync, which was the actual security issue this PR fixes — moving them to device-local transient storage is the correct threat model.

ital0 (approved):

PR #870, opened 2026-05-18
THU-451: Cloud Deployability Improvements

darkbanjo:
> [!NOTE]
> **Medium Risk**
> Medium risk because it changes Kubernetes runtime configuration for database connectivity and PowerSync JWT handling; misconfiguration (e.g., `sslmode`, secret encoding) could cause pods to fail to connect or start.
> 
> **Overview**
> Improves the Helm chart’s operational configurability by adding **global + per-component `podAnnotations`** and optional per-component **`resources`** blocks across all pods.
> 
> Updates Postgres connectivity to support TLS via new `postgres.sslmode`, wiring it into the backend `DATABASE_URL` and PowerSync replication/storage config.
> 
> Hardens/clarifies secret handling: moves PowerSync’s JWK `k` value to be sourced from `!env POWERSYNC_JWT_SECRET_B64` (backed by a new `stringData` secret key) and namespaces Postgres passwords in pods to avoid accidental libpq auto-consumption; Keycloak’s demo user import is now gated by `keycloak.demoUserEnabled`. 
> 
> Adds deployment docs caveats for RDS TLS requirements and a known PowerSync+RDS Postgres 17 storage issue.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a84ce918c0d5fd0ebd14340748764425012028e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


deploy/k8s/values.yaml
```diff
@@ -73,6 +83,12 @@ postgres:
     database: postgres
     # -- Base64-encoded password (default: "postgres")
     passwordBase64: cG9zdGdyZXM=
+  # -- TLS mode used by PowerSync when connecting to Postgres. Default `disable`
+  # matches the in-cluster StatefulSet which ships without TLS. Set to `require`
+  # (or stricter) when pointing at a TLS-terminating Postgres like RDS.
+  sslmode: disable
```
ital0, line 103: `postgres.sslmode` only feeds PowerSync's config. Backend `DATABASE_URL` ignores it, so operators setting `require` for RDS will silently still get a plaintext backend connection. Move under `powersync.` or also wire it into backend DSN.

deploy/k8s/values.yaml
```diff
@@ -111,6 +129,12 @@ keycloak:
     clientId: thunderbolt-app
     # -- Base64-encoded client secret (default: "thunderbolt-enterprise-secret")
     clientSecretBase64: dGh1bmRlcmJvbHQtZW50ZXJwcmlzZS1zZWNyZXQ=
+  # -- Include a `demo@thunderbolt.io` / `demo` user in the imported realm.
+  # Convenient for local/preview environments and docs walk-throughs; set to
+  # `false` for production deployments. Default `true` preserves current behavior.
```
ital0, line 161: Keycloak realm import is one-shot. Flipping this to `false` on an existing cluster does not delete the demo user. Worth saying so in the docstring so operators don't assume promotion cleans it up.

deploy/README.md
```diff
@@ -358,6 +358,10 @@ See `deploy/k8s/values.yaml` for all configurable values. Key ones:
 | `ingress.className` | `nginx`            | Ingress class                                |
 | `ingress.host`      | `""`               | Set for production (empty = default rule)    |
 
+### Known caveats
```
ital0, line 361: Useful caveat. Please add the exact PowerSync error signature, scope (15/16 also affected?), and link to an upstream issue. Otherwise the next person hitting an init error can't tell if it's this bug or something else.

deploy/k8s/templates/secrets.yaml
```diff
@@ -16,3 +16,9 @@ data:
   mistral-api-key: {{ .Values.backend.aiSecrets.mistralApiKeyBase64 | default "" | quote }}
   thunderbolt-inference-api-key: {{ .Values.backend.aiSecrets.thunderboltInferenceApiKeyBase64 | default "" | quote }}
   exa-api-key: {{ .Values.backend.aiSecrets.exaApiKeyBase64 | default "" | quote }}
+stringData:
```
ital0, line 19: JWK `k` is RFC 7518 base64url, but `secretBase64` naming suggests standard base64. A `openssl rand -base64` generated secret may contain `+/=` and trip the JWK parser. Worth a warning or rename.

deploy/README.md
ital0, line 361: Consider cross-linking `postgres.sslmode` here. RDS users hit the sslmode requirement first, then this caveat, and the symptoms can look similar.

deploy/k8s/values.yaml
```diff
@@ -16,6 +20,8 @@ frontend:
   replicas: 1
   service:
     port: 80
+  podAnnotations: {}
```
ital0, line 28: Per-component `podAnnotations` and `resources` lack `# --` docstrings while the top-level `podAnnotations` has one. helm-docs will mark them undocumented.

deploy/k8s/templates/backend.yaml
```diff
@@ -25,6 +25,11 @@ spec:
     metadata:
       labels:
         app: backend
+      {{- $podAnnotations := merge (deepCopy (.Values.backend.podAnnotations | default dict)) (.Values.podAnnotations | default dict) }}
+      {{- if $podAnnotations }}
+      annotations:
```
ital0: Suggestion: extract the duplicated merge+annotations block into `_helpers.tpl` so the merging strategy lives in one place. Same goes for the resources block.

deploy/k8s/templates/backend.yaml
```diff
@@ -44,8 +49,15 @@ spec:
               value: {{ .Values.backend.env.waitlistEnabled | quote }}
             - name: DATABASE_DRIVER
               value: {{ .Values.backend.env.databaseDriver }}
+            # POSTGRES_PASSWORD must appear before DATABASE_URL so the
+            # `$(POSTGRES_PASSWORD)` k8s env-var expansion resolves.
+            - name: POSTGRES_PASSWORD
```
ital0: Suggestion: rename to `THUNDERBOLT_POSTGRES_PASSWORD`. `POSTGRES_PASSWORD` is auto-consumed by libpq tools (psql, pgcli, pg_dump), so a future sidecar or `kubectl exec` will silently get superuser access.

deploy/k8s/templates/secrets.yaml
```diff
@@ -16,3 +16,9 @@ data:
...
+  # PowerSync JWKS expects `k` to be base64url-encoded key material. Storing the
+  # already-base64-encoded string here (via stringData, so k8s wraps it once)
```
ital0: Suggestion: tighten the wording. The comment uses `base64` for three things (JWK encoding, user-supplied value, env-injected output), and the key contrast with the `data:` block above isn't loud enough.

deploy/k8s/templates/backend.yaml
```diff
@@ -44,8 +49,15 @@ spec:
               value: {{ .Values.backend.env.waitlistEnabled | quote }}
             - name: DATABASE_DRIVER
               value: {{ .Values.backend.env.databaseDriver }}
+            # POSTGRES_PASSWORD must appear before DATABASE_URL so the
```
ital0: Nit: name the Kubernetes feature explicitly (`dependent environment variables`). `$(VAR)` reads as shell to many people; naming it prevents misreading.

deploy/k8s/values.yaml
```diff
@@ -5,6 +5,10 @@ imagePullSecrets: []
 # -- Base URL the app is served from (used for CORS, redirects, auth callbacks)
 appUrl: http://localhost
 
+# -- Annotations applied to every pod in the chart. Merged with per-component
+# `<component>.podAnnotations` (per-component wins on key conflict).
```
ital0: Nit: precedence is correct, but Sprig `merge` (dst wins) is counterintuitive and easy to misread. Worth a brief callout in the docstring.

deploy/k8s/templates/backend.yaml
```diff
@@ -44,8 +49,15 @@ spec:
...
+            # `$(POSTGRES_PASSWORD)` k8s env-var expansion resolves.
+            - name: POSTGRES_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: thunderbolt-secrets
+                  key: postgres-password
             - name: DATABASE_URL
-              value: "postgresql://{{ .Values.postgres.credentials.user }}:{{ .Values.postgres.credentials.passwordBase64 | b64dec }}@postgres:5432/{{ .Values.postgres.credentials.database }}"
+              value: "postgresql://{{ .Values.postgres.credentials.user }}:$(POSTGRES_PASSWORD)@postgres:5432/{{ .Values.postgres.credentials.database }}"
```
ital0: Nit (pre-existing, optional): document that the decoded postgres password must be URL-safe. URL-reserved chars (`@:/?#`) silently break `DATABASE_URL` here and the same applies to PowerSync's URIs.

ital0 (changes requested):
Everything reviewed here had been caught by Claude.

deploy/k8s/values.yaml
darkbanjo: ⚡ Wired `postgres.sslmode` into the backend `DATABASE_URL` as `?sslmode=...` (templates/backend.yaml). It now flows into both the PowerSync config and the backend DSN from the same value. Docstring updated to reflect this.

deploy/k8s/values.yaml
darkbanjo: ⚡ Good catch — added a note to the docstring that realm import is one-shot, so flipping this to `false` on an existing cluster does NOT remove a previously-imported demo user (operator must delete via Keycloak admin console or destroy storage).

deploy/README.md
darkbanjo: ⚡ Expanded the caveat with the observed signature (`PostgresLockManager.init` hang, no "Power up" log, swallowed pgwire error) and explicit scope (PowerSync 1.20.5 confirmed; 1.15/16 not retested). I don't have an upstream issue link to cite since we surfaced this through MZLA's internal deployment — added a pointer to file one against `powersync-ja/powersync-service` so future hits are searchable.

deploy/k8s/templates/secrets.yaml
darkbanjo: ⚡ Added an explicit WARNING in both `values.yaml` (powersync.jwt.secretBase64) and `secrets.yaml` (powersync-jwt-secret-b64) calling out that JWK `k` is RFC 7518 base64*url*, not standard base64. Kept the field name (renaming would be a breaking values change) and instead documented the safe generation commands: `openssl rand 32 | basenc --base64url --wrap=0` or post-process with `tr '+/' '-_' | tr -d '='`.

deploy/README.md
darkbanjo: ⚡ Done — added a TLS-to-RDS caveat directly above the PowerSync caveat in `deploy/README.md` and a cross-reference inside the PowerSync caveat steering operators to check `postgres.sslmode` first since the failure shapes can look similar.

deploy/k8s/values.yaml
darkbanjo: ⚡ Added `# --` docstrings to every per-component `podAnnotations` and `resources` so helm-docs picks them up.

deploy/k8s/templates/backend.yaml
darkbanjo: ⚡ Extracted both into `_helpers.tpl` (`thunderbolt.podAnnotations` and `thunderbolt.resources`) and rewrote all six workload templates (backend, frontend, marketing, postgres, powersync, keycloak) to use them. The merge precedence (per-component wins) is now documented in exactly one place. Verified with `helm template` + `helm lint`.

deploy/k8s/templates/backend.yaml
darkbanjo: ⚡ Renamed to `THUNDERBOLT_POSTGRES_PASSWORD` in both `templates/backend.yaml` and `templates/powersync.yaml` (and updated the $(VAR) interpolations in DATABASE_URL / PS_STORAGE_URI). Left the bare `POSTGRES_PASSWORD` on the postgres StatefulSet itself since the official postgres image requires that exact env name to bootstrap the cluster.

deploy/k8s/templates/secrets.yaml
darkbanjo: ⚡ Rewrote the comment to make the contrast explicit: `data:` stores the decoded JWT secret; `stringData:` surfaces the base64 string itself so PowerSync's JWK `k` round-trips. Also folded in the base64url warning so the two concerns are colocated.

deploy/k8s/templates/backend.yaml
darkbanjo: ⚡ Updated the comment to name the feature explicitly: "Kubernetes dependent environment variables — earlier entries are interpolated into later ones via \`$(VAR)\`, which is k8s syntax, not the shell."

deploy/k8s/values.yaml
darkbanjo: ⚡ Added a Sprig-merge-is-dst-wins callout to the top-level `podAnnotations` docstring in `values.yaml` and pointed to the helper that implements the merge (`thunderbolt.podAnnotations` in `_helpers.tpl`).

deploy/k8s/templates/backend.yaml
darkbanjo: ⚡ Documented this inline in `templates/backend.yaml` next to `DATABASE_URL` (URL-reserved chars `@:/?#` silently break both DATABASE_URL here and the equivalent PS_STORAGE_URI/PS_PG_URI in `templates/powersync.yaml`).

PR #897, opened 2026-05-26
test: remove process-wide leaks from mock.module and iframe window re…

darkbanjo:
- Replace mock.module('@/lib/auth-mode') in use-auth-gate and logout-modal tests with direct import.meta.env mutation cleaned up in afterEach; module-level mock.module persisted across files under --randomize and contaminated unrelated suites.
- Drop the iframe-based window.addEventListener/removeEventListener/ dispatchEvent rebind in use-credential-events and use-powersync-credentials-invalid-listener tests; the rebind itself was the leak, replacing real APIs globally for the rest of the process.
- Switch config.test from beforeEach(updateConfig({})) to afterEach(setState({ config: {} })) so the store is reset after each test instead of before, removing order-dependent state.

---

> [!NOTE]
> **Low Risk**
> Mostly test and optional DI props; production paths keep default `clearLocalData` and real auth-mode env reads.
> 
> **Overview**
> Hardens the test suite against **process-wide pollution** under `bun test --randomize` by removing leaky setup patterns and aligning with dependency injection in `docs/development/testing.md`.
> 
> **Auth / env:** `use-auth-gate` and `logout-modal` tests drop `mock.module('@/lib/auth-mode')` in favor of per-test `import.meta.env` flags cleared in `afterEach`. SSO coverage in logout tests uses `VITE_AUTH_MODE = 'sso'` instead of a global mock.
> 
> **Cleanup:** `LogoutModal` and `usePowerSyncCredentialsInvalidListener` accept an optional injectable `clearLocalData` (production still uses the real implementation). Tests pass stubs instead of `mock.module('@/lib/cleanup', ...)`.
> 
> **Other test hygiene:** `config.test` resets the config store in `afterEach` via `setState` instead of `beforeEach(updateConfig({}))`. Credential-event tests no longer rebind `window` event APIs from an iframe. `ThunderboltConnector`’s quiet-403 test only asserts the credentials-fetch `console.error` is suppressed, not silence from unrelated PostHog leakage.
> 
> **Production behavior** for logout and credentials-invalid flows is unchanged aside from the optional test hooks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2cd1f467b6beb26ac6ac53932cb9ae57eab547a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


raivieiraadriano92 (dismissed):

src/components/logout-modal.test.tsx
```diff
@@ -14,15 +14,12 @@ import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, mock
 import { LogoutModal } from './logout-modal'
 
 const mockClearLocalData = mock(() => Promise.resolve())
-const mockIsSsoMode = mock(() => false)
 
 mock.module('@/lib/cleanup', () => ({
```
ital0: I see this mock.module here. I know it's originally not part of this PR, but would like to understand if we could follow the testing docs and get rid of this.

ital0 (changes requested):

src/components/logout-modal.test.tsx
ital0: Sorry, I added the comment in the wrong line, but I'm talking about `mock.module` at line 18

src/hooks/use-powersync-credentials-invalid-listener.test.tsx
```diff
@@ -31,20 +31,6 @@ const authToken = 'test-auth-token'
...
-  const apis = {
-    addEventListener: win.addEventListener,
-    removeEventListener: win.removeEventListener,
-    dispatchEvent: win.dispatchEvent,
-  }
-  document.body.removeChild(iframe)
-  return apis
-}
-
 mock.module('@/lib/cleanup', () => ({
```
ital0: what about this one? Is it required?

ital0 (changes requested):

src/hooks/use-powersync-credentials-invalid-listener.test.tsx
darkbanjo: ⚡ Good catch — fixed in `f2cd1f46`. Same DI pattern we applied to `LogoutModal` earlier in this PR: `clearLocalData` is now an optional hook option on `usePowerSyncCredentialsInvalidListener`, defaulting to the real `@/lib/cleanup` import. Production callers don't change; the test passes its mock via the option instead of `mock.module(...)`. The shared-module mock and the leak risk it represented are gone.

PR #903, opened 2026-05-26
feat(THU-534): skills frontend persistence + settings page

darkbanjo:
> [!NOTE]
> **Medium Risk**
> New synced user data path (skills table, encryption, reconcile) plus a large UI surface; logic is well-tested but touches init reconciliation and E2EE config.
> 
> **Overview**
> Adds **Skills** as user-managed slash-command templates: a new **DAL** (`create`/`update`/soft-delete, AgentSkills name validation, pin cap of 10, reorder, legacy `/` name migration), **default seed skills**, E2EE column coverage, and **default reconciliation** on app init.
> 
> Wires **`/settings/skills`** (route, sidebar, layout without duplicate header) and a **master–detail Skills UI**: searchable library with pinned drag-reorder, enable/disable (auto-unpin on disable), create/edit forms, detail view, and **Run in chat** via router state. **Dependents** detection warns before disable/delete when other skills reference `/slug` in text.
> 
> Includes reducer-driven view state, PowerSync/React Query hooks, Storybook stories, DAL/unit/integration tests, and a shared **framer-motion** test mock.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5515c5b296a3e16684dc0cd5abb169ff081682c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/skills/skills-view.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const createForm = (
+    <SkillForm
+      key="create"
+      mode="create"
+      onCancel={() => {
+        requestLeave({ type: 'cancel' })
+        if (isMobile) {
+          setMobileView('list')
+        }
+      }}
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in d04c78a9. Moved the `setMobileView('list')` call into `performLeave` so the panel only slides away when the leave actually commits (i.e. after the user confirms discard, or immediately if the form is clean). If they pick "Keep editing", the panel stays visible.

src/skills/skills-view.tsx
```diff
@@ -0,0 +1,367 @@
...
+    } else {
+      await removeSkill(skill.id)
+    }
+  }
+
+  const onJumpToDependent = (id: string) => {
+    setActiveId(id)
+    setMode('edit')
+    setPendingDependents(null)
+  }
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in d04c78a9. `onJumpToDependent` now calls `setMobileView('panel')` on mobile so the edit form is actually visible after dismissing the dependents dialog.

src/skills/skill-form.tsx
```diff
@@ -0,0 +1,196 @@
...
+              placeholder="meeting-notes"
+              value={name}
+              onChange={(e) => handleNameChange(e.target.value)}
+              className="h-9 pl-7"
+              aria-describedby="skill-name-help"
+              aria-invalid={localNameError || nameError ? true : undefined}
+            />
+          </div>
+          {localNameError || nameError ? (
+            <p className="text-sm text-destructive">{localNameError ?? nameError}</p>
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed by clearing the parent-side error as soon as the user edits the name. Added a CLEAR_NAME_ERROR action wired into a new onNameChange callback on SkillForm, with a reducer unit test covering the drop + no-op-same-ref behavior.

src/skills/skills-view.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+  const removeSkill = useCallback(
+    async (id: string) => {
+      await softDeleteSkill(id)
+      if (pinnedSet.has(id)) {
+        await togglePin(id)
+      }
+    },
+    [softDeleteSkill, pinnedSet, togglePin],
+  )
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Done — dropped the redundant togglePin after softDeleteSkill (the DAL already nulls pinnedOrder in the same write), and extracted a disableSkill helper so handleToggleEnabled and confirmPendingDependents share the same disable+auto-unpin logic instead of duplicating it inline.

src/skills/skill-form.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const [name, setName] = useState(initialName)
+  const [description, setDescription] = useState(initialDescription)
+  const [instruction, setInstruction] = useState(initialInstruction)
+
+  // Surface AgentSkills-spec violations inline as soon as the user has typed
+  // something, but don't shout at an empty initial state.
+  const localNameError = name.trim() === '' ? null : validateSkillName(name)
+  const canSubmit =
+    name.trim() !== '' && description.trim() !== '' && instruction.trim() !== '' && localNameError === null
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Fixed — validateSkillName now runs against name.trim(), matching what handleSubmit actually submits.

src/skills/skill-form.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const [instruction, setInstruction] = useState(initialInstruction)
+
+  // Surface AgentSkills-spec violations inline as soon as the user has typed
+  // something, but don't shout at an empty initial state. Validate against the
+  // trimmed value — `handleSubmit` submits the trimmed name, so the two must
+  // agree (otherwise " meeting-notes " reads as invalid even though it isn't).
+  const trimmedName = name.trim()
+  const localNameError = trimmedName === '' ? null : validateSkillName(trimmedName)
+  const canSubmit =
+    trimmedName !== '' && description.trim() !== '' && instruction.trim() !== '' && localNameError === null
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in 6643d358. Added `!nameError` to the `canSubmit` predicate so a server-side name error (e.g. `SkillNameTakenError`) disables submit until the user edits the name. `handleNameChange` already clears `nameError` on the next keystroke, so the button re-enables naturally.

src/settings/layout.tsx
```diff
@@ -4,9 +4,18 @@
...
-import { Outlet } from 'react-router'
+import { Outlet, useLocation } from 'react-router'
+
+// Sub-routes that provide their own page chrome (heading + actions + mobile
+// sidebar trigger inside their own component) and want the full content height
+// available. The settings-level Header would otherwise add ~56px of unused
+// space at the top.
+const routesWithOwnHeader = new Set(['/settings/skills'])
 
 export default function SettingsLayout() {
```
ital0: it's not part of your changes, but could please migrate this to arrow function? Since this PR is touching this file.

src/skills/skills-view.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+  const panel =
+    mode === 'create' ? (
+      createForm
+    ) : !active ? (
+      emptyPanel
+    ) : mode === 'detail' ? (
+      <SkillDetail
+        name={active.name ?? ''}
```
ital0: the `Skill` type is declared as `WithRequired<..., 'name' | 'description' | 'instruction' | 'enabled'>` which promises these fields are non-null, and since every DAL query filters `isNull(deletedAt)` this matches reality. But there are quite a few consumers doing `target.name ?? ''` (here on lines 219-221 and 236-238, plus `skills-view.tsx:91`, `skills-view.tsx:138`, `skills-view.tsx:296`, `skills-view.tsx:314`, and `skills-list.tsx:53`). It's living in both worlds: the type guarantees it and the code still defends. I think it's worth picking one side. Dropping the `?? ''` and trusting the type feels more aligned with the "prefer optimistic code over defensive code" rule in CLAUDE.md.

src/dal/skills.ts
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * One-shot migration: strip the leading `/` from any non-deleted skill row
+ * whose `name` still carries the legacy prefix. Skills migrated to bare slugs
+ * per the AgentSkills spec mid-PR; default rows the reconciler hasn't yet
+ * rewritten (or rows the user has edited, which the reconciler treats as
+ * user-modified and never touches) need this fix.
+ *
+ * Idempotent — running it twice is a no-op once no rows match.
+ *
+ * TODO: drop this once the legacy population has churned through one app
```
ital0, line 272: The TODO says to remove this ~30 days post-release, but TODOs like this tend to become fossils. Would you mind opening a Linear ticket with a scheduled date instead? That way the reminder has an owner and a deadline, and we can keep this codepath from quietly outliving its usefulness.

ital0 (changes requested):

src/skills/use-skills.ts
```diff
@@ -0,0 +1,155 @@
...
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+import { useMemo } from 'react'
+
+// `@powersync/tanstack-react-query` auto-invalidates this query whenever the
+// `skills` table fires a `tablesUpdated` event (Drizzle writes through
+// PowerSync's `writeLock`, which routes through SQLite's commit hook). That
+// path *does* fire — but PowerSync wraps the listener in a 30ms
+// `throttleTrailing` (`DEFAULT_WATCH_THROTTLE_MS` in `@powersync/common`),
+// so the row a mutation just wrote can miss the very next render. Explicit
+// invalidation on mutation success cuts that to zero latency.
```
raivieiraadriano92, line 30: Question, we are no longer using manual invalidation on other screens - do we really want or need to do it for skills?

src/skills/skills-view-state.ts
```diff
@@ -0,0 +1,190 @@
...
+      return {
+        ...state,
+        activeId: action.id,
+        mode: 'edit',
+        pendingDependents: null,
+        // The dependents dialog can be opened from a list-row action while
+        // mobileView is still 'list'; sliding the panel in here gives the
+        // edit form a surface to render on.
+        mobileView: action.isMobile ? 'panel' : state.mobileView,
+      }
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in 5515c5b2. Both `OPEN_DEPENDENTS` and `JUMP_TO_DEPENDENT` now clear `isDirty` and `nameError` so a stale dirty flag from the prior edit session can't trigger a spurious discard-changes dialog on the new, untouched form.

src/settings/layout.tsx
darkbanjo: ⚡ Good call — done in 5515c5b2.

src/skills/skills-view.tsx
darkbanjo: ⚡ Agreed — dropped the `?? ''` fallbacks in 5515c5b2 across `skills-view.tsx` (6 sites) and `skills-list.tsx` (1 site). The `Skill` type's `WithRequired` guarantee now stands on its own. The only remaining `?? ''` is in `skill-form.tsx` where `initialValues` is genuinely optional (create vs. edit).

src/dal/skills.ts
darkbanjo: ⚡ Fair point on TODO-rot. I'll file a Linear ticket for the cleanup so it has a real owner and deadline — keeping the TODO comment as the in-code breadcrumb that points back to it.

src/skills/use-skills.ts
darkbanjo: ⚡ Reasonable question. The block comment above `skillsQueryKey` lays out the reasoning: `@powersync/tanstack-react-query` auto-invalidates on `tablesUpdated`, but PowerSync wraps that listener in a 30ms `throttleTrailing` (`DEFAULT_WATCH_THROTTLE_MS` in `@powersync/common`), so the row a mutation just wrote can miss the very next render. Explicit `onSuccess` invalidation closes that 0–30ms gap. If other screens have hit the same race and quietly worked around it some other way I'd love to hear — happy to revisit if there's a cleaner pattern in the codebase.

src/defaults/skills.ts
```diff
@@ -0,0 +1,97 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
ital0, line 1: suggestion: I would run some rounds of evaluations on these prompts just to check its consistency and if the results are good enough. Not a blocker.

darkbanjo: oh great call, thank you

ital0 (approved):

PR #905, opened 2026-05-26
feat(acp): add agent client protocol with WebSocket transport

ital0:
> [!NOTE]
> **High Risk**
> Introduces a new WebSocket auth path, external Deepset API proxying with server-held API keys, and DB/schema changes tied to chat threads—areas that affect security, reliability, and multi-turn chat behavior.
> 
> **Overview**
> Adds **Agent Client Protocol (ACP)** support so deployments can expose managed and custom agents over WebSocket, with **Deepset Haystack** as the first managed provider.
> 
> **Discovery and config:** New authenticated `GET /agents` aggregates registered providers (Haystack registers on route mount), honors `ENABLED_AGENTS` / `ALLOW_CUSTOM_AGENTS`, and blocks anonymous users. Public `GET /config` now exposes `builtInAgentEnabled` and `allowCustomAgents`. Haystack is configured via `HAYSTACK_*` env vars and optional pipeline JSON in `.env.example`.
> 
> **Haystack adapter:** `WS /v1/haystack/ws?pipeline=…` runs a JSON-RPC ACP server that bootstraps Deepset `search_sessions`, streams `chat-stream` SSE into `session/update`, supports `session/load` resume, cold-pipeline retries on specific 503 bodies, and turn-scoped cancel. Authenticated `GET /v1/haystack/files/:fileId` proxies Deepset file downloads. WebSocket auth uses a **bearer token in `Sec-WebSocket-Protocol`** (validated in `open()` like REST), not URL query params or cookies.
> 
> **Data model:** Migration adds `powersync.agents` for user-defined agents and `acp_session_id` / `agent_id` on `chat_threads`; Drizzle/PowerSync schemas updated accordingly.
> 
> **Dependencies:** `@agentclientprotocol/sdk` added to the backend.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df7b98ecd0dcd9adabeef7c4fd00eb51774eab8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


package.json
```diff
@@ -107,12 +108,14 @@
     "framer-motion": "^12.23.12",
     "input-otp": "^1.4.2",
     "lucide-react": "^1.8.0",
+    "mammoth": "^1.12.0",
```
cjroth, line 117: Just wanted to mention here- this is for docx parsing, I believe, which is needed for Deepset only. I know I originally was the one that added this to the Deepset POC. I'm not sure that this should be part of this PR. Tbh, I actually don't even feel like this should be part of our core codebase.

I hate that we're adding `react-pdf` too, but at least pdfs are a pretty common data format. docx is somewhat common but is kind of a microsoft-specific format that is becoming less popular over time. I'd love to find a way to abstract this file preview stuff so that we can just support ONLY the necessary file types for each deployment if that's possible.

I don't want this to be a blocker of getting ACP merged, but just something to flag. I think the two big concerns with this stuff are (1) every npm package introduced some supply-chain / vendor risk and (2) it just increases bundle size and bloat.

src/widgets/document-result/pdf-sidebar-viewer.tsx
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Renders an inline preview of a Haystack-managed document in the sideview slot.
+ * Loads the file as a blob via the backend, renders PDFs with react-pdf, and
+ * DOCX content via mammoth into a sandboxed iframe. Falls back to a download
+ * prompt for other extensions.
+ */
+export const PdfSidebarViewer = ({ fileId, fileName, initialPage }: PdfSidebarViewerProps) => {
+  const { close } = useContentView()
+  const httpClient = useHttpClient()
+  const [blobUrl, setBlobUrl] = useState<string | null>(null)
```
cjroth: This is a lot of useStates - would it make sense to organize this into a useReducer? Generally I find that that makes it less bug-prone

src/widgets/document-result/pdf-sidebar-viewer.tsx
```diff
@@ -0,0 +1,191 @@
...
+      const blob = await response.blob()
+      if (cancelled) {
+        return
+      }
+
+      const url = URL.createObjectURL(blob)
+      blobUrlRef.current = url
+
+      if (fileType === 'docx') {
+        const mammoth = await import('mammoth')
```
cjroth: I'm relieved that this is a lazy import :)

src/widgets/document-result/pdf-sidebar-viewer.tsx
```diff
@@ -0,0 +1,191 @@
...
+    return () => {
+      cancelled = true
+      if (blobUrlRef.current) {
+        URL.revokeObjectURL(blobUrlRef.current)
+        blobUrlRef.current = null
+      }
+    }
+  }, [fileId, fileType, httpClient])
+
+  const handleDownload = useCallback(() => {
```
cjroth, line 53: I wonder if we should abstract this into a global utility - I feel like there will be a lot of places where we let the user download a file in the future (although that might be premature abstraction?)

backend/src/agents/routes.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Mounts `GET /agents`, the ACP discovery endpoint.
+ *
+ * - Unauthenticated → 401 `{ error: 'Unauthorized' }`
+ * - Anonymous user → 403 `ANONYMOUS_DISCOVERY_FORBIDDEN` (anonymous sessions
+ *   never see system agents; the FE falls back to the built-in only)
+ * - Authenticated regular user → `AgentDiscoveryResponse`
+ *
+ * The agent list is built from {@link getRegisteredProviders}; the Haystack
+ * module registers its provider into the same registry. `ENABLED_AGENTS` (comma-separated)
```
cjroth, line 24: Just noting here that the `ENABLED_AGENTS` is a temporary thing imo - once we have workspaces, the correct way to set up backend agents will be for a workspace admin to create them at the workspace level.

PR #913, opened 2026-05-28
feat(THU-536): skills v1 telemetry + edge polish (PR 4/5)

darkbanjo:
Fourth of five PRs implementing Skills v1 ([spec](https://github.com/thunderbird/thunderbolt-spec/blob/spec/skillsv1/specs/skills-v1.md)). Adds the seven \`skill_*\` telemetry events called out in the spec, plus a polish pass on the edge cases that the PR-3 review surfaced.

> **Stacked on [#906](https://github.com/thunderbird/thunderbolt/pull/906)** (THU-535). Base will be re-targeted to \`main\` once 906 merges.

## Telemetry — spec §6

| Event | Properties | Fires from |
|---|---|---|
| \`skill_used\` | \`skill_id\`, \`via\` ∈ \`{slash, chip, settings-nav}\` | chat composer — slash popup pick, pinned-chip click, settings "+ Add to chat" |
| \`skill_pinned\` | \`skill_id\` | chat skills bar — \`+\` popover pick |
| \`skill_unpinned\` | \`skill_id\` | chat skills bar — chip menu "Unpin" |
| \`skill_reordered\` | \`skill_id\`, \`from_index\`, \`to_index\` | chat skills bar — reorder commit (single-move attribution via \`findMovedItem\`) |
| \`skill_created\` | \`skill_id\`, \`instruction_length\` | \`SkillsView\` create branch |
| \`skill_edited\` | \`skill_id\`, \`renamed\` (bool) | \`SkillsView\` edit branch |
| \`skill_deleted\` | \`skill_id\` | \`SkillsView\` \`removeSkill\` wrapper |

**Privacy:** \`skill_id\` is \`sha256(user_id + ':' + skill.id)\` truncated to 16 hex chars. The user's authored \`name\` / \`description\` / \`instruction\` never leaves the device.

\`useSkillTelemetry\` is a hook so it can read \`AuthContext\` for the \`user_id\` half of the hash; it tolerates a missing provider (tests / Storybook) by no-op'ing the tracker.

## Edge polish

- **Pin-cap UX** — \`+\` popover trigger now disables when \`pinned >= maxPinnedSkills\` even if pinnable candidates remain, with tooltip "Pin limit reached (10). Unpin one first." Previously the popover would let the user click the 11th candidate, the DAL would throw \`PinLimitExceededError\`, and the catch in the click handler swallowed it silently.
- **Empty-state copy** — \`/settings/skills\` now distinguishes "no skills yet" (with a "Create your first skill" CTA button) from "library has skills, none selected" (idle).
- **Token-budget integration** — two new tests in \`use-context-tracking.test.tsx\` pin down the \`additionalInputTokens\` overflow path: \`isOverflowing\` flips true when (used + user-input + resolved-skill-tokens) exceeds the model's context window, false otherwise.
- **Per-mode availability** — comment-only note in \`aiFetchStreamingResponse\` documenting Skills v1 OQ6 ("skills are mode-agnostic; a skill is text injection, not a tool"), so future readers don't accidentally introduce per-mode gating.
- **Dependents-aware delete** — already covered by tests inherited from THU-534 (\`find-dependents.test.ts\` + \`skills-view-state.test.ts\`); verified, no change.

## What's not in this PR

- The PostHog \`posthogClient\` singleton is still module-level. Hashing is unaffected, but the underlying singleton leak surfaces under \`bun test --randomize\` and is tracked as a separate ticket.
- The "skills disappearing in prod" bug Chris hit was a missing PowerSync Cloud dashboard sync rule — ops fix, no code change.

## Test plan

- [x] \`bun test --cwd=src --randomize\` — 2281 pass, 1 pre-existing skip, 0 fail
- [x] \`bun run type-check\` — clean
- [ ] Manual: each of the 7 events fires in the right spot, payload contains only the hashed id + declared extras, no skill content
- [ ] Manual: try to pin an 11th skill — trigger disabled, tooltip explains
- [ ] Manual: empty-state CTA on \`/settings/skills\` lands in the create form

---

> [!NOTE]
> **Low Risk**
> Changes are analytics side effects, UI guards, and tests; no auth, sync, or inference behavior changes beyond comments.
> 
> **Overview**
> Adds **Skills v1 §6 analytics**: a `useSkillTelemetry` hook that emits seven `skill_*` PostHog events with a **hashed** `skill_id` (no skill text leaves the client), wired from the chat composer, pinned-skills bar, and settings **Skills** view. Pin/unpin/reorder telemetry runs **after** DAL mutations succeed so failed actions are not logged.
> 
> **UX polish:** the pin **+** control disables at `maxPinnedSkills` with clearer tooltip copy; reorder reports **from/to** indices from dnd-kit; settings empty library gets a **Create your first skill** CTA. Tests cover pin-cap UI and `additionalInputTokens` driving context overflow. `ai/fetch` gains a comment that skills apply in all chat modes (no per-mode gate).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45d82649fdde42af9733dde99c64d5f989ee19fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/skills/skills-view.tsx
```diff
@@ -191,18 +204,30 @@ export const SkillsView = () => {
...
+          <Plus />
+          Create your first skill
+        </Button>
+      </section>
+    ) : (
+      <section className="flex h-full flex-1 flex-col items-center justify-center gap-2 bg-background px-6 text-center text-foreground">
+        <h2 className="text-xl text-muted-foreground">Select a skill</h2>
+        <p className="max-w-md text-sm text-muted-foreground">Pick a skill from the list to view or edit it.</p>
+      </section>
+    )
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in `96f9444c`. You're right that `active` always falls back to `skills.at(0)` so the `skills.length > 0` branch was unreachable. Collapsed `emptyPanel` to just the genuinely-empty case ("No skills yet" + the create CTA); when the library has rows, the panel renders the auto-selected first skill's detail as before. Comment updated to spell out why the branch is single-case.

src/hooks/use-context-tracking.test.tsx
```diff
@@ -91,4 +111,82 @@ describe('useContextTracking reactivity', () => {
...
+    await createModel(db, {
+      id: modelId,
+      provider: 'openai',
+      name: 'Tiny Model',
+      model: 'gpt-4',
+      isSystem: 0,
+      enabled: 1,
+      contextWindow: 100,
+    })
+    const model = await import('@/dal').then((m) => m.getModel(db, modelId))
```
ital0: `@/dal` is already imported at the top of this file, so we can pull `getModel` into that existing import instead of reaching for a dynamic `await import('@/dal')` here.

src/skills/find-moved-item.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Used by `skill_reordered` telemetry: the chat composer's reorder
+ * panel commits an entire `ids[]` array via `reorderPins`, but the
+ * telemetry payload wants the moved id alone plus its indices.
+ */
+export const findMovedItem = (oldIds: readonly string[], newIds: readonly string[]): Move | null => {
+  if (oldIds.length !== newIds.length) {
+    return null
+  }
+  let firstDiff = -1
```
ital0: this loop just finds the first index where the two arrays diverge, which `findIndex` says in one line and lets us drop the `let`. Something like `const firstDiff = oldIds.findIndex((id, i) => id !== newIds[i])`, and the not found check right below stays exactly the same. Totally optional, it just reads a little cleaner.

src/skills/telemetry.ts
```diff
@@ -0,0 +1,100 @@
...
+    <E extends SkillEventName>(event: E, skillId: string, extras: SkillEventProps[E]) => {
+      if (!enabled) {
+        return
+      }
+      void hashSkillId(userId, skillId).then((hashedId) => {
+        // Cast through `unknown` because PostHog's `trackEvent` signature
+        // takes a generic `Record<string, unknown>` — our per-event prop
+        // shape is the stricter type-checked contract for *callers* of
+        // `useSkillTelemetry`, but PostHog itself doesn't know about it.
+        trackEvent(event, { skill_id: hashedId, ...extras } as unknown as Record<string, unknown>)
```
ital0: the double hop through `unknown` is a touch heavy. Every value in `SkillEventProps` is already an object, so a single `as Record<string, unknown>` on the spread often satisfies the compiler on its own. Worth a quick try to see if the `unknown` bridge is really needed.

ital0 (changes requested):

src/hooks/use-context-tracking.test.tsx
darkbanjo: Done — pulled `getModel` into the top-level `@/dal` import and dropped the dynamic `await import` here and at the two other matching call sites in this file (commit `7b9071cb`).

src/skills/find-moved-item.ts
darkbanjo: Done in `7b9071cb` — collapsed to `const firstDiff = oldIds.findIndex((id, i) => id !== newIds[i])`. Reads cleaner and drops the `let`.

src/skills/telemetry.ts
darkbanjo: You were right — `as Record<string, unknown>` alone satisfies the compiler. Dropped the intermediate `unknown` and tightened the comment (commit `7b9071cb`).

src/skills/telemetry.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (!enabled) {
+        return
+      }
+      void hashSkillId(userId, skillId).then((hashedId) => {
+        // PostHog's `trackEvent` accepts a generic `Record<string, unknown>`;
+        // our per-event prop shape is the stricter type-checked contract for
+        // *callers* of `useSkillTelemetry`. The widening cast is just to
+        // satisfy that boundary.
+        trackEvent(event, { skill_id: hashedId, ...extras } as Record<string, unknown>)
+      })
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — added a .catch on the hashSkillId promise that logs via console.warn, so crypto.subtle.digest rejections (insecure contexts) no longer surface as unhandled rejections.

ital0 (dismissed):

src/components/chat/chat-skills-bar.tsx
```diff
@@ -135,6 +162,7 @@ export const ChatSkillsBar = ({
                   <button
                     type="button"
                     onClick={() => {
+                      trackSkillEvent('skill_pinned', skill.id, {})
```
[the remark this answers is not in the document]
darkbanjo: Fixed in `45d82649`. All three handlers (`onReorder`, `onUnpin`, the pin-popover `onClick`) are now async and `await` the underlying mutation before firing `trackSkillEvent`, so a rejected mutation (e.g. a race past the pin-cap guard) no longer records a phantom event. Failures are caught and logged via `console.warn` to match the existing pattern.

The pin popover's `setAddOpen(false)` was moved before the await so the popover doesn't sit open while the mutation lands.

src/skills/find-moved-item.ts
```diff
@@ -0,0 +1,83 @@
...
+  const forwardId = oldIds[firstDiff]
+  if (forwardId !== undefined) {
+    const forwardTo = newIds.indexOf(forwardId)
+    if (forwardTo !== -1) {
+      const reconstructed = arrayMove(oldIds, firstDiff, forwardTo)
+      if (sequenceEquals(reconstructed, newIds)) {
+        return { id: forwardId, from: firstDiff, to: forwardTo }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
darkbanjo: This is already resolved by `2515d848` (landed earlier in this branch). The diff-based heuristic was replaced — `findMovedItem` is gone, and `ReorderPanel`'s `onReorder` callback now receives an explicit `move` object derived from dnd-kit's `active.id` plus the index lookup. The adjacent-swap ambiguity Bugbot flagged is unreachable in the new path: we no longer infer the moved id by diffing the arrays.

ital0 (approved):

PR #915, opened 2026-05-29
feat(THU-545): update models

arienemaiara:
## Summary

  Refreshes the Thunderbolt-provided default model lineup and wires a backend proxy for confidential inference via Tinfoil.

  - **Lineup:** swap Sonnet 4.5 → **Opus 4.8** (in-place, same row id), drop Mistral Medium 3.1, add three confidential models served by Tinfoil — **DeepSeek V4 Pro**, **Kimi K2.6**, **GLM 5.1**.
  - **Backend Tinfoil proxy:** new authenticated `/v1/tinfoil/*` route that forwards HPKE-encrypted bodies to the Tinfoil enclave with the server bearer key injected. The browser never sees the key, and bodies stay encrypted end-to-end.
  - **Reconciliation cleanup:** removed defaults are soft-deleted automatically — no hardcoded legacy lists to maintain.

  ## Default models (frontend)

  - `src/defaults/models.ts` — `defaultModelOpus48` reuses Sonnet 4.5's row id (`019af08a-c27b-7074-8aac-95315d1ef3fd`) so unmodified rows upgrade in place via reconciliation; edited rows stick. New `defaultModelDeepseekV4Pro`, `defaultModelKimiK26`, `defaultModelGlm51` — all `provider: 'tinfoil'`, `isSystem: 1`, `isConfidential: 1`.
  - `src/defaults/model-profiles/` — `sonnet.ts` → `opus.ts`; per-vendor files `deepseek.ts`, `kimi.ts`, `glm.ts`; `mistral.ts` deleted.
  - `src/ai/eval/scenarios.ts` — eval models reduced to `gpt-oss` + `opus`.
  - `src/defaults/model-profiles/` — `sonnet.ts` → `opus.ts`; per-vendor files `deepseek.ts`, `kimi.ts`, `glm.ts`; `mistral.ts` deleted.
  - `src/ai/eval/scenarios.ts` — eval models reduced to `gpt-oss` + `opus`.
  - `src/defaults/model-profiles/` — `sonnet.ts` → `opus.ts`; per-vendor files `deepseek.ts`, `kimi.ts`, `glm.ts`; `mistral.ts` deleted.
  - `src/ai/eval/scenarios.ts` — eval models reduced to `gpt-oss` + `opus`.

  ## Reconciliation

  - `src/lib/reconcile-defaults.ts` — `cleanupRemovedDefaults` now scans all `isSystem=1` rows on boot and soft-deletes any row whose id is no longer in `defaultModels` AND whose contents still hash to its stored `defaultHash` (edited rows survive). Same logic for `defaultModelProfiles`. No more hardcoded retired-model list.

  ## Backend

  - `backend/src/tinfoil/routes.ts` (new) — Elysia route group at `/tinfoil`, auth-guarded, `parse: 'none'`, derives the upstream path from `ctx.params['*']` so it survives mount-prefix changes. Strips client `Authorization`, injects `Bearer ${TINFOIL_API_KEY}`, forwards opaque body with `duplex: 'half'` and `decompress: false` so the HPKE bytes stay intact. 503 when `TINFOIL_API_KEY` is unset, 405 for disallowed methods, 401 unauthenticated.
  - `backend/src/inference/routes.ts` — added `'opus-4.8'` → `claude-opus-4-8` to `supportedModels`.
  - `backend/src/config/settings.ts`, `backend/.env.example` — new `TINFOIL_API_KEY` and `TINFOIL_ENCLAVE_URL` (defaults to `https://inference.tinfoil.sh/v1`).
  - `backend/src/db/powersync-schema.ts` — added `'tinfoil'` to the `models.provider` enum to match the frontend.
  - `backend/src/index.ts` — mounts `createTinfoilRoutes` alongside the universal proxy.

  ## Frontend (Tinfoil wiring)

  - `src/ai/fetch.ts` — `getSystemTinfoilClient()` constructs `SecureClient` pointing at `${cloudUrl}/tinfoil`. In the `tinfoil` case of `createModel`, the system path
  wraps `SecureClient.fetch` to inject the real Thunderbolt session token (Bearer) or include cookies (SSO) so the backend's auth guard accepts the encrypted request.
  `getTinfoilClient()` keeps the BYOK direct-to-enclave flow for user-added models.
  - `src/components/ui/model-selector/model-selector.tsx` — `needsApiKey` returns `false` for system Tinfoil rows (key lives server-side).

### Evidences
image
image
image

---

> [!NOTE]
> **High Risk**
> Adds auth-gated inference proxying with server-side API key injection and changes default model data migration behavior; mistakes could affect auth, billing, or user-visible model lists.
> 
> **Overview**
> Refreshes the **default model lineup** and adds **server-mediated Tinfoil** for confidential inference.
> 
> **Models:** Sonnet 4.5 becomes **Opus 4.8** on the same row id (in-place upgrade for unmodified rows); Mistral Medium 3.1 is removed. Three new system models use `provider: tinfoil` — DeepSeek V4 Pro, Kimi K2.6, GLM 5.1 — marked confidential. Backend inference adds `opus-4.8` → `claude-opus-4-8`. Eval scenarios now use gpt-oss + opus only.
> 
> **Tinfoil proxy:** New authenticated `/v1/tinfoil/*` route forwards opaque HPKE bodies to the enclave, strips client `Authorization`, injects `TINFOIL_API_KEY`, and uses `parse: 'none'` plus `decompress: false` so bytes stay encrypted. Frontend system Tinfoil clients target `${cloudUrl}/tinfoil` and wrap fetch with the real session token; BYOK user models still talk to the enclave directly. Model picker treats system Tinfoil rows as not needing an API key.
> 
> **Reconciliation:** `cleanupRemovedDefaults` soft-deletes retired system models/profiles whose content still matches `defaultHash`, while edited rows and profiles tied to surviving parent models are kept.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 329adce0984ae8d620d7e341d5292169f80f97e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/ai/fetch.ts
```diff
@@ -184,6 +202,20 @@ export const createModel = async (modelConfig: Model, getProxyFetch: () => Fetch
       return openrouter(modelConfig.model)
     }
     case 'tinfoil': {
+      // System Tinfoil models proxy through Thunderbolt's backend; the bearer
```
cjroth, line 204: I think it would make sense to move this logic into the `thunderbolt` part of the switch statement (circa line 144) since this is technically a Thunderbolt model even though it's ultimately getting served through Tinfoil. I'd leave this section exclusively for user-added models.

src/lib/reconcile-defaults.ts
```diff
@@ -88,8 +88,47 @@ export const reconcileDefaultsForTable = async <T extends { defaultHash: string
...
 }
 
+/**
+ * Soft-delete legacy system defaults whose hash still matches the original.
+ * Edited rows survive. Compute hashes with `hashValues([...])` against the
+ * legacy field order when adding entries.
+ */
+const removedDefaults: ReadonlyArray<{ modelId: string; modelHash: string; profileHash: string }> = [
+  {
+    // Mistral Medium 3.1 (removed in THU-545)
```
cjroth: * Should Sonnet 4.5 and gpt-oss be included here too?
* Over time, this list will get really long - kind of hate to see this grow over time and bloat the codebase - is there any way to do this without having to bundle a list of every model that's ever been included as a system model here? Perhaps we could use a different approach where it just removes any unedited system model that's not in the current list of defaults without comparing to removed ones?

backend/src/tinfoil/routes.test.ts
```diff
@@ -0,0 +1,223 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { ConsoleSpies } from '@/test-utils/console-spies'
+import { setupConsoleSpy } from '@/test-utils/console-spies'
+import { mockAuth, mockAuthUnauthenticated } from '@/test-utils/mock-auth'
+import { afterAll, beforeAll, beforeEach, describe, expect, it, mock } from 'bun:test'
+import { Elysia } from 'elysia'
+import { createTinfoilRoutes } from './routes'
+
+const ENCLAVE_URL = 'https://inference.tinfoil.sh'
```
cjroth: Note on code styling - these should be camel case instead of all-caps to remain consistent with the rest of the codebase

backend/src/inference/routes.ts
```diff
@@ -43,6 +43,10 @@ export const supportedModels: Record<string, ModelConfig> = {
     provider: 'anthropic',
     internalName: 'claude-sonnet-4-5',
   },
+  'opus-4.8': {
```
cjroth, line 46: Should the other 3 models be added here too?

backend/src/tinfoil/routes.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export const createTinfoilRoutes = (options: CreateTinfoilRoutesOptions) => {
+  const { auth, rateLimit } = options
+  const fetchFn = options.fetchFn ?? globalThis.fetch
+  const settings = getSettings()
+  const apiKey = options.apiKey ?? settings.tinfoilApiKey
+  const enclaveUrl = (options.enclaveUrl ?? settings.tinfoilEnclaveUrl).replace(/\/$/, '')
+
+  return new Elysia({ prefix: '/tinfoil' })
```
cjroth: This is an interesting approach - I guess we're basically just proxying through our backend here, so I get why it's a separate route than the other `/chat` endpoint. However, the other chat endpoint is ultimately where we would hook into for things like token usage tracking. I'm thinking of Tinfoil as another model provider that we can add to the existing code paths - not a separate backend integration - so I'm wondering if we could somehow modify our /chat endpoint to support the encrypted HTTP bodies that Tinfoil uses. I could imagine that maybe not making sense, which is fine, but I have a feeling it could work. If possible it would be preferable to go that route so that there is one single chat endpoint that serves all providers - even Tinfoil - and where we can centralize hooks like token tracking in the future.

cjroth (dismissed):

src/lib/reconcile-defaults.ts
arienemaiara: Done

backend/src/tinfoil/routes.ts
```diff
@@ -0,0 +1,106 @@
...
+    .onError(safeErrorHandler)
+    .use(createAuthMacro(auth))
+    .guard({ auth: true }, (g) => {
+      const chain = rateLimit ? g.use(rateLimit) : g
```
ital0: This is what's breaking the `backend` check 🙂 the ternary makes `chain` a union of two Elysia types, so `.all()` isn't callable (TS2349) and `ctx`/`value`/`key` fall back to `any`.

src/components/ui/model-selector/model-selector.tsx
```diff
@@ -30,10 +30,18 @@ type ModelItemData = {
 /**
  * Models that require an API key but don't have one yet need configuration.
  * Thunderbolt is server-authenticated; custom (OpenAI-compatible) endpoints treat
- * the key as optional, so neither flags as missing.
+ * the key as optional; system Tinfoil rows are also server-authenticated (the key
+ * is injected by the backend proxy) — none of those flag as missing.
  */
-export const needsApiKey = (model: Model) =>
-  model.provider !== 'thunderbolt' && model.provider !== 'custom' && !model.apiKey
+export const needsApiKey = (model: Model) => {
```
ital0, line 40: This now has three branches but `model-selector.test.ts` only covers `categorizeModels`. Per `docs/development/testing.md`, could we add cases for system tinfoil (`false`), non system tinfoil with/without key, and another provider without a key (`true`)?

ital0 (changes requested):

backend/src/tinfoil/routes.ts
arienemaiara: ohh I thought it was a problem on the main because I had other PRs breaking the backend too
thanks for the heads up!

ital0 (approved):

PR #924, opened 2026-05-29
fix(backend): stabilize WS test flake (budgets + open() race)

darkbanjo:
## Summary

Two distinct code fixes targeting the three high-frequency backend test failures observed across 11 recent backend CI runs.

### 1. `backend/src/proxy/ws-e2e.test.ts` — bump WS polling budgets

Two tests had time budgets that work locally but reliably miss on CI:

- **`relays text messages bidirectionally`** (55% fail rate): replaced the 100ms hard `setTimeout` with a polling loop (60 × 50ms = 3s ceiling). Exits early on success, tolerates slow CI round-trips.
- **`upstream close code propagates through the relay`** (82% fail rate): bumped polling from 50 × 50ms = 2500ms to 90 × 50ms = 4500ms. Failures landed at 2587ms — the close event WAS arriving, just past budget. 4500ms leaves a 500ms cushion under the framework's 5s `--timeout`.

### 2. `backend/src/haystack/routes.ts` — fix `open()` / `message()` race

The `opens the socket and answers initialize when a valid bearer is offered` test (45% fail rate) was failing with "socket closed before message" — the server was silently dropping the client's first frame.

**Root cause:** `async open(ws)` awaits `authorizeWsBearer(...)` (DB hit). The client's `onopen` fires as soon as the WS handshake completes, so it can `send('initialize')` while the server's open handler is still mid-await. Bun delivers the message immediately, but `__haystackServer` on `ws.data` is still undefined, so the message handler was dropping the frame on the comment `Auth already rejected this socket; drop quietly` — *incorrect* when the auth check is still in progress.

**Fix:** introduce a per-socket pending queue (`__pendingMessages`) set synchronously *before* any await in `open()`. The message handler checks the queue when the server isn't ready, draining it in arrival order once `open()` finishes setup. `null` signals "auth rejected" or "torn down" → drop quietly.

## Out of scope

A separate failure pattern — auth-anonymous test cascade (PGlite singleton state degrading across `--rerun-each 5`) — isn't touched here. Expect the WS fixes alone may stabilize CI enough; will follow up if needed.

## Test plan

- [x] `bunx tsc --noEmit` clean
- [x] `bun test src/proxy/ws-e2e.test.ts src/haystack/routes.test.ts --timeout 5000 --randomize --rerun-each 5` → 85 pass / 0 fail (full 5x rerun mirroring CI invocation)
- [ ] CI on this PR: backend job passes consistently across reruns

---

> [!NOTE]
> **Medium Risk**
> Haystack WS message handling changes real connection behavior during auth/setup; incorrect queue drain could affect first-frame protocol handling, though scope is limited to the Haystack upgrade path.
> 
> **Overview**
> Stabilizes flaky backend WebSocket behavior in production and in CI tests.
> 
> **Haystack `/haystack/ws`:** Adds a per-socket `__pendingMessages` queue set synchronously at the start of `open()`, before `authorizeWsBearer` and other awaits. Frames that arrive while setup is still running are queued and drained in order once `HaystackAcpServer` is ready, instead of being dropped when `__haystackServer` is still undefined. Failed auth or invalid pipeline clears the queue (`null`) so the message handler still drops quietly after close.
> 
> **Proxy WS e2e:** Replaces a fixed 100ms wait on the bidirectional echo test with polling (up to ~3s). Extends the upstream-close propagation poll from 50 to 90 iterations (~4.5s) so slow CI runners stay within the test timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cd382d38e54eb2148330b9041a40929c240e363. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/src/haystack/routes.ts
+++ b/backend/src/haystack/routes.ts
@@ -138,23 +138,38 @@ export const createHaystackRoutes = (settings: Settings, auth: Auth, deps?: Hays
       },
       async open(ws) {
         const log = createStandaloneLogger(settings)
-        const data = ws.data as unknown as { request?: Request }
+        const slot = ws.data as unknown as {
+          request?: Request
+          __haystackServer?: HaystackAcpServer
+          __pendingMessages?: string[] | null
+        }
+
+        // Buffer any client message that arrives during `open()`'s async work.
+        // The client's `onopen` fires as soon as the WS handshake completes,
+        // and Bun delivers `message(ws, ...)` before this handler's `await`s
+        // resolve. Without a queue, the message hits `__haystackServer` while
+        // it's still undefined and gets dropped, leaving the client hanging
+        // on the response. Set this synchronously, before any await.
+        const queue: string[] = []
+        slot.__pendingMessages = queue
 
         // Validate the bearer exactly once per accepted socket. Doing this in
         // `beforeHandle` instead is unsafe because Bun's adapter can call it
         // more than once per upgrade. The bearer rides a subprotocol entry
         // (browsers can't set `Authorization` on `new WebSocket()`); it is
         // verified via the same Better Auth path REST uses (HMAC + DB lookup).
-        const subprotocolHeader = data.request?.headers.get('sec-websocket-protocol') ?? null
+        const subprotocolHeader = slot.request?.headers.get('sec-websocket-protocol') ?? null
         const user: User | null = await authorizeWsBearer(auth, subprotocolHeader)
         if (!user) {
+          slot.__pendingMessages = null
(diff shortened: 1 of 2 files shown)


darkbanjo:
Closing — consolidating the WS-flake fix. Synthesis: the failures all trace to the whole backend suite running in one long-lived Bun process under --rerun-each 5, accumulating pollution (leaked timers, degrading shared PGlite connection) that tips the new same-process WS e2e tests from 'late' into 'dropped'. Pursuing a non-source process-hygiene fix (bun --isolate) instead of source changes.

PR #932, opened 2026-06-01
feat(THU-549): Workspaces Foundation

raivieiraadriano92:
## Summary

Foundational infrastructure for Workspaces v1 — no user-visible change. Implements the THU-549 ticket plus the PR review feedback that followed.

The trust-domain model: every per-server resource (auth token, device id, encryption-key IndexedDB, SQLite filename) is namespaced by a stable `serverId` returned from `GET /v1/config`. The frontend keeps a `trust-domain registry` in `localStorage` that holds known servers + an `activeTrustDomain` pointer; switching servers (post-v1) becomes a pointer flip with each server's session intact.

## Architecture

```mermaid
flowchart TB
  subgraph Boot["Boot decision (resolveBootTrustDomain)"]
    direction TB
    Start([app start])
    HasActive{registry has<br/>activeTrustDomain?}
    Standalone{VITE_STANDALONE_MODE_ENABLED?}
    DefaultUrl{VITE_THUNDERBOLT_CLOUD_URL set?}
    ChooseStandalone["activate Standalone"]
    FetchConfig["fetch GET /v1/config<br/>→ pin serverId"]
    ChooseServer["activate Server (serverId)"]
    NoDomain["fatal: NO_TRUST_DOMAIN<br/>(mode picker — PR 5)"]

    Start --> HasActive
    HasActive -- yes --> Use[use cached]
    HasActive -- no --> Standalone
    Standalone -- yes --> ChooseStandalone
    Standalone -- no --> DefaultUrl
    DefaultUrl -- yes --> FetchConfig --> ChooseServer
    DefaultUrl -- no --> NoDomain
  end

  Boot ==> Registry

  subgraph Registry["Trust-Domain Registry (localStorage 'thunderbolt-trust-domains-v1')"]
    direction TB
    Local["localUserId<br/>(device-scoped UUIDv7)"]
    Active["activeTrustDomain (pointer)<br/>{ kind: 'standalone' }<br/>| { kind: 'server', serverId }"]
    Servers["servers: Record&lt;serverId, ServerEntry&gt;<br/>ServerEntry { serverId, cloudUrl, userId?, isAnonymous? }"]
  end

  Active -. points to .-> Servers
```

## What's in the PR

The branch is split into focused commits — easiest to review one at a time.

| Commit | Subject |
|---|---|
| `8c1dd033` | `serverId` + extended `/v1/config` payload (BE env, Zod schema, `make doctor` auto-mint, FE `AppConfig`) |
| `f0a99eb6` | Trust-domain registry + namespaced auth/device/keys (`thunderbolt_*__<serverId>`, `thunderbolt-keys__<serverId>`) |
| `e2cd8b6f` | Boot env vars + trust-domain decision tree (`VITE_STANDALONE_MODE_ENABLED`, `VITE_ALLOW_USER_ADDED_SERVERS`, `resolveBootTrustDomain`) |
| `190650d2` | Scope DB filename by trust domain (`standalone.db` / `server-<serverId>.db`) + `BroadcastChannel('thunderbolt-db-lifecycle')` |
| `f6315fd5` | Logout = unconditional wipe (drops the "keep my data" prompt; sequence per ticket §5) |
| `5d577651` | Scope cloud URL by active server (registry entry is authoritative; local-settings drops `cloudUrl`) |
| `4ed54043` | Collapse registry actions into `activateServer` / `activateStandalone` (atomic upsert + activate; `patchActiveServer` warns on no-op) |
| `0253fd7a` | Persist per-server session on `ServerEntry` (`userId` / `isAnonymous`) + wire `SessionToRegistryMirror` |
| `77f7d8a7` | Drop unused `db-reopened` event + `subscribeDbLifecycle` export |
| `c8a271ef` | Add `useActiveCloudUrlOrThrow` + migrate React subscribers (no more silent relative-URL fallback) |
| `74e603a7` | Extract shared `sync-worker-name` format (round-trip-tested; producer + SharedWorker consumer stay in sync) |

## Notable decisions

- **`serverId` is required, no default** — TS-enforced via `z.string().uuid()`; prevents ID duplication across deployments. `make doctor` auto-generates one for local dev.
- **Single env var for default URL** — kept `VITE_THUNDERBOLT_CLOUD_URL` instead of renaming to `VITE_DEFAULT_SERVER_URL` (spec deviation; smaller blast radius).
- **`/signed-out` is SSO-only** — consumer mode logout reloads back to the unauth landing (waitlist / sign-in). Matches pre-existing behavior.
- **Standalone is unreachable in v1 production** — boot fails with `STANDALONE_NOT_SUPPORTED` if `VITE_STANDALONE_MODE_ENABLED=true` until end-to-end standalone support lands (post-v1). The flag + resolver branch ship so the mode picker can exercise the plumbing in dev.
- **Multi-tab DB lifecycle** — `db-closing` / `db-deleted` reload other tabs on logout/wipe. `db-reopened` was dropped (no producer, no consumer — YAGNI).
- **Per-server session lives on `ServerEntry`** — `userId` + `isAnonymous` are mirrored from Better Auth by `SessionToRegistryMirror`. Multi-server-ready: each known server keeps its own session across boots; `activeTrustDomain` is just a pointer.

## Test plan

- [ ] Sign in (consumer / OTP) → reload → still signed in.
- [ ] Sign in → sign out → land on unauth UI; verify OPFS `server-<serverId>.db` is gone, `thunderbolt_auth_token__<serverId>` cleared, `thunderbolt-keys__<serverId>` IDB removed, `localStorage['thunderbolt-local-settings']` reset.
- [ ] Sign in → sign out → sign in again → fresh DB, no leftover state.
- [ ] SSO sign in → sign out → lands on `/signed-out` (not waitlist).
- [ ] Revoked device flow → confirms wipe, replaces to `/`.
- [ ] Boot Step 0 timing: `step3_waitForInitialSync` should NOT hit the 10s ceiling on the boot after a fresh logout (`syncEnabled` flipped off as part of the wipe).
- [ ] Dev settings cloud URL edit → next HTTP request hits the new URL (registry-backed now).
- [ ] Encryption: a chat message round-trips through PowerSync after a fresh sign-in (verifies `thunderbolt-keys__<serverId>` is reachable from the SharedWorker via `self.name` recovery).
- [ ] Multi-tab: tab A → logout. Tab B (same server) reloads on `db-closing` / `db-deleted`.
- [ ] `make check` and `bun test` green.

---

> [!NOTE]
> **High Risk**
> Destructive migrations wipe existing synced user data; logout now always wipes local DB/secrets; changing SERVER_ID resets all connected clients.
> 
> **Overview**
> Introduces **per-server trust domains** keyed by a required deployment **`SERVER_ID`** returned from **`GET /v1/config`**, plus the plumbing to pick an active domain at boot, namespace auth/device/encryption/OPFS DB paths by `serverId`, mirror session onto registry entries, and **unconditionally wipe** local state on logout (with multi-tab DB lifecycle signaling).
> 
> On the backend/data side, adds **workspace** tables (memberships, permissions, invites) and migrates PowerSync content tables from **user-scoped** to **workspace-scoped** keys via a **pre-release reset** that deletes existing rows before adding `NOT NULL workspace_id`. **`/v1/config`** also exposes anonymous-user and workspace-creation policy flags; env examples document boot flags and **`make doctor`** minting **`SERVER_ID`**. CI installs root deps so **`shared/`** type-check resolves **`uuid`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5783e59807af15eb435fa1535c7c7a812adf9b81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/cleanup.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  calls.push('clearDeviceId')
+})
+
+const handleFullWipe = mock(async () => {
+  calls.push('handleFullWipe')
+})
+
+// Include the full surface so this test's mock.module call doesn't shadow the real
+// exports for other test files (see docs/development/testing.md §65).
+mock.module('@/db/db-lifecycle-broadcast', () => ({
```
ital0, line 30: are these `mock.module` really required?

backend/src/config/settings.ts
```diff
@@ -153,6 +166,8 @@ const parseSettings = (): Settings => {
     microsoftClientSecret: process.env.MICROSOFT_CLIENT_SECRET || '',
     authMode: (process.env.AUTH_MODE || 'consumer').toLowerCase(),
     authAllowAnonymous: process.env.AUTH_ALLOW_ANONYMOUS === 'true',
+    allowWorkspaceCreationByAnon: process.env.ALLOW_WORKSPACE_CREATION_BY_ANON === 'true',
+    allowWorkspaceCreationByMembers: process.env.ALLOW_WORKSPACE_CREATION_BY_MEMBERS === 'true',
     oidcClientId: process.env.OIDC_CLIENT_ID || '',
     oidcClientSecret: process.env.OIDC_CLIENT_SECRET || '',
     oidcIssuer: process.env.OIDC_ISSUER || '',
```
ital0, line 187: `serverId` is required here with no default, so any backend test that parses settings from env again via `clearSettingsCache` throws a ZodError in CI, where `SERVER_ID` is never set. That's what's reddening the backend job right now. Seeding a fixture `SERVER_ID` in the test preload, right next to `BETTER_AUTH_SECRET`, fixes all of them in one place. `createTestSettings` got updated but the env path didn't.

src/components/logout-modal.tsx
```diff
@@ -2,116 +2,74 @@
...
+  // Per addendum decision #16 (THU-549 §5): signing out always wipes the active trust
+  // domain's local data. The previous "keep my data" affordance was incompatible with
+  // per-trust-domain SQLite files — leftover data without a matching auth token has no
+  // path back to the user. AlertDialogAction closes the dialog synchronously; the async
+  // wipe + navigate happen after the close, mirroring the delete-account flow.
+  const handleLogout = () => {
+    // SSO lands on `/signed-out` because IdP-bounce-back would silently re-auth the
+    // user on reload; consumer mode reloads so the user re-enters the normal unauth
+    // landing (sign-in or waitlist).
+    void signOutAndWipe({
```
ital0, line 56: I think the dialog closes right away and the wipe runs in the background, so we lose the old loading state and the app stays live while its db is being closed and deleted under it. 

I believe worth a check on that.

src/lib/auth-token.ts
```diff
@@ -3,45 +3,80 @@
...
+ * the device headers cleanly in that case.
+ */
 export const getDeviceId = (): string => {
-  let id = localStorage.getItem(deviceIdKey)
+  const serverId = getActiveServerId()
+  if (!serverId) {
+    return ''
+  }
+  const key = deviceIdKeyFor(serverId)
+  let id = localStorage.getItem(key)
```
ital0: do you think we can get rid of this `let`?

ital0 (changes requested):

backend/src/config/settings.ts
raivieiraadriano92: fixed

src/lib/auth-token.ts
raivieiraadriano92: fixed

src/components/logout-modal.tsx
raivieiraadriano92: fixed

src/lib/platform.ts
```diff
@@ -203,22 +205,28 @@ export const getDatabaseType = async (): Promise<DatabaseType> => {
...
+  const domain = getActiveTrustDomain()
+  if (!domain) {
+    throw new Error('Cannot resolve database path: no active trust domain (boot decision tree must run first)')
+  }
+  const filename = getDbFilenameFor(domain)
+
   // For native databases (bun-sqlite), use file path directly
   if (databaseType === 'bun-sqlite') {
-    return `${appDataDirPath}/thunderbolt.db`
+    return `${appDataDirPath}/${filename}`
```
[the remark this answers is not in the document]
raivieiraadriano92: Not a concern for this PR — we're doing a full DB reset on deploy, so existing local data will be wiped. Old unprefixed localStorage keys (thunderbolt_auth_token, thunderbolt_device_id) become dead weight but pose no security risk since the tokens are invalidated server-side.

src/lib/cleanup.ts
```diff
@@ -2,61 +2,135 @@
...
-    useLocalSettingsStore.setState(initialLocalSettings)
   }
 
-  if (clearAuth) {
-    clearAuthToken()
-    clearDeviceId()
-  }
+  // onComplete fires synchronously right after signOut resolves — no await
+  // between them so React cannot schedule a re-render before the hard navigation.
+  onComplete()
```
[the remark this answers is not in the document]
raivieiraadriano92: False positive — this is the intentional tradeoff of the SSO race fix (see previous comment). Wipe must run before signOut so Better Auth's session cache stays populated during the async window, preventing AuthGate from redirecting to /sso-redirect. The DB-closed window is ~100–500ms (length of the signOut HTTP call), during which PowerSync is already disconnected (setSyncEnabled(false) runs first) and the logout modal blocks all user interaction. Any stale query errors in that window are irrelevant — onComplete hard-navigates immediately after.

src/lib/cleanup.ts
```diff
@@ -2,61 +2,135 @@
...
+  // a logout into the next user's session.
+  useLocalSettingsStore.setState(initialLocalSettings)
+
+  // Step 6: clear per-server credentials. No-op in standalone.
+  clearAuthToken()
+  clearDeviceId()
+
+  // Step 7: clear encryption keys (server-only — `handleFullWipe` throws on standalone
+  // via `key-storage.ts`'s active-server guard; we skip it cleanly when there's no server).
+  if (trustDomain?.kind === 'server') {
```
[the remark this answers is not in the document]
raivieiraadriano92: False positive — intentional by design. ServerEntry.userId surviving sign-out is documented in auth-context.tsx: the value is last-known session metadata for the multi-server UI, not a credential. With the auth token cleared, no authenticated request can be made using it. getActiveUserId() has no callers in the current codebase.

backend/src/config/settings.ts
```diff
@@ -9,6 +9,12 @@ import { z } from 'zod'
  */
 const settingsSchema = z
   .object({
+    // Stable per-deployment UUID. Returned by GET /v1/config and used by the frontend to
+    // key the trust-domain registry (auth token, device ID, encryption keys, DB filename
+    // are all namespaced by this). No default — TS-enforced to prevent ID duplication
+    // across deployments. `make doctor` auto-generates one for local dev.
+    serverId: z.string().uuid(),
```
ital0: Missing `SERVER_ID` throws a raw ZodError at boot. What about a `superRefine` message like BETTER_AUTH_SECRET has?

src/lib/cleanup.ts
```diff
@@ -38,35 +56,105 @@ export const clearLocalData = async (options?: ClearLocalDataOptions): Promise<v
...
+  }
+
+  // Reset local settings to defaults. Pre-trust-domain these lived in the SQLite DB
+  // and disappeared with `resetAppDir`; now that they're in their own Zustand store,
+  // the wipe has to clear them explicitly so theme/sync/native-fetch/etc. don't survive
+  // a logout into the next user's session.
+  useLocalSettingsStore.setState(initialLocalSettings)
+
+  // Step 6: clear per-server credentials. No-op in standalone.
+  clearAuthToken()
```
ital0: I think this kills the session only locally in consumer mode. The token is cleared here, before `signOut()` runs at `:151`, so the request goes out with no bearer and the server never revokes the session. What about clearing the token after `signOut()` instead?

src/settings/dev-settings.tsx
```diff
@@ -9,21 +9,35 @@ import { SectionCard } from '@/components/ui/section-card'
...
   const setLocalSetting = useLocalSettingsStore((s) => s.setLocalSetting)
 
+  // Cloud URL lives on the active server entry in the trust-domain registry; editing
+  // it here updates the registry directly so runtime consumers (HTTP, PowerSync, etc.)
+  // see the change on next read. Resetting falls back to the env-var default that the
+  // boot resolver also uses. NOTE: changing the URL does NOT update the active server's
+  // `serverId` — pointing at a different backend (different serverId) is post-v1 territory.
+  const cloudUrl = useActiveCloudUrl() ?? defaultCloudUrl
+  const patchActiveServer = useTrustDomainRegistry((s) => s.patchActiveServer)
+  const setCloudUrl = (value: string) => patchActiveServer({ cloudUrl: value || defaultCloudUrl })
```
ital0: `setCloudUrl` silently no-ops when not in server mode, so the input looks editable but nothing saves. Can you disable the field unless the active domain is a server?

src/components/revoked-device-modal.tsx
```diff
@@ -2,84 +2,63 @@
...
-              <Loader2 className="mr-2 h-4 w-4 animate-spin" />
-              {selectedOption === 'delete' ? 'Deleting...' : 'Signing out...'}
-            </>
-          ) : (
-            'Confirm'
-          )}
-        </Button>
-      </ResponsiveModalFooter>
-    </ResponsiveModal>
+    <AlertDialog open={open}>
```
ital0, line 47: No `onOpenChange` guard during the wipe here, but `LogoutModal` has one. Can you mirror it? Not sure if would make sense.

ital0 (changes requested):
Really nice PR overall. The trust-domain model is clean and the per-server namespacing is well thought out. I found one real bug, one thing to sort before a prod deploy, and a handful of small nits.

src/lib/cleanup.ts
raivieiraadriano92: Good catch — you're right. Fixed in dcc374eb: `signOutAndWipe` now defers `clearAuthToken`/`clearDeviceId` until after `signOut()` resolves, so the /sign-out HTTP call still carries the bearer and the BE can revoke the session row. The wipe-before-signOut order is preserved (still keeps the AuthGate-mid-wipe guarantee).

src/lib/cleanup.ts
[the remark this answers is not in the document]
raivieiraadriano92: Re-examining this — my earlier "false positive" call was incomplete. ital0 raised the same root cause from a different angle (https://github.com/thunderbird/thunderbolt/pull/932#discussion_r3388848553): clearing the auth token before `signOut()` runs leaves the /sign-out HTTP request unauthenticated, so the BE can't revoke the session row. Fixed in dcc374eb: the wipe still runs before `signOut` (preserves the AuthGate redirect guard) but `clearAuthToken`/`clearDeviceId` are deferred to after `signOut()` returns, so the revoke call is authenticated.

src/db/db-lifecycle-broadcast.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  reloadListenerInstalled = true
+  channel.addEventListener('message', (event: MessageEvent<DbLifecycleEvent>) => {
+    const active = getActiveTrustDomain()
+    if (!active || !isSameTrustDomain(active, event.data.trustDomain)) {
+      return
+    }
+    console.info(`[db-lifecycle] received ${event.data.kind} from another tab — reloading`)
+    window.location.reload()
+  })
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 7c779961.

src/components/logout-modal.tsx
```diff
@@ -2,116 +2,91 @@
...
+        await authClient.signOut()
+      },
+      onComplete: () => {
+        if (isSsoMode()) {
+          window.location.replace('/signed-out')
+        } else {
+          window.location.reload()
+        }
+      },
+    })
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 71a08204.

backend/src/config/settings.ts
raivieiraadriano92: Fixed in 8e33ff79.

src/settings/dev-settings.tsx
raivieiraadriano92: Fixed in 5d89e0a7 — removed the Cloud URL input entirely. Switching backends should go through the sign-out → ModePicker → re-activate flow (the input only patched cloudUrl on the existing server entry without changing serverId, so cross-serverId pointing was silently broken anyway).

src/lib/cleanup.test.ts
raivieiraadriano92: Yes — `clearLocalData` calls module-level imports (`broadcastDbLifecycle`, `resetDatabase`, `setSyncEnabled`, `deleteDbFile`) that aren't part of its public signature, so observing them requires module mocks. The spread pattern (`...realModule, mock`) at the top of the file is the `testing.md §65` mitigation against the global-shadow leak. The deps the function *does* accept (`clearAuthToken`, `clearDeviceId`, `handleFullWipe`) are injected via DI — those are the ones that matter for asserting interactions.

src/ai/eval/runner.ts
```diff
@@ -21,12 +20,15 @@ import type { EvalResult, EvalScenario } from './types'
...
+// matches the boot resolver's default so dev and eval point at the same backend.
+const evalCloudUrl = import.meta.env.VITE_THUNDERBOLT_CLOUD_URL || 'http://localhost:8000/v1'
+
 let _evalHttpClientPromise: Promise<import('@/lib/http').HttpClient> | null = null
 const getEvalHttpClient = () => {
   if (!_evalHttpClientPromise) {
     _evalHttpClientPromise = (async () => {
-      const cloudUrl = getLocalSetting('cloudUrl')
-      return createAuthenticatedClient(cloudUrl, getAuthToken, {
+      return createAuthenticatedClient(evalCloudUrl, getAuthToken, {
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in ad08c09b.

src/components/revoked-device-modal.tsx
raivieiraadriano92: Looked at this — the guard isn't needed in this modal. RevokedDeviceModal has no Cancel button, no `onOpenChange` prop, and `onEscapeKeyDown={(e) => e.preventDefault()}` blocks ESC. The only close path Radix exposes is unwired, and the parent (`use-credential-events`) only flips `open` from false→true on the revocation signal and never flips it back during the wipe. The `wipingRef` double-click guard inside `handleConfirm` covers the only remaining race.

PR #941, opened 2026-06-02
test(backend): make proxy WS e2e tests event-driven (load-immune)

darkbanjo:
## Root cause

The two flaky `proxy/ws — e2e` tests asserted on a **wall-clock deadline** (fixed 100ms wait / 50-iteration poll) for an asynchronous WebSocket round-trip. Under the `--rerun-each 5 --randomize` load (~20k test executions on one Bun event loop), the frame's JS delivery callback gets *scheduled late* — so the deadline expires before the event fires. The test fails not because the relay is wrong, but because the event loop was busy. That couples a deterministic question ("did the relay forward the message?") to a non-deterministic one ("did it happen within N ms?").

## Fix

Await the **actual event** — a promise resolved from `onmessage` / the upstream socket's `onclose` — with a 15s per-test timeout as a failsafe (not a correctness signal). The test now passes the instant the event fires, however long the loop took to get there, and only fails on a genuine hang.

- `relays text messages bidirectionally`: `await echoed` instead of `sleep(100); expect(messages).toContain(...)`.
- `upstream close code propagates`: `await upstreamClosed` instead of polling `observed.code`.

No CI/config changes — the tests stay in the normal randomized 5× run, so a green here proves they're load-*immune*, not just load-*tolerant*. Verified locally at `--rerun-each 10` → 110/110.

## Not in scope
- The `haystack/ws` flake is a *separate, real server race* (frames arriving during the async `open()` get dropped) — fixed by #929's `__pendingMessages` buffering, not by a test change.
- The rate-limit cascade (multi-minute hangs) is also separate (#929's `clearExpiredByTimeout` guard).

---

> [!NOTE]
> **Low Risk**
> Test-only changes in ws-e2e.test.ts; no production proxy or relay logic modified.
> 
> **Overview**
> Two **proxy WebSocket e2e** tests in `ws-e2e.test.ts` no longer use fixed delays or poll loops to observe async behavior. **Bidirectional relay** now resolves a promise from `client.onmessage` when the expected `echo: hello` frame arrives, then asserts on that value. **Upstream close propagation** awaits a promise from the hooked upstream socket’s `onclose` instead of polling `observed.code` for up to ~2.5s.
> 
> Each test gets a **15s Bun timeout** as a hang failsafe only; correctness is tied to the event firing, not a millisecond budget, so heavy randomized reruns are less likely to flake on busy event loops.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21a08469c991f8149661b90fd427621a34d3a936. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/src/proxy/ws-e2e.test.ts
+++ b/backend/src/proxy/ws-e2e.test.ts
@@ -156,7 +156,9 @@ describe('Universal proxy WebSocket relay /v1/proxy/ws — e2e', () => {
     // The per-test timeout below is only a failsafe for a genuine hang.
     const echoed = new Promise<string>((resolve) => {
       client.onmessage = (e: MessageEvent) => {
-        if (e.data === 'echo: hello') resolve(e.data as string)
+        if (e.data === 'echo: hello') {
+          resolve(e.data as string)
+        }
       }
     })
 
--- a/backend/src/proxy/ws-e2e.test.ts
+++ b/backend/src/proxy/ws-e2e.test.ts
@@ -150,12 +150,15 @@ describe('Universal proxy WebSocket relay /v1/proxy/ws — e2e', () => {
       buildProtocols('wss://upstream.test/path', handle.bearerToken, ['acp.v1']),
     )
 
-    const messages: string[] = []
-    // Use onmessage rather than addEventListener: in Bun's same-process WS the
-    // addEventListener path has been observed to drop late-bound 'message' events.
-    client.onmessage = (e: MessageEvent) => {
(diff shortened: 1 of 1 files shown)


darkbanjo:
Closing: event-driven assertions alone don't fix this — the events are *dropped* (not delayed) under load (a Bun in-process-WS limitation), so awaiting times out. And converting to mock-socket unit tests would violate the proxy's integration-testing conventions. Regrouping on a convention-preserving fix on a fresh branch.

PR #942, opened 2026-06-02
feat: support authenticated remote MCP servers (bearer / API key)

ital0:
> [!NOTE]
> **Medium Risk**
> Dropping synced `mcp_servers` is a one-way schema change that removes cross-device MCP config; chat now depends on local credentials and reconnect logic for external tool calls.
> 
> **Overview**
> **MCP config is now device-local instead of PowerSync-synced.** Migration `0019` drops `powersync.mcp_servers`; client schema registers `mcp_servers` and new `mcp_secrets` as local-only tables, with backend/PowerSync service wiring removed accordingly.
> 
> **Remote MCP (HTTP/SSE) gains bearer credentials, richer settings UX, and safer runtime behavior.** Secrets live in `mcp_secrets` and are written/deleted atomically with server rows (`createMcpServerWithCredentials`, transactional delete). The settings flow adds name/transport/credential fields, URL-derived default names, and connection tests that use the same proxy transport path as production.
> 
> **Chat tool integration is namespaced and reconnect-aware.** MCP tools are prefixed per server (`mergeMcpTools` / `sanitizeToolPrefix`) with optional reconnect on closed transports; the chat store holds a live `getMcpClients()` getter and `reconnectClient` from the provider (coalesced connect/reconnect, orphan cleanup). Transport creation adds SSE support and fixes the SDK `protocolVersion` assignment seam for `@ai-sdk/mcp`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5cab14584b76a2b18436a75e49f36517fccbb935. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/ai/fetch.ts
```diff
@@ -96,19 +109,106 @@ export const getTinfoilClient = async (): Promise<SecureClient> => {
...
+        } catch (retryErr) {
+          console.warn('MCP server still failing after reconnect; skipping its tools for this send', retryErr)
+          return 0
+        }
+      }
+    }
+
+    const added = await merge()
+    if (added > 0) {
+      mcpServerEntries.push(`- ${prefix} (${added} tools)`)
```
raivieiraadriano92: Just a small detail - this could render something like `1 tools`.

The fix could be `- ${prefix} (${added} ${added === 1 ? 'tool' : 'tools'})`

src/ai/fetch.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    const second = named('render', async () => ({ deploy: tool('second') }))
+
+    const { toolset, summary } = await mergeMcpTools({}, [first, second], async () => null)
+
+    // Same base prefix → first keeps `render`, second becomes `render_2`; both
+    // `deploy` tools survive because the prefixes differ.
+    expect(toolset.render_deploy).toEqual(tool('first'))
+    expect(toolset.render_2_deploy).toEqual(tool('second'))
+    expect(Object.keys(toolset).sort()).toEqual(['render_2_deploy', 'render_deploy'])
+    expect(summary).toBe('- render (1 tools)\n- render_2 (1 tools)')
```
raivieiraadriano92: maybe we should fix the grammar for `1 tool`, `2 tools` and so on

src/settings/mcp-servers.tsx
```diff
@@ -17,32 +17,64 @@ import {
...
+ * Derives a short, meaningful server name from a remote MCP URL — used to
+ * pre-fill (and re-derive) the editable name field. The name namespaces the
+ * server's tools in the prompt, so a readable default like `github` or `render`
+ * beats the raw hostname.
+ * - Localhost: includes port for disambiguation (`localhost-3000`)
+ * - Remote: 3+ domain segments → second-to-last (`api.github.com` → `github`);
+ *   2 segments → first (`render.com` → `render`); 1 → as-is
+ */
+const generateServerName = (url: string): string => {
+  try {
```
raivieiraadriano92, line 55: could verify the edge case below?

new URL('http://192.168.1.100').hostname → '192.168.1.100'
parts.length >= 3 → parts[parts.length - 2] = '1'

The resulting name 1 would become a very unhelpful tool prefix. A guard like if (/^\d+\.\d+\.\d+\.\d+$/.test(hostname)) return 'server' would cover this edge case.

src/settings/mcp-servers.tsx
```diff
@@ -113,16 +154,30 @@ export default function McpServersPage() {
...
   const addServerMutation = useMutation({
     mutationFn: async ({ name, url }: { name: string; url: string }) => {
+      const id = uuidv7()
+      // Persist credentials BEFORE the server row. useMcpSync reacts to the new
+      // mcp_servers row and the provider connects by reading credentials from the
+      // DB at connect time — if the token isn't stored yet the first connect is
+      // unauthenticated and nothing reconnects it. Writing the secret first
+      // (keyed by the same id) guarantees the connect sees the token.
+      if (newServerToken) {
+        await setMcpServerCredentials(db, id, { type: 'bearer', token: newServerToken })
```
raivieiraadriano92: can you check if the new created credential would be orphaned if the createMcpServer fails?
maybe these two operations should be in a transaction

src/settings/mcp-servers.tsx
```diff
@@ -145,27 +211,16 @@ export default function McpServersPage() {
...
+      // Create a real MCP client using the same method as the provider —
+      // route through the universal proxy so the test matches the real
+      // connection path (web CORS would otherwise fail for remote servers).
+      const headers = buildMcpHeaders(newServerToken || undefined)
+      const transport = createMcpTransport(newServerUrl, newServerTransport, cloudUrl, headers)
+      const mcpClient = await createMCPClient({ transport })
 
       // Try to get tools to verify the connection works
-      console.log('Requesting tools...')
       const tools = await mcpClient.tools()
```
raivieiraadriano92: I think here if the `tools()` throws and exception the mcp connection would stay alive.
maybe we we should call `mcpClient.close()` at the `finally` block

not sure about this, could you check?

src/db/tables.ts
```diff
@@ -122,12 +122,18 @@ export const integrationsSecretsTable = sqliteTable('integrations_secrets', {
   enabled: integer('enabled').default(0),
 })
 
+/** Local-only table for MCP server credentials (bearer tokens / API keys). Never synced via PowerSync. */
+export const mcpSecretsTable = sqliteTable('mcp_secrets', {
+  id: text('id').primaryKey(), // = mcp_servers.id
+  credentials: text('credentials'), // JSON blob (e.g. { type: 'bearer', token })
```
raivieiraadriano92, line 128: Question, the secrets are stored in a local-only table, so they are never synced across other devices.
Do we have some sort of icon/indicator to show the user that the MCP is missing the auth credentials? Does it make sense?

ital0: Awesome catch Raí!

When I firstly architecting this, it was supposed to sync MCP servers. But then I noticed syncing secrets is not a good pattern. I completetly removed that. I think we can move back here if we decide to sync.

raivieiraadriano92 (approved):

PR #954, opened 2026-06-04
fix(THU-580): cache Better Auth session so app boots offline

raivieiraadriano92:
Closes [THU-580](https://linear.app/mozilla-thunderbolt/issue/THU-580/desktop-app-fails-to-start-when-network-is-offline).

Better Auth's `useSession` needs a `/get-session` round-trip to populate `data`. Offline at boot that fetch fails, `data` stays `null`, and `useAuthGate` redirects authenticated routes to `/waitlist` — the user is stuck on the login screen until they reconnect.

Cache the last successful session payload in `localStorage` and seed `client.\$store.atoms.session` on auth-client creation. `useAuthQuery.onError` preserves non-null `data` on non-401 errors, so the cached value survives the failing offline fetch. Better Auth's own online listener replaces it with fresh server data as soon as the network returns. Cache is cleared on 401 and on `clearLocalData` (logout / wipe).

---

> [!NOTE]
> **Medium Risk**
> Touches auth boot path and stores richer profile data in localStorage alongside the token; invalidation on 401/logout/wipe and expiry checks limit stale-session risk.
> 
> **Overview**
> Fixes offline startup for users who already have a bearer token: the last successful session is stored in **localStorage** and used to **seed** Better Auth’s session atom before `/get-session` runs, so protected routes are not treated as logged-out when the network is down.
> 
> **`session-cache`** adds read/write/clear plus **`isCachedSessionValid`** (future `session.expiresAt` only). **`hydrateSessionFromCache`** runs when the auth client is created (token + valid cache required); **`subscribeSessionCachePersist`** mirrors atom updates that include both user and session, with unsubscribe on client swap.
> 
> The cache is cleared on **401**, **`clearLocalData`**, and **cross-tab token rotation** (before reload) so stale or wrong-user data is not reused offline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df5b928bb810c695224a1efdbae7d9ac606c1ebc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/contexts/auth-context.tsx
+++ b/src/contexts/auth-context.tsx
@@ -8,7 +8,7 @@ import { usePowerSyncCredentialsInvalidListener } from '@/hooks/use-powersync-cr
 import { isSsoMode } from '@/lib/auth-mode'
 import { clearAuthToken, getAuthToken, onAuthTokenChangedInOtherTab, setAuthToken } from '@/lib/auth-token'
 import { getPlatform } from '@/lib/platform'
-import { clearCachedSession, getCachedSession, setCachedSession } from '@/lib/session-cache'
+import { clearCachedSession, getCachedSession, isCachedSessionValid, setCachedSession } from '@/lib/session-cache'
 import { anonymousClient, emailOTPClient } from 'better-auth/client/plugins'
 import { createAuthClient } from 'better-auth/react'
 import { consumePendingSsoAnonAlias } from '@/lib/analytics/anonymous-promotion-sso-bridge'
@@ -54,16 +54,25 @@ const createAuthClientInstance = (cloudUrl: string) => {
  * (Better Auth's online listener fires this automatically when the network
  * returns).
  *
+ * An expired cache is discarded rather than seeded: if the device has been
+ * offline past `session.expiresAt`, every subsequent API call would 401
+ * anyway, so it's safer to force a fresh `/get-session` than to flash a
+ * logged-in UI that will be torn down moments later.
+ *
  * Only the seed is synchronous (no lifecycle to clean up). Persisting future
  * updates is wired in {@link subscribeSessionCachePersist} from the provider so
  * the listener is released when the auth client is replaced.
  */
-const hydrateSessionFromCache = (client: ReturnType<typeof createAuthClient>) => {
+export const hydrateSessionFromCache = (client: ReturnType<typeof createAuthClient>) => {
   if (!getAuthToken()) {
     return
   }
   const cached = getCachedSession()
   if (!cached) {
+    return
+  }
+  if (!isCachedSessionValid(cached)) {
+    clearCachedSession()
     return
   }
   const sessionAtom = client.$store.atoms.session
@@ -216,6 +225,10 @@ export const AuthProvider = ({ children, cloudUrl, authClient: overrideClient }:
   useEffect(() => {
     return onAuthTokenChangedInOtherTab((next, prev) => {
       if (next && next !== prev) {
+        // Cross-tab rotation may swap in a different user identity; wipe the
+        // cache so the reloaded tab doesn't seed the previous user's session
+        // under the new token before /get-session lands.
+        clearCachedSession()
         window.location.reload()
         return
       }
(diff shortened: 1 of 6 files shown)


src/contexts/auth-context.tsx
```diff
@@ -37,6 +38,64 @@ const createAuthClientInstance = (cloudUrl: string) => {
...
+ * Only payloads with both `user` and `session` are persisted. Better Auth's
+ * refresh manager normalizes empty responses to `null` on its own refetch path
+ * but the initial `useAuthQuery` `onSuccess` writes the raw server payload, so
+ * `{ user: null, session: null }` can transit the atom — we filter it out here
+ * to avoid caching a payload that would only resurrect as a logged-out flash on
+ * the next boot.
+ */
+const subscribeSessionCachePersist = (client: ReturnType<typeof createAuthClient>): (() => void) => {
+  return client.$store.atoms.session.subscribe((state: { data: unknown }) => {
+    const data = state.data as { user?: unknown; session?: unknown } | null
```
ital0: Small thought: the file already exports `Session` (`AuthClient['$Infer']['Session']`), so the subscriber could lean on that instead of the `unknown` casts here. It would also let `CachedSessionData` carry a real shape rather than `Record<string, unknown>`.

src/contexts/auth-context.tsx
```diff
@@ -37,6 +38,64 @@ const createAuthClientInstance = (cloudUrl: string) => {
...
+    return
+  }
+  const cached = getCachedSession()
+  if (!cached) {
+    return
+  }
+  const sessionAtom = client.$store.atoms.session
+  sessionAtom.set({
+    ...sessionAtom.get(),
+    data: cached,
```
ital0, line 81: Quick question: should we check the cached session's expiry before seeding? If the device stays offline past `expiresAt`, the app would still boot as logged in.

src/contexts/auth-context.tsx
```diff
@@ -105,6 +165,18 @@ export const AuthProvider = ({ children, cloudUrl, authClient: overrideClient }:
     return { authClient: client }
   }, [cloudUrl, overrideClient])
 
+  // Persist session updates back to the localStorage cache (THU-580). Scoped to
+  // an effect so the listener is released when `value` is replaced — otherwise
+  // a cloudUrl change would keep the old client's refresh manager alive forever.
+  // Skipped when the auth client is an injected test override (no `$store`).
+  useEffect(() => {
+    const client = value?.authClient
+    if (!client?.$store?.atoms?.session) {
```
ital0: Tiny one: since real clients always have `$store.atoms.session`, checking `client?.$store` alone should cover the test override case. The triple optional chain reads a bit more defensive than it needs to be.

src/lib/session-cache.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/** Read the cached session, or null when absent / malformed. */
+export const getCachedSession = (): CachedSessionData | null => {
+  try {
+    const raw = localStorage.getItem(sessionCacheKey)
+    if (!raw) {
+      return null
+    }
+    const parsed = JSON.parse(raw)
+    if (parsed && typeof parsed === 'object') {
```
ital0: `typeof parsed === 'object'` also lets arrays through. Harmless in practice, but adding `!Array.isArray(parsed)` would make the guard match its intent.

ital0 (changes requested):

src/contexts/auth-context.tsx
```diff
@@ -105,6 +165,18 @@ export const AuthProvider = ({ children, cloudUrl, authClient: overrideClient }:
...
+      return
+    }
+    return subscribeSessionCachePersist(client)
+  }, [value])
+
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 846100f6.

src/contexts/auth-context.tsx
raivieiraadriano92: Fixed in 846100f6.

src/contexts/auth-context.tsx
raivieiraadriano92: Fixed in 0bd5ba64.

src/contexts/auth-context.tsx
raivieiraadriano92: Fixed in 8a1662a9.

src/lib/session-cache.ts
raivieiraadriano92: Fixed in 8a1662a9.

PR #958, opened 2026-06-06
feat(THU-553): mode picker UI — layout, selection state, and connectivity

raivieiraadriano92:
Polishes the mode picker screen shown on first boot when no trust domain is configured.

## Changes

- Dots indicator is `position: absolute` so it stays perfectly centered regardless of button widths
- Option card icons use lime when selected, muted when not
- Standalone option is disabled (not yet implemented); server is pre-selected by default
- Explicit spacing: 92px top padding, 150px branding→content, 40px title→cards, 24px cards→input, 32px content→footer
- Continue always proceeds even when the server can't be reached — falls back to the normalized URL and lets the app handle the unreachable state post-reload

## How to trigger this screen locally

The mode picker renders when the app boots with no trust domain configured (`NO_TRUST_DOMAIN`). To force it:

1. In `.env.local`, ensure both flags are off:
   ```
   VITE_THUNDERBOLT_CLOUD_URL=
   VITE_STANDALONE_MODE_ENABLED=false
   ```
2. Clear the saved trust domain from localStorage:
   ```js
   // paste in the browser console
   localStorage.removeItem('thunderbolt-trust-domains-v1')
   ```
3. Reload — the mode picker will appear instead of the normal app boot.

To exit the picker without a real server, click **Skip** (activates standalone locally) or enter any URL and click the arrow button.

---

> [!NOTE]
> **Medium Risk**
> Touches signup/membership promotion and profile-update hooks plus additive DB migrations; incorrect sync could show stale member info but does not change authz rules.
> 
> **Overview**
> Extends workspace data for richer UI and member lists without cross-bucket user joins in PowerSync.
> 
> **Schema:** `workspaces` gains optional `slug` and `icon`, with a partial unique index on non-null slugs. `workspace_memberships` gains `user_name` and `user_email`; migration backfills from `user`.
> 
> **Runtime behavior:** Signup promotion now seeds display fields when pending invites become memberships. A Better Auth **user update** hook calls `syncMembershipDisplayInfo` so profile renames/emails propagate to every membership row. User DAL reads include `name`/`email` to support that flow. Workspace DAL picks up `updateWorkspace` (replacing shared-only naming), `insertPersonalWorkspaceIfMissing`, and slug/icon on upserts.
> 
> **DX:** `SERVER_ID` env validation errors now point devs at `make doctor` and UUID format requirements, with unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c90eb632ab205fadff2e65b6d125bd91b18cd485. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/boot/mode-picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+    if (state.selection === 'standalone') {
+      useTrustDomainRegistry.getState().activateStandalone()
+      window.location.reload()
+      return
+    }
+
+    if (state.selection === 'server') {
+      dispatch({ type: 'CONNECT' })
+      const result = await validate(state.serverUrl)
+      if (!result.ok) {
```
ital0, line 194: Quick question on the Continue flow. The PR description says `Continue always proceeds even when the server can't be reached and falls back to the normalized URL`, but here a failed check shows the inline error and stops instead of moving on. Is the description ahead of the code, or should this path fall through to the server anyway?

src/components/boot/mode-picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const handleContinue = async () => {
+    if (state.selection === 'standalone') {
+      useTrustDomainRegistry.getState().activateStandalone()
+      window.location.reload()
+      return
+    }
+
+    if (state.selection === 'server') {
+      dispatch({ type: 'CONNECT' })
+      const result = await validate(state.serverUrl)
```
ital0: I think here the URL is getting validated twice. The blur check already confirms the server and shows the green check, then clicking Continue fires the same request again.

Do you think would make sense reusing the blur result when the URL hasn't changed, so we skip the extra round trip?

ital0 (changes requested):

src/components/boot/mode-picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  const canContinue =
+    state.selection !== null && !state.isValidating && (state.selection !== 'server' || state.serverUrl.trim() !== '')
+
+  const handleSkip = () => {
+    useTrustDomainRegistry.getState().activateStandalone()
+    window.location.reload()
+  }
+
+  const handleBlur = async () => {
```
ital0, line 139: I think there's a sneaky interaction here: clicking Continue while the cursor is still in the URL field triggers the blur validation first, which briefly disables the Continue button, so the click gets swallowed and the user has to click twice. It also means the same URL gets validated twice for one submit. Maybe we could validate only on submit, or keep the button enabled while the blur check runs?

ital0 (changes requested):

src/components/boot/mode-picker.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+      const buttons = screen.getAllByRole('button')
+      const arrowBtn = buttons[buttons.length - 1]
+      expect(arrowBtn).toBeDisabled()
+    })
+
+    it('does not show URL input on initial render', () => {
+      render(<ModePicker />)
+
+      expect(screen.queryByPlaceholderText('app.thunderbolt.io/')).not.toBeInTheDocument()
+    })
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 6512b24a + 191820e6 — tests now assert the server-preselected initial state (see `describe('initial render')` block).

src/components/boot/mode-picker.tsx
```diff
@@ -0,0 +1,246 @@
...
+      window.location.reload()
+      return
+    }
+
+    if (state.selection === 'server') {
+      dispatch({ type: 'CONNECT' })
+      const result = await validateServerUrl(state.serverUrl)
+      if (!result.ok) {
+        dispatch({ type: 'VALIDATE_ERROR', message: result.message })
+        return
```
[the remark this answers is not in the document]
raivieiraadriano92: Block-on-fail is correct here. `ModePicker` only renders when boot returns `NO_TRUST_DOMAIN` (`app.tsx:284`) — i.e. first-time setup. Returning users with an activated server skip the picker entirely (`resolveBootTrustDomain` line 50-52). Without a successful `/v1/config` we have no `serverId` to namespace the registry entry with, so "proceed-on-fail" would create a ghost trust domain pointing at a backend we never confirmed exists. Updating PR description to match — the code is the source of truth.

src/components/boot/mode-picker.tsx
raivieiraadriano92: Description is ahead of the code — fixing the description. Block-on-fail is correct: ModePicker is first-time setup (returning users skip it via `resolveBootTrustDomain` line 50-52), and the validate call is also where we fetch `serverId` to namespace the registry entry. Without it, activating produces a ghost trust domain.

src/components/boot/mode-picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  const handleBlur = async () => {
+    if (!isServerMode || !state.serverUrl.trim()) {
+      return
+    }
+    dispatch({ type: 'VALIDATE_START' })
+    const result = await validate(state.serverUrl)
+    dispatch(result.ok ? { type: 'VALIDATE_SUCCESS' } : { type: 'VALIDATE_ERROR', message: result.message })
+  }
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 71816a01.

src/components/boot/mode-picker.tsx
raivieiraadriano92: Fixed in 71816a01.

src/components/boot/mode-picker.tsx
raivieiraadriano92: Fixed in 71816a01.

PR #964, opened 2026-06-09
feat(THU-591): expose ACP agent commands in the chat slash menu

darkbanjo:
Closes THU-591.

## What

ACP agents advertise the commands / "skills" they support via the `available_commands_update` session notification. We received that event but **ignored** it. This surfaces those commands in the chat `/` slash menu while chatting with the agent, clearly marked as **External** (from the agent) vs the user's own authored skills.

## How

- **Capture** — `available_commands_update` now maps to a `SessionSideEffect` (`available_commands_changed`) carrying the advertised commands (`name`, `description`, optional `inputHint`). (`src/acp/translators/acp-to-ai-sdk.ts`)
- **Store** — a small per-agent store (`src/acp/agent-commands-store.ts`) populated by the chat instance's session side-effect handler (keyed by agent, since commands are an agent-level capability).
- **Surface** — the slash machine is generalized to a `SlashItem` union (user skills + agent commands). The popup lists skills first, then the agent's commands with an **"External"** badge. Selecting one inserts the `/name` token; invocation is a normal prompt (ACP commands take unstructured input — everything after the name is the argument).

## Scope / safety

- Only lights up for ACP agents that actually advertise commands; the Deepset/Haystack bridge agents advertise none, so nothing changes for them.
- User-skill behavior is unchanged (commands default to none → identical menu).
- No new telemetry events; `skill_used` still fires only for user skills.

## Tests

- New: translator emits the side effect for `available_commands_update`; slash menu lists agent commands as external items and inserts the token on select.
- Existing slash-command + translator suites stay green. typecheck + lint clean (one pre-existing `console` warning in `chat-instance.ts`, untouched).

---

> [!NOTE]
> **Medium Risk**
> Selecting a remote ACP agent now opens/warms a connection and session earlier than first send; routing and prompt composition for skills changed for ACP only.
> 
> **Overview**
> Surfaces **ACP agent-advertised commands** in the chat `/` slash menu before the first message, alongside user skills.
> 
> **Capture & storage:** `available_commands_update` is handled in the ACP adapter’s session router (not the per-prompt translator), mapped via `toAcpCommands` into an `onAvailableCommands` callback that writes a per-agent Zustand store (`agent-commands-store`). Adapter teardown (`disposeAdapter` / `disposeAllAdapters`) clears stored commands.
> 
> **Early session warm-up:** Adds `AgentAdapter.ensureSession` to resolve `session/new` (or load) without sending a prompt. `useWarmAcpCommands` runs when a non–built-in agent is selected so the agent can emit commands early; the first send reuses the cached session.
> 
> **UI:** Slash autocomplete uses a `SlashItem` union (skills + agent commands). Agent commands appear after skills, show an **agent-name badge**, count as valid slash tokens in the highlighter, and insert `/name` on select (no `skill_used` telemetry for commands).
> 
> **ACP user skills:** For remote ACP agents, `/slug` skill instructions are resolved in routing and passed as `skillInstructions`; the adapter **prepends** them to prompt text (ACP has no separate system channel). Built-in agents are unchanged (still resolve skills in `ai/fetch.ts`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5200c3ed6558e2c3a87d6f454e66f0f5b2a1f4a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/chats/use-warm-acp-commands.ts
```diff
@@ -0,0 +1,87 @@
...
+            if (chatThread) {
+              await updateChatThread(getDb(), chatThread.id, { acpSessionId: sessionId })
+            }
+          },
+        })
+        .catch(() => {})
+    })()
+
+    return () => {
+      cancelled = true
```
ital0, line 103: noticed the cleanup doesn't reset `warmedKey`, so a cancelled-then-remount leaves the guard set and `ensureSession` never runs. What if we clear `warmedKey.current = null` here?

src/chats/use-warm-acp-commands.ts
```diff
@@ -0,0 +1,87 @@
...
+      return
+    }
+    warmedKey.current = key
+
+    let cancelled = false
+    void (async () => {
+      const adapter = await getOrConnectAdapter(selectedAgent, {
+        httpClient,
+        getProxyFetch,
+        onAvailableCommands: (commands) => useAgentCommandsStore.getState().setCommands(selectedAgent.id, commands),
```
ital0: is this the same as `makeCommandSink` in `chat-instance.ts`? Could we reuse?

src/skills/use-slash-command.ts
```diff
@@ -85,29 +94,41 @@ export const useSlashCommand = ({
...
+    const matches = (name: string) => query === '' || name.toLowerCase().startsWith(query)
+    const skillItems: SlashItem[] = library
+      .filter((s) => isEnabled(s.name) && matches(s.name))
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .map((s) => ({ kind: 'skill', id: s.id, name: s.name, description: s.description, skill: s }))
+    const commandItems: SlashItem[] = [...agentCommands]
+      .filter((c) => matches(c.name))
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .map((c) => ({ kind: 'command', id: `command:${c.name}`, name: c.name, description: c.description }))
+    return [...skillItems, ...commandItems]
```
ital0, line 131: maybe a skill + agent command with the same name shows two `/foo` rows. Could dedupe by name if you want, low priority.

src/acp/agent-commands-store.ts
```diff
@@ -0,0 +1,29 @@
...
+ * read by the chat input's slash menu.
+ */
+type AgentCommandsState = {
+  byAgentId: Record<string, AcpCommand[]>
+  setCommands: (agentId: string, commands: AcpCommand[]) => void
+}
+
+export const useAgentCommandsStore = create<AgentCommandsState>((set) => ({
+  byAgentId: {},
+  setCommands: (agentId, commands) => set((state) => ({ byAgentId: { ...state.byAgentId, [agentId]: commands } })),
```
ital0, line 24: should we clear commands when disconnect?

ital0 (changes requested):

src/chats/use-warm-acp-commands.ts
darkbanjo: ⚡ Good catch — fixed. The cleanup now releases the guard (`warmedKey.current = null`) when teardown beats warming completion, so a same-key remount retries instead of skipping `ensureSession` forever. Tracked via a `warmed` flag set after `ensureSession` resolves.

src/chats/use-warm-acp-commands.ts
darkbanjo: ⚡ Yep, identical. Exported `makeCommandSink` from `chat-instance.ts` and reused it here, dropping the inline duplicate (and the now-unused store import).

src/skills/use-slash-command.ts
darkbanjo: ⚡ Done — deduped by name (case-insensitive), with the user skill winning so a collision shows a single `/foo` row. Added a test.

src/acp/agent-commands-store.ts
darkbanjo: ⚡ Done — added `clearCommands(agentId)` to the store and call it from `disposeAdapter`/`disposeAllAdapters`, so a disconnected agent's commands stop showing in the slash menu. Covered by a new adapter-cache test.

src/chats/use-warm-acp-commands.ts
```diff
@@ -0,0 +1,94 @@
...
+    const key = `${selectedAgent.id}:${id}`
+    if (warmedKey.current === key) {
+      return
+    }
+    warmedKey.current = key
+
+    let cancelled = false
+    let warmed = false
+    void (async () => {
+      const adapter = await getOrConnectAdapter(selectedAgent, {
```
ital0, line 72: one thing worth confirming: this now opens a live connection and creates a remote session the moment someone picks an ACP agent, even if they never type anything. For agents where starting a session has a real cost or side effect, or when a user clicks through several agents quickly, that could add up. If that tradeoff is intentional (it reads that way from the PR notes), all good. Just wanted to flag it clearly.

src/chats/use-warm-acp-commands.ts
```diff
@@ -0,0 +1,94 @@
...
+ * The warmed session is cached per thread inside the adapter, so the first send
+ * reuses it — no extra `session/new`. The commands sink matches the one the send
+ * path wires (`chat-instance.ts`), so it doesn't matter which connect wins the
+ * race for the shared per-agent connection. Built-in agents are skipped.
+ *
+ * Legitimate effect: it opens an external connection (an async on-select side
+ * effect that can't be expressed in render). Re-warming is guarded to once per
+ * (agent, thread); a failed warm clears the guard so a later render retries.
+ */
+export const useWarmAcpCommands = (session: {
```
ital0: would be great to cover with tests the main paths here. 
The logic included here is easy to break silently later.

src/skills/use-slash-command.ts
```diff
@@ -85,29 +94,44 @@ export const useSlashCommand = ({
...
+    const matches = (name: string) => query === '' || name.toLowerCase().startsWith(query)
+    const skillItems: SlashItem[] = library
+      .filter((s) => isEnabled(s.name) && matches(s.name))
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .map((s) => ({ kind: 'skill', id: s.id, name: s.name, description: s.description, skill: s }))
+    // A skill and an agent command can share a name; the skill wins so the
+    // menu doesn't show two identical `/foo` rows.
+    const skillNames = new Set(skillItems.map((s) => s.name.toLowerCase()))
+    const commandItems: SlashItem[] = [...agentCommands]
+      .filter((c) => matches(c.name) && !skillNames.has(c.name.toLowerCase()))
```
ital0, line 128: small thing on the name collision handling: when a user skill and an agent command share the same name, the skill wins and the agent command disappears from the menu entirely. That feels like a reasonable default, but it also means a user can never reach that particular agent command. 

I think it worth a quick thought on whether that is the behavior we want, or whether those rare cases deserve some way to still surface the agent one. 

Personally I faced something similar when adding MCPs and the way I solved was basically naming the subsequent MCPs with a number `mcp_1`, `mcp_2`...

ital0 (changes requested):

src/chats/use-warm-acp-commands.ts
darkbanjo: ⚡ Good call — added `src/chats/use-warm-acp-commands.test.tsx` covering the main paths: built-in skip, connect+warm on select, the once-per-(agent, thread) guard, fresh-session-id persistence via `onAcpSessionId`, the failed-connect guard release (so a later effect run retries), re-warming when the selected agent changes, and the per-agent command-sink wiring. To make the hook testable without leaking a global `mock.module()` into sibling suites (the adapter cache / DAL / db modules are imported for real by many other tests), I added a small DI seam: an optional `deps` arg that defaults to the real module functions in production.

src/chats/use-warm-acp-commands.ts
darkbanjo: ⚡ Yes, that tradeoff is intentional — warming on select is the whole point: the agent only emits its `available_commands_update` once a session exists, so without an eager connect the slash menu stays empty until after the first send. A few mitigations already bound the cost: built-in agents are skipped entirely, the warm is guarded to once per (agent, thread) so clicking through agents doesn't re-warm the same one, the connection is shared per-agent via the adapter cache (one transport, reused across threads), and the warmed session is reused by the first real send (no extra `session/new`). The remaining cost is one `session/new` per distinct agent the user lands on. If that proves heavy for a specific agent we could debounce the warm behind a short dwell timer, but I'd rather not add that complexity speculatively.

src/skills/use-slash-command.ts
darkbanjo: ⚡ Considered this, and I think skill-wins is the right default for v1: it keeps the menu deterministic (no duplicate /foo rows) and matches the principle that the user's own authored skills take precedence over an agent's advertised commands. The collision is genuinely rare — it requires a user skill and an agent command to share an exact name. I'd rather not adopt the mcp_1/mcp_2 numeric-suffix scheme here because it changes the slug the user types (and the agent advertises), which would be surprising in an autocomplete menu. If this matters in practice, the cleaner fix is to badge the colliding agent command and disambiguate by source rather than mangling the name — happy to revisit if you feel strongly. Left as-is for now.

PR #985, opened 2026-06-15
feat(THU-448): App version management: track device versions and enforce minimum version (BACKEND)

arienemaiara:
> [!NOTE]
> **Medium Risk**
> Adds a nullable DB column and changes PowerSync token/device upsert behavior; misconfigured MIN_APP_VERSION fails startup but enforcement impact depends on the frontend companion change.
> 
> **Overview**
> Adds **minimum app version** enforcement plumbing on the backend: `MIN_APP_VERSION` is read from the environment, validated as semver at startup, and returned from public **GET `/config`** as `minAppVersion` (omitted when unset so clients treat it as no enforcement).
> 
> Devices gain an **`app_version`** column (migration `0020`). On **PowerSync token** issuance, the server reads **`X-App-Version`** (trimmed, max 64 chars), stores it via `upsertDevice`, and ignores oversized values. Clients cannot set **`app_version`** through PowerSync upload PATCH—it is listed in upload deny columns alongside trust/crypto fields.
> 
> Tests cover config exposure, settings validation, token persistence, header omission, long header rejection, and upload stripping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 240d0c7e5fd270cbc3ac605840f947bc60eb7e74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


backend/src/config/settings.ts
```diff
@@ -92,6 +92,10 @@ const settingsSchema = z
     // E2E encryption — when true, devices must complete the trust flow before syncing
     e2eeEnabled: z.boolean().default(false),
 
+    // Minimum app version clients must run. Empty string disables enforcement.
+    // Surfaced to the frontend via GET /config; clients below this hard-block until they update.
+    minAppVersion: z.string().default(''),
```
ital0: Would it make sense to validate that `MIN_APP_VERSION` is actually semver before the app boots? Right now a typo like `0,2,0` or `banana` starts up fine and gets served to every client, where the frontend will then choke trying to compare it. Since this file already uses `superRefine` for the PowerSync secret check, a quick regex guard there would fail fast at startup and save operators some painful debugging down the line.

backend/src/api/config.ts
```diff
@@ -18,4 +18,6 @@ export const createConfigRoutes = (settings: Settings) =>
     // contract reads as a positive capability ("enabled").
     builtInAgentEnabled: !settings.disableBuiltInAgent,
     allowCustomAgents: settings.allowCustomAgents,
+    // Omit when unset so the frontend treats it as "no enforcement" without parsing an empty string as semver.
+    minAppVersion: settings.minAppVersion || undefined,
```
ital0, line 22: The `|| undefined` correctly omits empty strings, but a whitespace only value like `MIN_APP_VERSION=' '` would slip through as truthy and reach every client as a padded string. Trimming either here or in the settings parser would harden this nicely. No test covers this edge case today either.

backend/src/api/powersync.ts
```diff
@@ -102,6 +102,9 @@ const issuePowerSyncToken = async (
   const rawDeviceName = request.headers.get('x-device-name')?.trim()
   const deviceName =
     rawDeviceName && rawDeviceName.length > 0 && rawDeviceName.length <= 100 ? rawDeviceName : 'Unknown device'
+  const rawAppVersion = request.headers.get('x-app-version')?.trim()
+  // Cap to a sane length so a malformed/oversized header never bloats the row.
+  const appVersion = rawAppVersion && rawAppVersion.length > 0 && rawAppVersion.length <= 32 ? rawAppVersion : undefined
```
ital0: One thing worth a quick thought: a full semver with prerelease and build metadata can exceed 32 characters (something like `1.234.567-rc.1+build.2024.01.01` is 35). When that happens the value gets silently dropped and the prior version is kept, which could confuse operators looking at device rows. If rich semver is on the roadmap for the frontend, bumping the cap or documenting why 32 was chosen would help future readers. Totally fine to defer if build metadata is never expected.

ital0 (changes requested):

backend/src/db/powersync-schema.ts
```diff
@@ -241,6 +241,7 @@ export const devicesTable = powersyncSchema.table(
     lastSeen: timestamp('last_seen').defaultNow(),
     createdAt: timestamp('created_at').defaultNow(),
     revokedAt: timestamp('revoked_at'),
+    appVersion: text('app_version'),
```
ital0, line 244: Quick confirm: since `devices` syncs `SELECT *`, `app_version` reaches clients but isn't declared in the frontend schema (`src/db/tables.ts`), so PowerSync just keeps it internally. Reads as a deliberate server side field, all good. If you ever want it in the device list UI, the follow up is adding `appVersion` to the frontend `devicesTable`.

backend/src/api/config.ts
ital0, line 22: Confirming intent: enforcement is client side only (the upgrade screen), and the backend still issues sync tokens and accepts uploads from clients below the floor. Reads as a UX nudge by design, so fine. If old clients ever need a hard block, that check would live on `/powersync/token`.

ital0 (approved):
Re-reviewed after the fixes and this all looks great. A few small optional notes below, nothing blocking.

backend/src/db/powersync-schema.ts
arienemaiara: Confirmed — that's the intent. `app_version` is server-managed (written from `X-App-Version` on `/v1/token`, denied on PowerSync uploads via `uploadDenyColumns`), so it lives on the synced `devices` table for operator visibility without needing client-side reads. Surfacing it in the device list UI is a natural follow-up: declare `appVersion` on `src/db/tables.ts`'s `devicesTable` and it'll start landing in the SQLite store. Happy to file a ticket if useful.

backend/src/api/config.ts
arienemaiara: Confirmed — client-side UX nudge by design. The original ticket scoped this as "hard-block the app with a non-dismissable modal/screen … no access to chat, settings, or sync until they upgrade," which the upgrade screen + provider tree short-circuit covers from the user's side. Backend-side hard block on `/v1/token` (and likely `/v1/powersync/upload`) is the right place if/when we want server-enforced refusal — happy to file a follow-up if that becomes a requirement.

PR #986, opened 2026-06-15
feat(THU-448): App version management: track device versions and enforce minimum version (FRONTEND)

arienemaiara:
> [!NOTE]
> **Medium Risk**
> A misconfigured server `minAppVersion` can hard-block all users until they upgrade; comparison is defensive so bad version strings should not trigger a block.
> 
> **Overview**
> Adds **client-side minimum version enforcement** aligned with the backend `MIN_APP_VERSION` / `GET /config` `minAppVersion` field. The build now injects **`VITE_APP_VERSION`** from `package.json`, and that value is sent on app-backend requests as **`X-App-Version`** (HTTP client and `getAuthenticatedHeaders`).
> 
> When `minAppVersion` is set and the installed version compares lower via new **`compareSemver`** (malformed semver pairs do not block), **`App`** short-circuits to a full-screen **`UpgradeRequired`** message before normal init/routing, and re-evaluates when the config store updates after `fetchConfig`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 233148d514eeafbf92d8f2c67d464d519c81ece9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Evidences
Run the backend with `MIN_APP_VERSION=0.1.103 bun dev` where the current app version is `0.1.102`

image


src/lib/compare-semver.ts
```diff
@@ -0,0 +1,40 @@
...
+    const nums = parts.map((p) => Number(p))
+    if (nums.some((n) => !Number.isInteger(n) || n < 0)) {
+      return null
+    }
+    return [nums[0], nums[1], nums[2]]
+  }
+
+  const av = parse(a)
+  const bv = parse(b)
+  if (!av || !bv) {
```
ital0, line 30: Returning 0 so malformed values never brick the app is a nice safeguard. One angle worth weighing: since this is an enforcement gate, a malformed `minAppVersion` coming from the server would silently disable the floor for every client rather than failing closed. Would it make sense to fail closed only when the server value is unparseable, but keep failing open when the client's own version can't be parsed? No change needed if the current fail open policy is the intended call, just wanted to raise the security side.

src/vite-env.d.ts
```diff
@@ -9,6 +9,7 @@ interface ImportMetaEnv {
   readonly VITE_THUNDERBOLT_CLOUD_URL?: string
   readonly VITE_AUTH_MODE?: 'thunderbolt' | 'sso'
   readonly VITE_AUTH_ENABLE_ANONYMOUS?: 'true' | 'false'
+  readonly VITE_APP_VERSION: string
```
ital0: Small consistency note: `VITE_APP_VERSION` is typed as a required `string`, but the guards in `auth-token.ts` and `http.ts` treat it as possibly absent, which is actually true under Bun tests where the Vite `define` doesn't run. Typing it as `string | undefined` would match how it's really used. Tiny thing, feel free to skip.

ital0 (changes requested):

ital0 (dismissed):

ital0 (approved):

PR #991, opened 2026-06-16
feat(THU-596): user data export & import

raivieiraadriano92:
## Summary

- Adds **Export My Data** and **Import Data** actions to Settings → Preferences → Data: one-click versioned JSON snapshot of every user-content table, plus restore via SELECT + UPDATE/INSERT inside a single transaction (works on PowerSync's view-backed synced tables, where `INSERT ... ON CONFLICT DO UPDATE` is rejected).
- File format is versioned (`schemaVersion: 1`); v1 ships without a workspace concept so it can land before Workspaces v1. The post-workspaces refactor (THU-597) bumps to `schemaVersion: 2` with `workspaceId` on rows and no selector UI — spec in `docs/architecture/export-format.md`.
- Schema-driven: both halves walk `includedTables` derived from `src/db/powersync/schema.ts`; adding a new synced/local-only table flows through automatically with the export's allowlist test as the safety net.

## Test plan

- [ ] `bun test src/dal/export.test.ts src/dal/import.test.ts src/lib/export-download.test.ts src/lib/import-upload.test.ts` all green (~40 tests)
- [ ] `bun tsc --noEmit`, `bun lint`, `bun run format-check` all clean
- [ ] Sign in, seed chats / tasks / a custom model with an API key → **Export My Data** downloads `thunderbolt-export-YYYY-MM-DD.json` with `schemaVersion: 1` and the expected table buckets
- [ ] Open the file: `tables.integrations_secrets` / `tables.devices` / `tables.agents_system` absent; `models_secrets` contains the typed API key
- [ ] Sign out → sign back in (fresh DB) → **Import Data** → pick the file → confirmation dialog shows row count + source email + export date → confirm → success message, chats / tasks / models all restored
- [ ] Re-import on top of the populated DB → imported file wins on overlapping IDs, no duplicates; non-overlapping local rows untouched
- [ ] Malformed file (`{}` or wrong `schemaVersion`) → inline error, no DB writes; tampered file containing `devices` / `integrations_secrets` lands in `ignoredTableNames`

---

> [!NOTE]
> **High Risk**
> Backup files hold plaintext API keys and MCP/agent secrets; import overwrites matching IDs locally and can propagate those changes to other devices via PowerSync.
> 
> **Overview**
> Adds **Export My Data** and **Import Data** under Settings → Preferences → Data for signed-in users, backed by a new `schemaVersion: 1` `thunderbolt-export` envelope.
> 
> **Export** snapshots local SQLite rows from a schema-driven table list (`syncedTables` + `localOnlyTables`, minus `devices`, `integrations_secrets`, and `agents_system`), then downloads **gzipped JSON** (`thunderbolt-export-YYYY-MM-DD.json.gz`). **Import** validates the envelope, previews row count / source email / cross-account mismatch, then restores in one transaction using **SELECT + UPDATE or INSERT** (avoiding upserts on PowerSync views). Imported rows win on PK collision; `userId` on synced tables is always taken from the session; unknown table keys are ignored.
> 
> Supporting pieces: `readJsonFile` (gzip by magic bytes, 200 MB cap), architecture spec in `docs/architecture/export-format.md`, DAL re-exports, and PostHog `settings_data_export` / `settings_data_import` events. Broad unit test coverage for export, import, download, and upload helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21a59dee7a5de7d2d1b1fa22e2cf298a24357b8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/dal/import.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (!isRecord(row)) {
+          throw new ImportFormatError(`Row in table "${tableName}" is not an object.`)
+        }
+        const pkValue = row[pkField]
+        if (typeof pkValue !== 'string') {
+          throw new ImportFormatError(`Row in "${tableName}" is missing string primary key "${pkField}".`)
+        }
+        const existing = await tx.select({ pk: pkColumn }).from(table).where(eq(pkColumn, pkValue)).get()
+        if (existing) {
+          await tx.update(table).set(row).where(eq(pkColumn, pkValue))
```
ital0: **[BLOCKING] architecture / security** — Imported rows are written VERBATIM into synced tables, bypassing every DAL write path and all per-row invariants (createChatThread's isEncrypted stamp, derived columns) and accepting attacker-controlled columns. Only isRecord(row)+string-PK is validated; userId/timestamps/any column are mass-assignable from a tampered/cross-account file. The source file's user_id is written straight into local synced views; the backend re-stamps user_id on upload (INV-01), so cross-tenant *server* writes are blocked, but the LOCAL row keeps the foreign userId until reconciliation and local-only secret tables have no such gate. This is a second, schema-agnostic write path that sidesteps DAL invariant enforcement.

**Fix:** Validate/whitelist per-table importable columns before writing: strip server-owned/identity/derived columns (user_id, createdAt/updatedAt, encryption/defaultHash) and re-stamp userId to the current session user. Document why a generic verbatim upsert is safe for synced tables; confirm with powersync-sync-reviewer that writing user_id on import is correct under INV-01.

_rule: architecture — write path bypassing DAL invariant enforcement; H mass-assignment of untrusted input; INV-01 user_id from JWT (review-heuristics §2)_

src/dal/import.ts
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+          await tx.insert(table).values(row)
```
ital0: **[BLOCKING] powersync** — Importing into SYNCED tables is not a passive local snapshot restore: every upserted row (settings/chat_threads/chat_messages/tasks/models/model_profiles/prompts/skills/triggers/modes/agents) is captured into ps_crud and uploaded via connector.uploadData, fanning out to all the user's other devices. Nothing in import.ts, export-format.md, or the confirm dialog states this — the dialog only warns about 'this device'. Restoring an old backup silently overwrites newer data on every other device.

**Fix:** Document the synced-upload consequence in export-format.md (Import behavior) and the confirm-dialog copy: importing synced tables propagates to the backend and all devices. If unintended, scope import to local-only tables or use a non-replicating write path.

_rule: powersync-sync-reviewer Checklist #7 / silent cross-device effect_

src/dal/import.ts
```diff
@@ -0,0 +1,164 @@
...
+
+      for (const row of rows) {
+        if (!isRecord(row)) {
+          throw new ImportFormatError(`Row in table "${tableName}" is not an object.`)
+        }
+        const pkValue = row[pkField]
+        if (typeof pkValue !== 'string') {
+          throw new ImportFormatError(`Row in "${tableName}" is missing string primary key "${pkField}".`)
+        }
+        const existing = await tx.select({ pk: pkColumn }).from(table).where(eq(pkColumn, pkValue)).get()
```
ital0, line 249: **[BLOCKING] powersync / encryption** — Exported chat_messages content/parts/metadata are PLAINTEXT in the file (per export-format.md Encryption section). On import with E2EE enabled these rows ride back up the sync layer; encodeForUpload only encrypts string columns, and if the importing device has no Content Key yet (needsSyncSetupWizard true) encoding fails or uploads in a state inconsistent with the device's key. Importing a plaintext backup into an E2EE account is an untested correctness/security path.

**Fix:** Before importing into synced tables when isEncryptionEnabled(), assert a Content Key exists (getCK / needsSyncSetupWizard) and block/guide the user through the sync setup wizard first. Add a test importing an E2EE-content envelope asserting the uploaded payload is encrypted.

_rule: powersync-sync-reviewer Checklist #4 (encryption) + e2e-encryption.md_

src/settings/preferences.tsx
```diff
@@ -284,6 +319,108 @@ export default function PreferencesSettingsPage() {
...
+      // Light envelope shape check so we can show source attribution + counts
+      // in the confirmation dialog. Full validation runs inside `importUserData`.
+      if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+        throw new ImportFormatError("This file doesn't look like a Thunderbolt export.")
+      }
+      const obj = payload as Record<string, unknown>
+      if (obj.format !== 'thunderbolt-export') {
+        throw new ImportFormatError("This file doesn't look like a Thunderbolt export.")
+      }
+      const totalRows = countTablesRows(obj.tables)
```
ital0: **[BLOCKING] architecture** — The component reimplements a partial, parallel copy of the envelope validation that canonically lives in import.ts (isRecord + format check + tables shape), plus module-level envelope parsers (countTablesRows, formatExportedAt, source-email extraction). The two validators can drift (import.ts also checks schemaVersion; the component does not). Envelope shape is domain knowledge owned by the export/import DAL; spreading it into the settings presentation layer is a coupling/duplication smell.

**Fix:** Move envelope-preview extraction (format gate + countTablesRows + formatExportedAt + sourceEmail) into the DAL as summarizeExportEnvelope(payload): ExportSummary | null returning typed {totalRows, exportedAtLabel, sourceEmail}; call it from the component. Keep all envelope-shape logic in the DAL.

_rule: architecture — parallel validation path vs canonical; presentation reaching into data format; logic belonging in DAL (review-heuristics §1/§2)_

docs/architecture/export-format.md
```diff
@@ -0,0 +1,104 @@
...
+- **Soft-deleted rows preserved.** `deletedAt` rides through verbatim — a row in the trash in the file remains in the trash after import.
+- **Atomic.** The whole import runs inside a single `db.transaction`. A single row-level failure (e.g. a constraint violation) rolls back every preceding write in the call.
+- **Forward-compatible.** Table keys in the file that the importer doesn't recognize (e.g. tables added in a future `schemaVersion`) are surfaced in `ignoredTableNames` and otherwise skipped silently. Unknown columns within a recognized table are dropped by SQLite if they don't match the schema.
```
ital0: **[CONVENTION] docs-intent** — Forward-compat bullet claims 'Unknown columns within a recognized table are dropped by SQLite if they don't match the schema.' This is wrong for this importer: Drizzle builds the SQL column list from the row's own keys, so an unknown key produces a SQLite 'no column named Y' error that aborts the whole single-transaction import — not silently dropped. The doc gives a false sense of forward-compat safety for v2+ files that add columns.

**Fix:** Correct the wording: extra/unknown columns on a recognized table throw and roll back the entire import. Or, if silent-drop is intended, change import.ts to strip unknown keys before writing and keep the doc.

_rule: docs-intent accuracy (review-heuristics §4)_

src/settings/preferences.tsx
```diff
@@ -284,6 +319,108 @@ export default function PreferencesSettingsPage() {
...
+    setImportSuccess(null)
+    try {
+      const payload = await readJsonFile(file)
+      // Light envelope shape check so we can show source attribution + counts
+      // in the confirmation dialog. Full validation runs inside `importUserData`.
+      if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+        throw new ImportFormatError("This file doesn't look like a Thunderbolt export.")
+      }
+      const obj = payload as Record<string, unknown>
+      if (obj.format !== 'thunderbolt-export') {
```
ital0: **[CONVENTION] architecture** — Light envelope pre-check hardcodes the literal obj.format !== 'thunderbolt-export' instead of importing the canonical exportFormat const (src/dal/export.ts, re-exported from @/dal). preferences.tsx already imports from @/dal, so the format slug now lives in two places; a future rename silently breaks this guard while the real validator keeps working.

**Fix:** Import exportFormat from '@/dal' and compare obj.format !== exportFormat. Single-source the format string the way import.ts does.

_rule: single-source config (review-heuristics §1/§4)_

src/dal/export.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Exported so the importer ({@link src/dal/import.ts}) can iterate the same
+ * set without redefining it.
+ */
+export const includedTables = Object.fromEntries(
+  Object.entries({
+    ...syncedTables,
+    ...Object.fromEntries(Object.entries(localOnlyTables).map(([name, def]) => [name, def.tableDefinition])),
+  }).filter(([name]) => !isExcluded(name)),
+) as unknown as Record<IncludedTableName, SQLiteTable>
```
ital0, line 52: **[CONVENTION] conventions** — `as unknown as Record<IncludedTableName, SQLiteTable>` double-launders the type. R-NOANY flags `as unknown as`. Object.fromEntries already yields Record<string, SQLiteTable>; only the key union is being narrowed, which a single `as` covers.

**Fix:** Drop the `as unknown` and use a single `as Record<IncludedTableName, SQLiteTable>`. If it doesn't type-check, add a one-line comment naming the exact structural mismatch forcing the double cast.

_rule: R-NOANY_

src/settings/preferences.tsx
```diff
@@ -284,6 +319,108 @@ export default function PreferencesSettingsPage() {
   }
 
   const [deleteAccountError, setDeleteAccountError] = useState<string | null>(null)
+  const [exportError, setExportError] = useState<string | null>(null)
```
ital0: **[CONVENTION] conventions / react** — Export/import state is split: loading flags (isExporting/isImporting) live in preferencesReducer, but exportError/importError/importSuccess/pendingImport were added as 4 standalone useState hooks. The component now carries the reducer plus 6 useState calls. R-REDUCER/R-STATEHOOK favor consolidating related state, especially since the loading siblings already live in the reducer.

**Fix:** Move exportError/importError/importSuccess/pendingImport into preferencesReducer, or extract a useDataExportImport() state hook so the concern is unit-testable and the state surface stays cohesive.

_rule: R-REDUCER / R-STATEHOOK_

src/lib/import-upload.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Reads a `File` (from an `<input type="file">`) as text and parses it as
+ * JSON. Returns `unknown` so the caller validates the shape before using it.
+ *
+ * Throws a `SyntaxError` with the original file name in the message when the
+ * JSON doesn't parse — easier to surface in the UI than the raw parser
+ * error.
+ */
+export const readJsonFile = async (file: File): Promise<unknown> => {
+  const text = await file.text()
```
ital0: **[CONVENTION] security / perf** — No size or row-count bound anywhere on the import path: readJsonFile does file.text()+JSON.parse on the whole user-chosen file with no max, then importUserData iterates every row with no cap. A multi-hundred-MB or millions-of-rows file loads fully into memory and parses on the main thread, freezing/OOM-ing the WebView (Tauri iOS especially). The file is from an untrusted <input type=file>.

**Fix:** Add a file-size guard before reading (reject above a sane cap) and/or a total-row cap in importUserData; surface a clear ImportFormatError when exceeded.

_rule: H untrusted-input DoS / unbounded memory on import_

src/settings/preferences.tsx
```diff
@@ -737,6 +874,83 @@ export default function PreferencesSettingsPage() {
...
+            <>
+              <div className="h-px bg-border -mx-6" />
+
+              <div className="flex flex-col gap-2">
+                <label htmlFor="export-data-button" className="text-sm font-medium">
+                  Export Your Data
+                </label>
+                <p id="export-data-description" className="text-sm text-muted-foreground">
+                  Download a JSON snapshot of your chats, tasks, models, and the API keys you've entered. Treat the file
+                  like a password backup — anyone who can read it can spend your model credits. Google and Microsoft
```
ital0: **[CONVENTION] security / docs-intent** — Export bundles mcp_secrets (MCP bearer/API/OAuth tokens) and agents_secrets (ACP agent keys) in plaintext, but the Export warning only says 'anyone who can read it can spend your model credits.' MCP/agent bearer & OAuth tokens grant direct third-party access far beyond model credits. Inconsistent stance: integrations_secrets (Google/MS OAuth) is excluded while MCP OAuth tokens are included.

**Fix:** Broaden the warning to cover MCP/agent credentials (bearer/OAuth tokens), e.g. 'anyone who can read it can use any API keys and tokens you've entered.' Or reconsider excluding mcp_secrets/agents_secrets OAuth-type creds like integrations_secrets, applying one consistent rule to all three secret tables.

_rule: H credential disclosure — exported secrets exceed user warning_

src/dal/export.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Tables intentionally excluded from the user-data export:
+ * - `devices`: per-device trust state, meaningless on the importing device.
+ * - `integrations_secrets`: third-party OAuth tokens (Google / Microsoft).
+ *   The importing user re-authenticates instead.
+ * - `agents_system`: backend-hydrated catalog, not user content.
+ *
+ * Typed against {@link AllTableName} so a future schema rename/removal trips
+ * the compiler. Anything not in this list is included.
+ */
+const excludedFromExport = [
```
ital0, line 25: **[CONVENTION] architecture / powersync** — includedTables mixes SYNCED and LOCAL-ONLY tables into one flat set with no classification carried through; excludedFromExport is a hand-maintained allowlist-by-exclusion that mixes three rationales (per-device devices, third-party integrations_secrets, backend agents_system) while OTHER secret tables (models_secrets/mcp_secrets/agents_secrets) are intentionally INCLUDED. On import the synced siblings upload while local-only secrets stay local — asymmetric propagation never documented (config syncs out, key stays local).

**Fix:** Document the asymmetric import behavior in export-format.md (synced tables propagate on import, local-only secrets do not) and confirm intended. Consider deriving exclusion from a per-table portableInBackup classification colocated with the schema rather than a separate hand-maintained list.

_rule: single-source config / sibling-table classification consistency (review-heuristics §2; INV-68; powersync Checklist #7)_

src/dal/import.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (existing) {
```
ital0, line 250: **[CONVENTION] powersync / privacy** — Cross-account restore: import writes the source file's user_id into local synced views; backend re-stamps to the authenticated user, but the LOCAL row keeps the foreign user_id until down-sync reconciles. A backup whose PK collides with another user's cloud row is silently dropped server-side (onConflictDoNothing, importer-scoped WHERE userId) while the local view shows it imported — local/cloud diverge with no error. Secret tables (models_secrets/mcp_secrets/agents_secrets) have no user_id, so another account's raw keys/tokens land on this device attributed to whatever parent id collides.

**Fix:** Either restrict import to same-account files (compare envelope.user.id to session user id, warn on mismatch — sourceEmail is already partially extracted) or document cross-account import as unsupported with transient mis-attribution. Add a test importing a row whose user_id != current user.

_rule: powersync-sync-reviewer Checklist #5/#7 / data-isolation + privacy_

ital0 (commented):
🤖 **thunder-deep-review** — advisory (pending; submit or discard after review).

CONCERNS — core feature works, but import writes untrusted/verbatim rows into synced tables, bypassing DAL invariants and silently propagating cross-device (and potentially cross-account/E2EE) with no size/column guards; several blocking architecture+powersync+security items plus a wrong forward-compat doc claim should be resolved before merge.

**5 blocking · 7 convention · 17 nit**

All 5 blocking + 7 convention findings are attached as inline comments. The 17 nits:

- `src/dal/import.ts:147-152` — N+1 write pattern: per-row SELECT then UPDATE/INSERT, serial inside one tx; batch-load existing PKs with a single IN(...) SELECT into a Set
- `src/dal/import.ts:149` — UPDATE writes set(row) including the PK column; omit pk from the SET payload
- `src/dal/export.ts:91-93` — Export materializes every row of every table via Promise.all then pretty JSON.stringify (~2x); large transient memory spike, no streaming/pagination
- `src/settings/preferences.tsx:376` — handleConfirmImport clears importError but never importSuccess; stale success banner + new error can render together
- `src/settings/preferences.tsx:877` — Export/Import UI gated on isAuthenticated (anonymous too) though component derives isFullUser; copy implies a connected account; gate vs copy vs cross-user provenance
- `src/dal/import.ts:132` — Recognized table whose value is a non-array is silently skipped by the empty-array branch (not counted, not ignored); rows-not-object inside a valid array throw — inconsistent contract
- `src/settings/preferences.tsx:939` — aria-describedby resolves to the success element id after import, so the button is permanently described by the last result instead of its instructional description
- `src/settings/preferences.tsx:362` — Inline 7-line sourceEmail type-guard ternary breaks symmetry with its extracted siblings; double cast appears twice — extract a readSourceEmail helper
- `src/dal/import.ts:39` — Single-use candidates const split from the .find on the next line; collapse into one chained expression
- `src/db/tables.ts:315` — Comment on agentsSecretsTable says 'Never leaves the device' but this PR's export now includes agents_secrets — invariant no longer true (NOT in diff → body)
- `docs/architecture/export-format.md:82` — Grammar: 'a `ImportFormatError`' → 'an `ImportFormatError`'
- `docs/architecture/export-format.md:87` — Cites multi-device-sync.md for 'no FK constraints' but that doc only says 'No composite foreign keys'; weaken wording or drop citation
- `src/settings/preferences.tsx:1448` — Dialog 'contains {totalRows} rows' counts every bucket (incl. excluded/unknown) while success toast 'Imported {total} rows' sums only upserted — numbers silently disagree
- `src/dal/export.ts:13` — Widening syncedTables/localOnlyTables const→export to feed export.ts removes a guardrail; add a comment marking them the canonical classification source
- `docs/architecture/export-format.md:44-46` — New SYNCED table is auto-included in export AND import-with-upload with zero sync-classification review; note + import test to mirror the export allowlist gate
- `src/dal/import.test.ts:284-292` — derivePkSpec tests suppress console.warn with a hand-rolled save/restore; use spyOn(console,'warn').mockImplementation(()=>{}) per R-SUPPRESSCONSOLE
- `src/settings/preferences.tsx:67-68` — New reducer actions modeled as setters (SET_IS_EXPORTING) not events; R-REDUCER prefers EXPORT_STARTED/FINISHED (low priority, file is locally consistent)

_Note: the `src/db/tables.ts:315` nit is outside the PR diff, so it could not be anchored inline — listed above only._

src/dal/import.ts
raivieiraadriano92: Fixed in ee5b565b.

src/dal/import.ts
raivieiraadriano92: Fixed in 6466d07e.

src/dal/import.ts
raivieiraadriano92: False positive - `handleSignInSuccess` in `src/contexts/sign-in-modal-context.tsx` forces the sync setup wizard before `setSyncEnabled(true)` runs when `e2eeEnabled` is on, and `useSyncEnabledToggle` auto-disables sync for legacy pre-encryption users. Import writes can only reach `encodeForUpload` after a CK exists, so the failure mode the gate would catch is unreachable.

src/settings/preferences.tsx
raivieiraadriano92: Fixed in [b30c66c6](https://github.com/thunderbird/thunderbolt/commit/b30c66c6).

docs/architecture/export-format.md
raivieiraadriano92: Fixed in [c3eff284](https://github.com/thunderbird/thunderbolt/commit/c3eff284).

src/settings/preferences.tsx
raivieiraadriano92: Fixed in [b30c66c6](https://github.com/thunderbird/thunderbolt/commit/b30c66c6) — the inline block was replaced by `summarizeExportEnvelope` in the DAL, which uses the canonical `exportFormat` constant.

src/dal/export.ts
raivieiraadriano92: Tried the single cast first — TS rejected it (`{ [k: string]: SQLiteTable }` can't prove each named key of `Record<IncludedTableName, SQLiteTable>` is present, so the compiler itself suggests `unknown` first). Kept the double cast and added the explanatory comment in [8569e3ad](https://github.com/thunderbird/thunderbolt/commit/8569e3ad).

src/settings/preferences.tsx
raivieiraadriano92: Fixed in [cabbe8b9](https://github.com/thunderbird/thunderbolt/commit/cabbe8b9).

src/lib/import-upload.ts
raivieiraadriano92: Fixed in [6a91e40f](https://github.com/thunderbird/thunderbolt/commit/6a91e40f) — files larger than 200 MB are now rejected before `file.text()` runs.

src/settings/preferences.tsx
raivieiraadriano92: Fixed in [dae5938c](https://github.com/thunderbird/thunderbolt/commit/dae5938c) — warning now names model API keys, MCP server tokens, and agent keys, and frames the file as a password vault rather than just model-credit exposure. The asymmetric MCP-vs-Google/MS OAuth treatment is intentional for v1 and out of scope for this PR; we can revisit excluding `mcp_secrets`/`agents_secrets` separately.

src/dal/export.ts
raivieiraadriano92: Fixed in [f890b89d](https://github.com/thunderbird/thunderbolt/commit/f890b89d) — the Cross-device-propagation bullet in `export-format.md` now states explicitly that local-only secret tables stay local on import. Leaving the classification refactor (per-table `portableInBackup` colocated with the schema) out of scope for v1 — the current allowlist-by-exclusion is small enough to maintain by hand, and the doc + the type-checked `AllTableName` constraint on `excludedFromExport` catch schema renames at compile time.

src/dal/import.ts
raivieiraadriano92: Fixed in [2f5b997a](https://github.com/thunderbird/thunderbolt/commit/2f5b997a) — `summarizeExportEnvelope` now compares `envelope.user.email` against the session email (case-insensitive) and exposes an `accountMismatch` flag; the confirm dialog renders a destructive-styled warning when it's set. Using email rather than `user.id` so the legitimate delete-and-recreate-with-same-email recovery flow isn't blocked. Soft warn rather than hard block — the dialog already requires explicit destructive confirmation, and a hard block would lock out the recovery flow if better-auth ever rotates the `user.id` for the same human. The synced-tables half of cross-account leakage is closed by the earlier `userId` re-stamp ([ee5b565b](https://github.com/thunderbird/thunderbolt/commit/ee5b565b)).

darkbanjo (dismissed):

darkbanjo (approved):

PR #1001, opened 2026-06-18
feat: add generalized quiz widget (THU-607)

darkbanjo:
## What

Brings the interactive quiz widget prototyped on the OUP demo branch into the product as a first-class, demo-agnostic widget.

- Adds `src/widgets/quiz`: `single` / `multiple` / `choice` modes, client-side grading, answer persistence in the message cache, and restore-on-reload. Includes schema, parser, instructions, the presentational `Quiz`, stories, and lib tests.
- Registers `quiz` in the widget registry and the cache-data union.
- Surfaces persisted answers to the model via `formatQuizResultsNote`, injected as a system note in `aiFetchStreamingResponse`, so the model can report scores without asking the user to re-enter their choices.

The model emits `<widget:quiz mode="..." prompt="..." options='[...]' explanation="..." />`.

## Decision: hidden-turn machinery left behind

The demo branch also had a `hiddenFromUi` / `ChatTurnActionsProvider` mechanism for dispatching a hidden user turn on answer submit. The quiz **grades client-side and reports via the system note instead** — and on the demo branch that machinery was wired into the provider tree but **never actually invoked**. Porting it here would be dead code, so it's intentionally excluded. It can be introduced on its own if a real hidden-turn use case appears.

## Test

- `src/widgets/quiz/lib.test.ts`: 9/9 pass (grading across modes, cache round-trip, results-note formatting). Typecheck clean.

Closes THU-607

---

> [!NOTE]
> **Medium Risk**
> Changes the chat streaming path (extra system context and conditional per-message DB reads) and auto-dispatches user turns from widget submit for choice/free modes, though behavior is guarded and covered by tests.
> 
> **Overview**
> Adds a first-class **`ask`** widget (`<widget:ask … />`) so the model can render inline prompts instead of markdown option lists. It supports **`single` / `multiple`** (client-side designated-answer checking with neutral post-submit copy), **`choice`** (immediate commit, no grading), and **`free`** (typed answers).
> 
> Responses are **persisted on the assistant message cache** (stable keys per prompt + shape), **restored on reload**, and for **`choice` / `free`** optionally **sent as a normal user message** so the assistant can continue; graded modes intentionally do not auto-send to avoid quiz loops.
> 
> **`aiFetchStreamingResponse`** now prepends a **“User responses” system note** built from cached ask entries when the thread contains an ask tag (skipping DB reads otherwise), alongside existing skill system messages. **`assistant-message`** treats ask like other rich widgets for copy-button layout.
> 
> The widget is registered in **`widgetRegistry`** (parser, instructions, cache union) with display/component wiring, stories, and unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a61c5dd28e14376044fc26412a8a86524308d4c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/widgets/ask/widget.tsx
```diff
@@ -0,0 +1,69 @@
...
+    // `free` mode carries a typed answer; option modes map ids back to their texts.
+    const chosen = text !== undefined ? [text] : selectedIds.map((id) => options.find((o) => o.id === id)?.text ?? id)
+    const entry: AskCacheEntry = { prompt, mode, selectedIds, chosen, matched, text }
+    // Persist the response so the widget restores answered on reload. We
+    // intentionally do NOT dispatch a follow-up turn: the widget reveals any
+    // designated answer client-side, so a prompt stands on its own, and the
+    // persisted entry is surfaced to the model on later turns via
+    // formatAskResponsesNote. Sending a turn per response would also goad
+    // single-prompt-at-a-time backends into endlessly asking the next one.
+    await updateMessageCache(db, messageId, storageKey, entry)
```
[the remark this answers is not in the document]
darkbanjo: Fixed in ffc7e508 — `handleSubmit` now calls `queryClient.setQueryData(queryKey, entry)` after `updateMessageCache` resolves, so the infinitely-cached `askState` query stays in sync. An unmount/remount now restores the saved answer instead of serving the stale initial `null`.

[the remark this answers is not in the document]
darkbanjo: The UI side is closed by the cache-sync fix in ffc7e508. The remaining model-read race is genuine but very low probability — it needs a local SQLite write to lose to the user immediately sending a new turn. Fully closing it would mean awaiting persistence through the chat-send path; opted against that complexity for the likelihood. Leaving as intentional.

src/widgets/ask/lib.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/** Maps an option index to its display letter (0 → "A", 1 → "B", ...). */
+export const optionLetter = (index: number): string => String.fromCharCode(65 + index)
+
+/** Namespace prefix for ask entries stored in a message's cache blob. */
+export const askCachePrefix = 'ask'
+
+/** Cache key for a single ask instance within a message (one tag = one prompt). */
+export const askStorageKey = (prompt: string): string => `${askCachePrefix}/${prompt}`
```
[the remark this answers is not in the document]
darkbanjo: Fixed in 8abdafb1 — `askStorageKey` now keys on the prompt **plus** a stable djb2 hash of `mode`+`options`, so two ask widgets in one message that share a prompt but differ in their options no longer collide on `ask/<prompt>` and overwrite each other's answer. Two byte-identical widgets still share a key, which is harmless. Added tests covering the disambiguation.

src/widgets/ask/widget.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // (Persisted entries still reach the model via formatAskResponsesNote.)
+    if (mode === 'choice' || mode === 'free') {
+      const answer = (text ?? chosen[0] ?? '').trim()
+      if (answer) {
+        // Best-effort: the answer is already persisted, so a failed send (e.g.
+        // no model selected) loses nothing — surface it without breaking the UI.
+        await chatInstance.sendMessage({ text: answer }).catch((error) => {
+          console.error('Ask widget failed to dispatch answer turn', error)
+        })
+      }
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good catch — fixed in c1f888cd. Added a synchronous `committedRef` re-entry guard in `display.tsx`: both `commit` (choice/graded) and `commitFree` now short-circuit when the ref is already set, so a rapid double-click can't fire `onSubmit` (and the dispatched user turn) twice before `submitted` state updates. Covered by two new tests in `display.test.tsx` (choice + free rapid double-click → onSubmit called once).

src/widgets/ask/display.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const commit = (ids: Set<string>) => {
+    if (committedRef.current) {
+      return
+    }
+    committedRef.current = true
+    setSubmitted(true)
+    onSubmit?.({
+      selectedIds: [...ids],
+      matched: evaluateAnswer({ prompt, mode, options }, ids),
+    })
```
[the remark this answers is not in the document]
darkbanjo: ⚡ This is intentional, and the project explicitly favors optimistic over defensive code here. The flow is: persist first (`updateMessageCache` is awaited before any dispatch), then best-effort `sendMessage`. The persisted entry always reaches the model via `formatAskResponsesNote`, so a swallowed live-send error loses nothing functionally — the answer is durable.

Re-enabling the UI on a transient failure would reintroduce exactly the duplicate-turn bug the prior thread flagged (the form would unlock and a second click could re-dispatch). Optimistic locking is the correct trade-off: prevent double-send, keep the answer persisted, and surface send failures via `console.error` rather than blocking the UI. The 'Got it' affordance reflects that the answer was recorded, which is true even if the live turn didn't dispatch. Declining the retry/rollback to avoid defensive complexity that contradicts the documented design — happy to revisit if you feel strongly.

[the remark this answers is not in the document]
darkbanjo: Intentional — leaving as-is. The optimistic lock is deliberate: the answer is persisted to the message cache *before* the turn is dispatched, and it also reaches the model via `formatAskResponsesNote`, so a failed `sendMessage` loses nothing. Re-enabling the UI on failure would reintroduce the exact double-submit/double-dispatch bug the "Duplicate autoturn on rapid submit" thread flagged (fixed in c1f888cd with a synchronous re-entry guard). A genuinely failed send surfaces through the chat's existing connection-error UI, so the prompt won't silently look finished while the chat is broken.

src/widgets/ask/widget.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+// dispatch decision per mode. `useQuery` is stubbed to resolve synchronously so
+// the widget renders its options immediately (no provider / async wait needed).
+const sendMessage = mock(async (_message: { text: string }) => {})
+mock.module('@/chats/chat-store', () => ({ useCurrentChatSession: () => ({ chatInstance: { sendMessage } }) }))
+mock.module('@/contexts', () => ({ useDatabase: () => ({}) }))
+mock.module('@/dal/chat-messages', () => ({ getMessage: async () => null, updateMessageCache: async () => {} }))
+mock.module('@tanstack/react-query', () => ({
+  useQuery: () => ({ data: null, isPending: false }),
+  useQueryClient: () => ({ setQueryData: () => {} }),
+}))
```
ital0: can you check if would be possible to get rid of these mock.module?

src/widgets/ask/display.tsx
```diff
@@ -0,0 +1,265 @@
...
+  options,
+  explanation,
+  initialSelectedIds,
+  initialText,
+  initialSubmitted,
+  onSubmit,
+}: AskProps) => {
+  const [selected, setSelected] = useState<Set<string>>(() => new Set(initialSelectedIds))
+  const [text, setText] = useState(initialText ?? '')
+  const [submitted, setSubmitted] = useState(initialSubmitted ?? false)
```
ital0: maybe you should replace these states + the `useRef` for a `useReducer`, as per CLAUDE.md rules.

src/widgets/ask/widget.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // act on / reply to. Graded `single`/`multiple` reveal the answer
+    // client-side and are NOT auto-sent — doing so would goad
+    // single-prompt-at-a-time backends into endlessly asking the next question.
+    // (Persisted entries still reach the model via formatAskResponsesNote.)
+    if (mode === 'choice' || mode === 'free') {
+      const answer = (text ?? chosen[0] ?? '').trim()
+      if (answer) {
+        // Best-effort: the answer is already persisted, so a failed send (e.g.
+        // no model selected) loses nothing — surface it without breaking the UI.
+        await chatInstance.sendMessage({ text: answer }).catch((error) => {
```
ital0: Style nit, and your best effort comment makes the intent clear so this is fine to keep: mixing `await` with a trailing `.catch()` is slightly awkward against our async/await preference. A `try`/`catch` around the `await` would match the house style. No strong feelings either way.

src/widgets/ask/lib.ts
```diff
@@ -0,0 +1,137 @@
...
+    return `- "${entry.prompt}" — chose ${chosen}`
+  })
+
+  return [
+    '## User responses',
+    'The user responded to interactive prompts in this conversation. Use these if they refer back to their responses — do not ask them to re-enter their choices.',
+    ...lines,
+  ]
+    .join('\n')
+    .trim()
```
ital0: this trailing `.trim()` is a no op, since the joined string starts with `## User responses` and has no surrounding whitespace. Safe to drop.

ital0 (changes requested):

src/widgets/ask/widget.test.tsx
darkbanjo: Done in a61c5dd2 — removed the `mock.module`s. Extracted the autoturn decision into a pure `turnTextForAnswer(mode, chosen, text)` in `lib.ts` and unit-test it directly (no providers/mocks). Deleted the mock-heavy integration test; the decision is now covered purely, and the rendered submit behavior stays covered by `display.test.tsx`.

src/widgets/ask/display.tsx
darkbanjo: Done in a61c5dd2 — consolidated `selected`/`text`/`submitted` into a `useReducer`. Kept `committedRef`: it's a synchronous side-effect latch guarding the one-time `onSubmit` against a pre-render double-click (state lags a render, so it can't dedupe that) — it isn't UI state, so it lives outside the reducer with a comment explaining why.

src/widgets/ask/widget.tsx
darkbanjo: Done in a61c5dd2 — switched to `try`/`catch` around the `await`, matching the house async/await style.

src/widgets/ask/lib.ts
darkbanjo: Done in a61c5dd2 — dropped the no-op `.trim()`.

src/widgets/ask/widget.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * entries are later surfaced to the model (see `formatAskResponsesNote`).
+ */
+export const AskWidget = ({ prompt, mode, options = [], explanation, messageId }: AskWidgetProps) => {
+  const db = useDatabase()
+  const queryClient = useQueryClient()
+  const { chatInstance } = useCurrentChatSession()
+  const storageKey = askStorageKey({ prompt, mode, options })
+  const queryKey = ['askState', messageId, storageKey]
+
+  const { data: saved, isPending } = useQuery({
```
raivieiraadriano92, line 31: Just confirming - could this useQuery be replaced with `src/hooks/use-message-cache.ts`

raivieiraadriano92: Maybe a small refactor in that hook to re-use it

darkbanjo: Good question — it's not a clean drop-in. `useMessageCache` is built for *lazy enrichment*: on a cache miss it calls `fetchFn` and **persists** the result (how link-preview / weather fetch-and-cache). The ask read is the opposite shape — a pure read-or-null: on a miss there's nothing to fetch (the answer only exists once the user submits, written by `handleSubmit`), and we must NOT write on miss. Going through `useMessageCache` with `fetchFn: () => null` would persist a `null` entry to the cache on every *unanswered* render. Reusing it cleanly would need a read-only / `persistOnMiss: false` option on that shared hook (which link-preview, weather, and connect-integration all depend on) — happy to do that as a small follow-up to dedup the read, but didn't want to change a shared hook right at final review.

PR #1006, opened 2026-06-18
feat(THU-615): require successful connection test before saving a model

raivieiraadriano92:
Extracts the model-connection-test logic from the Add Model dialog into a reusable `useModelConnectionTest` hook and wires it into the Edit Model flow, so users editing an existing model (e.g. updating its API key or URL) can verify the credentials before saving — same Test Model button, success/error cards, and 10s timeout as the add path. Any edit to the model id, URL, or API key invalidates a prior successful test.

Fixes THU-615.

---

> [!NOTE]
> **Medium Risk**
> Touches model credential configuration and live provider round-trips; behavior changes for Edit save gating and timeout/abort handling, though scope is limited to settings UI.
> 
> **Overview**
> Introduces **`useModelConnectionTest`**, moving the Add Model “Test Model” flow out of `models/index.tsx` into a dedicated hook that runs a `generateText` probe (10s `AbortSignal.timeout`, injectable probe for tests), tracks idle/success/error, invalidates displayed success when model/URL/API key no longer match the last tested config (including `''` vs `null`), and ignores stale in-flight results after `reset`.
> 
> The **Models settings page** drops reducer-driven connection state and inline `createModel`/`generateText` logic; Add and Edit both use the hook plus a shared **`ConnectionTestSection`** (test button, API-key hint for Anthropic/Tinfoil, status cards). **Edit Model** gains the same test-before-save UX; Save/Add stay disabled while a test is in **error** (not gated on success). Add dialog **resets** connection test on open/close; the Enter-key shortcut that triggered test/submit was removed.
> 
> Adds **hook unit tests** and a shared **`mockProxyFetch`** in test utils (used by powersync-reactivity and test-provider).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4a821c18229df0b7005d4f5b17114f252ea7188. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/hooks/use-model-connection-test.ts
+++ b/src/hooks/use-model-connection-test.ts
@@ -1,0 +1,186 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { createModel } from '@/ai/fetch'
+import type { FetchFn } from '@/lib/proxy-fetch'
+import { useProxyFetchGetter } from '@/lib/proxy-fetch-context'
+import type { Model } from '@/types'
+import { generateText } from 'ai'
+import { useCallback, useReducer, useRef } from 'react'
+
+const connectionTestTimeoutMs = 10_000
+
+export type ModelConnectionConfig = {
+  provider: Model['provider']
+  model: string
+  url?: string | null
+  apiKey?: string | null
+}
+
+type NormalizedConfig = {
+  provider: Model['provider']
+  model: string
+  url: string | null
+  apiKey: string | null
+}
+
+type ConnectionTestState = {
+  isTesting: boolean
+  rawStatus: 'idle' | 'success' | 'error'
+  error: string | null
+  tested: NormalizedConfig | null
+}
+
+type ConnectionTestAction =
+  | { type: 'START' }
+  | { type: 'SUCCESS'; tested: NormalizedConfig }
+  | { type: 'FAILURE'; tested: NormalizedConfig; error: string }
+  | { type: 'RESET' }
+
+const initialState: ConnectionTestState = {
+  isTesting: false,
+  rawStatus: 'idle',
+  error: null,
+  tested: null,
+}
+
+const reducer = (_state: ConnectionTestState, action: ConnectionTestAction): ConnectionTestState => {
+  switch (action.type) {
+    case 'START':
+      return { isTesting: true, rawStatus: 'idle', error: null, tested: null }
+    case 'SUCCESS':
+      return { isTesting: false, rawStatus: 'success', error: null, tested: action.tested }
+    case 'FAILURE':
+      return { isTesting: false, rawStatus: 'error', error: action.error, tested: action.tested }
+    case 'RESET':
+      return initialState
+  }
+}
+
+const normalize = (config: ModelConnectionConfig): NormalizedConfig => ({
+  provider: config.provider,
+  model: config.model,
+  url: config.url || null,
+  apiKey: config.apiKey || null,
+})
+
+/**
+ * Probe that issues a real `generateText` round-trip against the configured
+ * provider. Split out so tests can swap in a synchronous stub without touching
+ * global module mocks.
+ */
+export type ConnectionTestProbe = (
+  config: NormalizedConfig,
+  getProxyFetch: () => FetchFn,
+  signal: AbortSignal,
+) => Promise<void>
+
+const defaultProbe: ConnectionTestProbe = async (config, getProxyFetch, signal) => {
+  const aiModel = await createModel(
+    {
+      id: 'test',
+      name: 'Test Model',
+      provider: config.provider,
+      model: config.model,
+      url: config.url,
+      apiKey: config.apiKey,
+      isSystem: 0,
+      enabled: 1,
+      toolUsage: 1,
+      isConfidential: 0,
+      startWithReasoning: 0,
+      supportsParallelToolCalls: 1,
+      contextWindow: null,
+      deletedAt: null,
+      defaultHash: null,
+      vendor: null,
+      description: null,
+      userId: null,
+    },
+    getProxyFetch,
+  )
+  await generateText({
+    model: aiModel,
+    prompt: 'Say "test successful" if you can read this.',
+    maxRetries: 0,
+    abortSignal: signal,
+  })
+}
+
+/**
+ * Manages the state machine for a "Test Model" round-trip: idle → testing →
+ * success/error. Returns a `status` derived at render time from the tested
+ * config vs. the current one: any credential divergence collapses status to
+ * `'idle'` in the same render, so a stale `'success'` can't survive a
+ * credential edit (no useEffect-based invalidation).
+ *
+ * The probe is bounded by a single 10s `AbortSignal.timeout` piped through
+ * both the provider construction (via `Promise.race`) and the `generateText`
+ * request (via `abortSignal`), so a hanging Tinfoil attestation or a stuck
+ * `generateText` request both surface as a timeout error.
+ */
+export const useModelConnectionTest = (current: ModelConnectionConfig, probe: ConnectionTestProbe = defaultProbe) => {
+  const getProxyFetch = useProxyFetchGetter()
+  const [state, dispatch] = useReducer(reducer, initialState)
+  const runIdRef = useRef(0)
+
+  const test = useCallback(
+    async (config: ModelConnectionConfig) => {
+      const runId = ++runIdRef.current
+      const isCurrent = () => runIdRef.current === runId
+
+      dispatch({ type: 'START' })
+
+      const tested = normalize(config)
+      const timeoutSignal = AbortSignal.timeout(connectionTestTimeoutMs)
+      const timeoutPromise = new Promise<never>((_, reject) => {
+        timeoutSignal.addEventListener('abort', () =>
+          reject(new Error(`Connection test timed out after ${connectionTestTimeoutMs / 1000} seconds`)),
+        )
+      })
+      // Swallow the losing side's rejection so `Promise.race` doesn't leave an
+      // unhandled rejection when the timeout wins (e.g. a Tinfoil attestation
+      // that resolves after we've already bailed).
+      const runPromise = probe(tested, getProxyFetch, timeoutSignal)
+      runPromise.catch(() => {})
+
+      try {
+        await Promise.race([runPromise, timeoutPromise])
+        if (!isCurrent()) {
+          return
+        }
+        dispatch({ type: 'SUCCESS', tested })
+      } catch (err) {
+        console.error('Connection test error:', err)
+        if (!isCurrent()) {
+          return
+        }
+        dispatch({
+          type: 'FAILURE',
+          tested,
+          error: err instanceof Error ? err.message : 'Failed to connect to model',
+        })
+      }
+    },
+    [getProxyFetch, probe],
+  )
+
+  const reset = useCallback(() => {
+    runIdRef.current += 1
+    dispatch({ type: 'RESET' })
+  }, [])
+
+  const normalizedCurrent = normalize(current)
+  const matchesCurrent =
+    state.tested !== null &&
+    state.tested.provider === normalizedCurrent.provider &&
+    state.tested.model === normalizedCurrent.model &&
+    state.tested.url === normalizedCurrent.url &&
+    state.tested.apiKey === normalizedCurrent.apiKey
+
+  const status: 'idle' | 'success' | 'error' = matchesCurrent ? state.rawStatus : 'idle'
+  const error = matchesCurrent ? state.error : null
+
+  return { isTesting: state.isTesting, status, error, test, reset }
+}
--- a/src/hooks/use-model-connection-test.ts
+++ b/src/hooks/use-model-connection-test.ts
@@ -33,7 +33,7 @@ type ConnectionTestState = {
 }
 
 type ConnectionTestAction =
-  | { type: 'START' }
+  | { type: 'START'; tested: NormalizedConfig }
   | { type: 'SUCCESS'; tested: NormalizedConfig }
   | { type: 'FAILURE'; tested: NormalizedConfig; error: string }
   | { type: 'RESET' }
@@ -48,7 +48,7 @@ const initialState: ConnectionTestState = {
 const reducer = (_state: ConnectionTestState, action: ConnectionTestAction): ConnectionTestState => {
   switch (action.type) {
     case 'START':
-      return { isTesting: true, rawStatus: 'idle', error: null, tested: null }
+      return { isTesting: true, rawStatus: 'idle', error: null, tested: action.tested }
     case 'SUCCESS':
       return { isTesting: false, rawStatus: 'success', error: null, tested: action.tested }
     case 'FAILURE':
@@ -110,10 +110,11 @@ const defaultProbe: ConnectionTestProbe = async (config, getProxyFetch, signal) 
 
 /**
  * Manages the state machine for a "Test Model" round-trip: idle → testing →
- * success/error. Returns a `status` derived at render time from the tested
- * config vs. the current one: any credential divergence collapses status to
- * `'idle'` in the same render, so a stale `'success'` can't survive a
- * credential edit (no useEffect-based invalidation).
+ * success/error. Returns `isTesting`, `status`, and `error` derived at render
+ * time from the tested config vs. the current one: any credential divergence
+ * collapses all three to their idle values in the same render, so a stale
+ * `'success'` can't survive a credential edit and a mid-flight edit stops the
+ * spinner immediately (no useEffect-based invalidation).
  *
  * The probe is bounded by a single 10s `AbortSignal.timeout` piped through
  * both the provider construction (via `Promise.race`) and the `generateText`
@@ -130,20 +131,16 @@ export const useModelConnectionTest = (current: ModelConnectionConfig, probe: Co
       const runId = ++runIdRef.current
       const isCurrent = () => runIdRef.current === runId
 
-      dispatch({ type: 'START' })
-
       const tested = normalize(config)
+      dispatch({ type: 'START', tested })
+
       const timeoutSignal = AbortSignal.timeout(connectionTestTimeoutMs)
       const timeoutPromise = new Promise<never>((_, reject) => {
         timeoutSignal.addEventListener('abort', () =>
           reject(new Error(`Connection test timed out after ${connectionTestTimeoutMs / 1000} seconds`)),
         )
       })
-      // Swallow the losing side's rejection so `Promise.race` doesn't leave an
-      // unhandled rejection when the timeout wins (e.g. a Tinfoil attestation
-      // that resolves after we've already bailed).
       const runPromise = probe(tested, getProxyFetch, timeoutSignal)
-      runPromise.catch(() => {})
 
       try {
         await Promise.race([runPromise, timeoutPromise])
@@ -181,6 +178,7 @@ export const useModelConnectionTest = (current: ModelConnectionConfig, probe: Co
 
   const status: 'idle' | 'success' | 'error' = matchesCurrent ? state.rawStatus : 'idle'
   const error = matchesCurrent ? state.error : null
+  const isTesting = matchesCurrent && state.isTesting
 
-  return { isTesting: state.isTesting, status, error, test, reset }
+  return { isTesting, status, error, test, reset }
 }
--- a/src/settings/models/index.test.tsx
+++ b/src/settings/models/index.test.tsx
@@ -5,16 +5,13 @@
 import { createModel } from '@/dal'
 import { resetTestDatabase, setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
 import { getDb } from '@/db/database'
-import type { FetchFn } from '@/lib/proxy-fetch'
-import { ProxyFetchProvider } from '@/lib/proxy-fetch-context'
 import { renderWithReactivity, waitForElement } from '@/test-utils/powersync-reactivity-test'
 import { getClock } from '@/testing-library'
-import type { Model } from '@/types'
 import '@testing-library/jest-dom'
-import { act, cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { act, cleanup, screen } from '@testing-library/react'
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from 'bun:test'
 import { v7 as uuidv7 } from 'uuid'
-import ModelsPage, { EditModelForm } from './index'
+import ModelsPage from './index'
 
 describe('ModelsPage reactivity', () => {
   beforeAll(async () => {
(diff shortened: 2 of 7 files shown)


src/settings/models/index.tsx
```diff
@@ -161,6 +142,93 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+      }
+    },
+    [getProxyFetch],
+  )
+
+  const reset = useCallback(() => {
+    setStatus('idle')
+    setError(null)
+    setIsTesting(false)
+  }, [])
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 3028a9a1.

src/settings/models/index.tsx
```diff
@@ -300,11 +408,50 @@ const EditModelForm = ({
...
+            className="bg-red-50/50 dark:bg-red-500/10 border-red-200/50 dark:border-red-500/20"
+          />
+        )}
+
         <div className="flex justify-end gap-3 pt-2">
           <Button type="button" variant="ghost" onClick={onCancel}>
             Cancel
           </Button>
-          <Button type="submit" disabled={isPending || !form.formState.isDirty}>
+          <Button type="submit" disabled={isPending || !form.formState.isDirty || connectionStatus !== 'success'}>
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 1196c5aa.

src/settings/models/index.tsx
```diff
@@ -300,11 +454,56 @@ const EditModelForm = ({
...
+                <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+                Testing Model...
+              </>
+            ) : (
+              'Test Model'
+            )}
+          </Button>
+        )}
+
+        {!canTest && credentialsDirty && (
```
ital0: I think this hint can mislead: `canTest` is also false when the Model field is empty, so clearing the Model shows "Enter an API key" even when the key is fine. The Add dialog's condition is narrower, mirroring it here would fix this.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+          supportsParallelToolCalls: 1,
+          contextWindow: null,
+          tokenizer: null,
+          deletedAt: null,
+          defaultHash: null,
+          vendor: null,
+          description: null,
+          userId: null,
+        }
+        const aiModel = await createModel(modelConfig, getProxyFetch)
```
ital0: Heads up: `createModel` runs outside the `Promise.race`, so when it throws, the 10s timer later rejects with no handler attached (unhandled rejection), and a hanging tinfoil attestation leaves `isTesting` stuck with no timeout at all. A single `AbortSignal.timeout(10_000)` through both calls would cover all of it.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+  url?: string | null
+  apiKey?: string | null
+}
+
+const useModelConnectionTest = (current: CurrentModelConfig) => {
+  const getProxyFetch = useProxyFetchGetter()
+  const [isTesting, setIsTesting] = useState(false)
+  const [rawStatus, setRawStatus] = useState<'idle' | 'success' | 'error'>('idle')
+  const [error, setError] = useState<string | null>(null)
+  const [tested, setTested] = useState<TestedConfig | null>(null)
```
ital0: These four states always move together, and the old code modeled them as reducer events. House style is a reducer at three or more, would you fold them back into one?

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+}
+
+type CurrentModelConfig = {
+  provider: Model['provider']
+  model: string
+  url?: string | null
+  apiKey?: string | null
+}
+
+const useModelConnectionTest = (current: CurrentModelConfig) => {
```
ital0: Since the description calls this reusable, what do you think about moving it to `src/hooks/use-model-connection-test.ts`? It would also remove the test only export of `EditModelForm` and let the hook get its own tests.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+          url: config.url || null,
+          apiKey: config.apiKey || null,
+          isSystem: 0,
+          enabled: 1,
+          toolUsage: 1,
+          isConfidential: 0,
+          startWithReasoning: 0,
+          supportsParallelToolCalls: 1,
+          contextWindow: null,
+          tokenizer: null,
```
ital0: `tokenizer` is not a field on `Model` or a column on `modelsTable`, so this is a ghost property. Removing it here and in the test's `makeModel` also lets the fixture drop its `as Model` cast.

src/settings/models/index.test.tsx
```diff
@@ -67,3 +70,72 @@ describe('ModelsPage reactivity', () => {
     expect(screen.getByText('Second Model')).toBeInTheDocument()
   })
 })
+
+const mockProxyFetch = (async () => new Response()) as unknown as FetchFn
```
ital0: maybe export one shared stub with a real `preconnect` so no cast is needed?

src/settings/models/index.tsx
```diff
@@ -300,11 +454,56 @@ const EditModelForm = ({
           />
         )}
 
+        {canTest && (
```
ital0: This button, the two status cards, and the hint are nearly identical to the Add dialog's, and the hint conditions have already diverged. A small shared section component would keep the siblings honest.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+  const [rawStatus, setRawStatus] = useState<'idle' | 'success' | 'error'>('idle')
+  const [error, setError] = useState<string | null>(null)
+  const [tested, setTested] = useState<TestedConfig | null>(null)
+  // Tracks the latest run so completions from superseded tests (started before
+  // a credential change or a newer test) can't flip status back to success/error.
+  const runIdRef = useRef(0)
+
+  const test = useCallback(
+    async (config: ConnectionTestConfig) => {
+      if (!config.provider || !config.model) {
```
ital0: I think this guard is dead: `provider` cannot be falsy at the type level and every call site already gates on a non empty model. I would remove it and let the types do the work.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+        }
+        const aiModel = await createModel(modelConfig, getProxyFetch)
+        const { text } = await Promise.race([
+          generateText({ model: aiModel, prompt: 'Say "test successful" if you can read this.', maxRetries: 0 }),
+          timeoutPromise,
+        ])
+        if (!isCurrent()) {
+          return
+        }
+        console.log('Model test response:', text)
```
ital0: Leftover debug log from the old code, it prints the model's reply on every successful test. I would drop it.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+        setRawStatus('success')
+        setTested(testedConfig)
+      } catch (err) {
+        if (!isCurrent()) {
```
ital0: Since this return runs before the `console.error`, errors from superseded runs are never logged. I would log first, then bail.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
   }
 }
 
+type ConnectionTestConfig = {
```
ital0: These three config types nearly coincide. `name` is dead weight (`createModel` never reads it), so hardcoding "Test Model" in the hook would let one type cover all three roles.

src/settings/models/index.tsx
```diff
@@ -1139,7 +1299,7 @@ export default function ModelsPage() {
                   <Button type="button" variant="ghost" onClick={() => handleDialogOpenChange(false)}>
                     Cancel
                   </Button>
-                  <Button type="submit" disabled={addModelMutation.isPending}>
+                  <Button type="submit" disabled={addModelMutation.isPending || connectionStatus !== 'success'}>
```
ital0: Confirming intent: this now blocks adding any model (including thunderbolt, or while offline) until a test passes, with no hint in the idle state explaining why. Intended?

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+      const runId = ++runIdRef.current
+      const isCurrent = () => runIdRef.current === runId
+
+      setIsTesting(true)
+      setRawStatus('idle')
+      setError(null)
+      setTested(null)
+
+      const timeoutPromise = new Promise<never>((_, reject) => {
+        setTimeout(() => reject(new Error('Connection test timed out after 10 seconds')), 10000)
```
ital0: I think the pattern is to use `10_000`.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+
+  // Derive the exposed status from the tested config vs. the current one: a
+  // test result is only meaningful for the credentials it ran against. Any
+  // divergence collapses status to 'idle' during the same render — no
+  // useEffect-based invalidation, no window where a stale 'success' leaks.
+  const matchesCurrent =
+    tested !== null &&
+    tested.provider === current.provider &&
+    tested.model === current.model &&
+    tested.url === (current.url || null) &&
```
ital0: Both call sites already normalize with `|| null`, so this runs twice. One layer is enough, probably here in the hook.

src/settings/models/index.tsx
```diff
@@ -233,10 +355,42 @@ const EditModelForm = ({
...
+      model: values.model,
+      name: values.name,
+      url: values.url,
+      apiKey: values.apiKey,
+    })
+  }
+
+  const canTest = canTestModelConnection(model.provider, watchedModel, watchedApiKey)
+  const credentialsDirty =
+    watchedModel !== (model.model || '') || watchedUrl !== (model.url || '') || watchedApiKey !== (model.apiKey || '')
```
ital0: `model.model` is required on the type, so its `|| ''` is a no op. The `url` and `apiKey` ones are legit.

src/settings/models/index.tsx
```diff
@@ -1108,6 +1259,15 @@ export default function ModelsPage() {
                   </Button>
                 )}
 
+                {!canTestConnection &&
+                  !!watchedModel &&
+                  ['anthropic', 'tinfoil'].includes(watchedProvider) &&
+                  !watchedApiKey && (
```
ital0: At this point `!watchedApiKey` is always true (it is the only way `canTestConnection` can be false here), so it can go.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+  // divergence collapses status to 'idle' during the same render — no
+  // useEffect-based invalidation, no window where a stale 'success' leaks.
+  const matchesCurrent =
+    tested !== null &&
+    tested.provider === current.provider &&
+    tested.model === current.model &&
+    tested.url === (current.url || null) &&
+    tested.apiKey === (current.apiKey || null)
+
+  const status: 'idle' | 'success' | 'error' = matchesCurrent ? rawStatus : 'idle'
```
ital0: No strong opinion: `error` stays stale when status collapses to idle. Returning `matchesCurrent ? error : null` would keep them consistent.

ital0 (changes requested):

src/settings/models/index.tsx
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in aa42a69a.

src/settings/models/index.tsx
```diff
@@ -233,10 +316,38 @@ const EditModelForm = ({
...
+      name: values.name,
+      url: values.url,
+      apiKey: values.apiKey,
+    })
+  }
+
+  const canTest = canTestModelConnection(model.provider, watchedModel, watchedApiKey)
+  const credentialsDirty =
+    watchedModel !== (model.model || '') || watchedUrl !== (model.url || '') || watchedApiKey !== (model.apiKey || '')
+  const requiresPassingTest = credentialsDirty && connectionStatus !== 'success'
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 92430036.

src/settings/models/index.tsx
```diff
@@ -161,6 +142,147 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+  // Derive the exposed status from the tested config vs. the current one: a
+  // test result is only meaningful for the credentials it ran against. Any
+  // divergence collapses status to 'idle' during the same render — no
+  // useEffect-based invalidation, no window where a stale 'success' leaks.
+  const matchesCurrent =
+    tested !== null &&
+    tested.provider === current.provider &&
+    tested.model === current.model &&
+    tested.url === (current.url ?? null) &&
+    tested.apiKey === (current.apiKey ?? null)
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in e59ada44.

src/settings/models/index.tsx
```diff
@@ -108,13 +99,6 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
     case 'CLOSE_DIALOG':
       return { ...initialState, isAddDialogOpen: false }
 
-    case 'START_CONNECTION_TEST':
-      return { ...state, isTestingConnection: true, connectionStatus: 'idle', connectionError: null }
-    case 'CONNECTION_TEST_SUCCESS':
-      return { ...state, isTestingConnection: false, connectionStatus: 'success' }
-    case 'CONNECTION_TEST_FAILURE':
-      return { ...state, isTestingConnection: false, connectionStatus: 'error', connectionError: action.error }
-
```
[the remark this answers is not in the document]
raivieiraadriano92: Obviated by e4a821c1 (Option 2c): stale success no longer gates Save, only a live failure does.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.test.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Not intended — softened in e4a821c1 (Option 2c): gate only on `connectionStatus === 'error'`. Idle/success both allow submit, so Thunderbolt/offline no longer stuck.

src/settings/models/index.tsx
```diff
@@ -300,11 +392,21 @@ const EditModelForm = ({
...
+          status={connectionStatus}
+          error={connectionError}
+        />
+
         <div className="flex justify-end gap-3 pt-2">
           <Button type="button" variant="ghost" onClick={onCancel}>
             Cancel
           </Button>
-          <Button type="submit" disabled={isPending || !form.formState.isDirty}>
+          <Button type="submit" disabled={isPending || !form.formState.isDirty || connectionStatus === 'error'}>
```
ital0: The PR title says a successful test is required before saving, but Save is only blocked while the test is in error, so saving without ever running a test is still possible (same for the Add button). Is that the intended reading of the ticket? If so, maybe the title could be updated to match what ships.

src/settings/models/index.tsx
```diff
@@ -898,7 +920,7 @@ export default function ModelsPage() {
               <ResponsiveModalDescription className="sr-only">Add a new AI model</ResponsiveModalDescription>
             </ResponsiveModalHeader>
             <Form {...form}>
-              <form onSubmit={form.handleSubmit(onSubmit)} onKeyDown={handleKeyDown} className="grid gap-4 pt-4 pb-2">
+              <form onSubmit={form.handleSubmit(onSubmit)} className="grid gap-4 pt-4 pb-2">
```
ital0, line 951: Now that the keydown handler is gone, pressing Enter in an input falls through to the native form submit, so it submits right away instead of running the test first like before. Just checking this was considered when the shortcut was removed.

src/hooks/use-model-connection-test.ts
```diff
@@ -0,0 +1,186 @@
...
+      const timeoutPromise = new Promise<never>((_, reject) => {
+        timeoutSignal.addEventListener('abort', () =>
+          reject(new Error(`Connection test timed out after ${connectionTestTimeoutMs / 1000} seconds`)),
+        )
+      })
+      // Swallow the losing side's rejection so `Promise.race` doesn't leave an
+      // unhandled rejection when the timeout wins (e.g. a Tinfoil attestation
+      // that resolves after we've already bailed).
+      const runPromise = probe(tested, getProxyFetch, timeoutSignal)
+      runPromise.catch(() => {})
```
ital0: I believe this catch is redundant: Promise.race already attaches handlers to both promises, so a late rejection from the losing side counts as handled and never reaches unhandledrejection. Happy to be proven wrong, but the comment above may be guarding against something that cannot happen.

src/settings/models/index.tsx
```diff
@@ -161,6 +141,92 @@ const modelReducer = (state: ModelState, action: ModelAction): ModelState => {
...
+  provider,
+  model,
+  apiKey,
+  isTesting,
+  onTest,
+  status,
+  error,
+}: ConnectionTestSectionProps) => {
+  const canTest = canTestModelConnection(provider, model, apiKey)
+  const showApiKeyHint = !canTest && !!model && ['anthropic', 'tinfoil'].includes(provider)
```
ital0: Tiny one: this provider list also appears in canTestModelConnection just above. A shared named const would keep the two in sync if the list ever grows.

src/hooks/use-model-connection-test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  const normalizedCurrent = normalize(current)
+  const matchesCurrent =
+    state.tested !== null &&
+    state.tested.provider === normalizedCurrent.provider &&
+    state.tested.model === normalizedCurrent.model &&
+    state.tested.url === normalizedCurrent.url &&
+    state.tested.apiKey === normalizedCurrent.apiKey
+
+  const status: 'idle' | 'success' | 'error' = matchesCurrent ? state.rawStatus : 'idle'
```
ital0, line 179: Small asymmetry I noticed: status collapses to idle when the credentials diverge, but isTesting does not, so editing a field mid flight keeps the spinner going and the result is then discarded silently. Probably fine as is, just flagging it in case you want the button to snap back too.

ital0 (changes requested):

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.test.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in 5b10e412.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in e4a821c1.

src/settings/models/index.tsx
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 5b10e412.

raivieiraadriano92: Fixed in 5b10e412.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in 5b10e412.

src/hooks/use-model-connection-test.ts
raivieiraadriano92: Fixed in 0dd72278.

src/settings/models/index.tsx
raivieiraadriano92: Fixed in 0dd72278.

src/hooks/use-model-connection-test.ts
raivieiraadriano92: Fixed in 0dd72278.

src/settings/models/index.tsx
```diff
@@ -300,11 +409,28 @@ const EditModelForm = ({
...
           <Button type="button" variant="ghost" onClick={onCancel}>
             Cancel
           </Button>
-          <Button type="submit" disabled={isPending || !form.formState.isDirty}>
+          <Button
+            type="submit"
+            disabled={
+              isPending ||
+              !form.formState.isDirty ||
+              (providerRequiresConnectionTest(model.provider) && connectionStatus !== 'success')
```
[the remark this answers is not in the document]
raivieiraadriano92: Yes, intended

PR #1010, opened 2026-06-19
fix: route deep-review findings via structured output so empty model fails loudly

ital0:
## Why

The `thunder-deep-review` workflow's review check was passing **green but doing nothing**.

The model step runs read-only (`--allowedTools Read,Grep,Glob` — no `Write`/`Bash`), yet the prompt's OUTPUT CONTRACT told it to *write* the findings JSON to `THUNDER_FINDINGS_FILE`. With no `Write` tool, the file was never created → the post phase's `readModelFindings()` hit `ENOENT` → the orchestrator fail-soft `exit(0)` → a green check that never posted a review.

## Fix

- Emit findings through `claude-code-action`'s `--json-schema` **structured output** channel instead of a model-written file. The model returns its findings as its final result; the action constrains it to the schema and exposes it on `steps.model.outputs.structured_output`. No write tool needed — model stays read-only.
- A new **non-model shell step** (`Materialize structured findings`) bridges `structured_output` → `THUNDER_FINDINGS_FILE` for the orchestrator.
- Per a Codex second opinion: that step **fails loudly on empty `structured_output`** (model auth failure, `--max-turns` exhaustion, schema reject, etc. → **red check**, not a silent green no-op). A genuine empty review (`{"findings":[]}`) is non-empty JSON and passes the gate; the post step then no-ops cleanly.

No change to `review-orchestrator.mjs` — the structured output is `{"findings":[...]}`, which already matches what the post phase parses.

---

> [!NOTE]
> **Low Risk**
> CI workflow wiring only; no app runtime or auth logic changes, with clearer failure signaling when the model step breaks.
> 
> **Overview**
> Fixes **`thunder-deep-review`** silently passing when the model never produced findings. The read-only model step (`Read,Grep,Glob` only) was still instructed to **write** `THUNDER_FINDINGS_FILE`, so the file was missing, the post orchestrator fail-soft exited 0, and the check went green with no comment.
> 
> The model step now returns findings through **`claude-code-action`** **`--json-schema`** structured output (`steps.model.outputs.structured_output`) with an inline findings schema, and the prompt **OUTPUT CONTRACT** tells the model to return structured JSON only (no file writes). A new **`Materialize structured findings`** shell step copies that output into `THUNDER_FINDINGS_FILE` for the existing post phase.
> 
> That materialize step **fails the job** if `structured_output` is empty or whitespace (broken auth, max turns, schema reject), while a real no-op review (`{"findings":[]}`) still passes and lets post no-op cleanly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a24ae3f09f353f3f3ee2314c0f50ffbd0a02bf7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/thunder-deep-review.yml
+++ b/.github/workflows/thunder-deep-review.yml
@@ -1,0 +1,238 @@
+# =============================================================================
+# Advisory, comment-only AI review. Never approves, never requests changes,
+# never merges. Human approval + merge stay exactly as today.
+#
+# Notes:
+#   - The `anthropics/claude-code-action` input names below were verified
+#     against the live @v1 `action.yml` (2026-06): `prompt`, `anthropic_api_key`,
+#     `claude_args`, and `use_sticky_comment` all EXIST. `model:` /
+#     `allowed_tools:` / `review_event:` do NOT exist in @v1 — do NOT
+#     reintroduce them (pass --model/--allowedTools via `claude_args`).
+#     Re-confirm only if the action major version changes.
+#   - Action commit SHAs are PINNED below (checkout / claude-code-action).
+#   - The only secret the team sets is the ANTHROPIC_API_KEY repo secret.
+#   - Recommended: CODEOWNERS-protect this workflow + .github/scripts/review-orchestrator.mjs
+#     (they hold the review logic + the API-key reference).
+#
+# Orchestration is deterministic CODE, not model tool-calls: .github/scripts/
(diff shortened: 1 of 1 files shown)


ital0:
Superseded: this fix belongs on PR #1003 (branch deep-review-tooling) and was folded there. This branch was created in error by a tooling mishap. Closing.

PR #1015, opened 2026-06-23
perf(THU-595): unblock app init by scheduling reconcile in background

raivieiraadriano92:
Render no longer blocks on PowerSync's initial-sync gate. The old `useAppInitialization` flow awaited `waitForInitialSync` (up to 10s) + `reconcileDefaults` + `runDataMigrations` before unblocking — violating local-first. These now run as a background pass scheduled right after the DB opens.

Race-safety the old gate provided is preserved: when sync is enabled, the reconciler waits internally before touching `ps_crud`-tracked tables; on a `timed_out` / `failed` first sync it defers via a one-shot listener so reconcile runs the moment priority-1 sync recovers later in the session — no page reload. Telemetry split: `app_init_timing` now measures only the render-blocking phase; reconcile/migration durations + sync outcome move to a new `app_init_reconcile` event.

---

> [!NOTE]
> **Medium Risk**
> Changes first-launch vs returning-user sync/reconcile ordering; empty-DB path still waits, but mis-detecting “empty” could reintroduce clobber risk or skip needed sync.
> 
> **Overview**
> **Returning users with a populated local database no longer block on PowerSync’s initial-sync gate during app init.**
> 
> Init now probes `settingsTable` (via `isLocalDataEmpty`) before step 3: if at least one settings row exists, it skips `waitForInitialSync`, records `initial_sync_outcome: 'not_required'` in `app_init_timing`, and proceeds straight to `reconcileDefaults`. The wait still runs on an empty local DB, where reconcile would otherwise insert shipped defaults into `ps_crud` and risk overwriting cloud customizations before sync completes.
> 
> Telemetry type `InitialSyncOutcome` gains `'not_required'`. Tests expect no `step3_wait_for_initial_sync_ms` when the test DB is pre-seeded, and a new suite with `resetTestDatabase` asserts step 3 is timed when the local DB is empty.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c85c3e3dd7aca0a960ab1a35bdb7949aee96b442. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/db/database-interface.ts
+++ b/src/db/database-interface.ts
@@ -11,8 +11,11 @@ import type * as schema from './schema'
  * - `synced`: priority-1 buckets completed (instant for returning users via offline-status restore)
  * - `timed_out`: sync didn't complete within the timeout (e.g. network down)
  * - `failed`: the wait rejected unexpectedly; app boots regardless
+ * - `not_required`: local DB already has reconciled data, so the wait was skipped.
+ *   `reconcileDefaults` is a no-op for unchanged rows (line 76 in reconcile-defaults.ts),
+ *   so a populated local DB can reconcile immediately without racing the sync stream.
  */
-export type InitialSyncOutcome = 'disabled' | 'synced' | 'timed_out' | 'failed'
+export type InitialSyncOutcome = 'disabled' | 'synced' | 'timed_out' | 'failed' | 'not_required'
 
 export type DatabaseInterface = {
   db: BaseSQLiteDatabase<'async', any, typeof schema>
--- a/src/hooks/use-app-initialization.test.tsx
+++ b/src/hooks/use-app-initialization.test.tsx
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
+import { resetTestDatabase, setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
 import { getInitTimingPayload, resetInitTiming } from '@/lib/init-timing'
 import { createMockHttpClient } from '@/test-utils/http-client'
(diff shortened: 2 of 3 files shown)


src/lib/post-init-reconcile.ts
```diff
@@ -0,0 +1,169 @@
...
+      db,
+      outcome,
+      waitDurationMs,
+      triggeredBy: 'init',
+      totalStartedAt,
+    })
+    return
+  }
+
+  // timed_out or failed — defer reconcile to whenever the priority-1 buckets land.
```
ital0: I think this changes a guarantee for one specific case and I want to make sure it's intentional. In the old init path, a `timed_out` wait still ran `reconcileDefaults` afterward (the 10s timer resolved, then reconcile ran unconditionally). Here, `timed_out`/`failed` skips reconcile entirely and only re-runs it if priority-1 sync recovers later in the same session via `onceFirstSyncCompletes`.

For a returning user that's clearly the safer call — it avoids seeding "missing" defaults into `ps_crud`-tracked tables and clobbering their already-synced cloud customizations, exactly the hazard the docstring calls out. The case I'd want a second look at: a brand-new device with sync *enabled* that launches genuinely offline (network down, first sync times out and never recovers this session). It now boots with zero seeded defaults — no default models/modes/skills/settings — and crucially no `anonymous_id` (seeded inside `reconcileDefaults`) for the whole session, where the old code would have seeded all of it after the 10s timeout.

Is that acceptable as a session-scoped degradation, or should we distinguish "empty local DB + sync never connected" (safe to seed — there's no cloud state to overwrite yet) from "sync in-flight" (must wait)? The PR body and docstring justify the deferral well for the overwrite hazard but don't mention this offline-new-device tradeoff, so flagging it for a conscious decision rather than as a blocker.

ital0 (changes requested):

darkbanjo (dismissed):

PR #1026, opened 2026-06-25
feat(THU-624): file-capability columns on models table (backend / PR1)

darkbanjo:
## THU-624 — PR1 of 2 (backend-only, synced table)

First half of the two-PR synced-table flow for sourcing per-model file capabilities from the [models.dev](https://models.dev) registry. **Backend schema only** — no frontend, no behavior change yet.

### What's here
Adds three nullable columns to the synced `models` table (`backend/src/db/powersync-schema.ts`), mirroring the existing `context_window` / `tool_usage` precedent:
- `accepts_images` (integer 0/1) — vision support
- `native_file_types` (text, JSON `string[]`, e.g. `["application/pdf"]`) — mime types the model ingests natively
- `max_file_bytes` (integer, nullable) — hard per-file ceiling if known (models.dev doesn't model this → null falls back to the coarse client guard)

Plus the generated Drizzle migration `0020_sudden_steve_rogers.sql` (additive `ADD COLUMN`s, all nullable) and its `_journal.json` / snapshot entries.

### Why nullable / additive
Existing rows resolve with `null` (treated as text-only by the client), so this is a safe, backward-compatible migration with no backfill.

### Sync rules
The `models` sync rule in `powersync-service/config/config.yaml` is `SELECT *`, so new columns sync automatically — **no rule change in this PR**. ⚠️ Before PR2 (frontend) lands: run the migration, and confirm the **PowerSync Cloud dashboard** rule for `models` is also `SELECT *` (so the new columns replicate in production).

### Follow-up (PR2, frontend)
Frontend Drizzle schema (`src/db/tables.ts`) + PowerSync client schema (`src/db/powersync/schema.ts`), the **live models.dev fetch at model-add time** to populate these columns, the DAL/defaults, and wiring the derived caps into `createBuiltInAdapter` (replacing `capabilities: null`). See `thunderbolt-spec` notes Parts 2 & 4.

Verified: `bunx tsc --noEmit` (backend) + eslint clean; migration is additive/nullable; `_journal.json` updated (idx 20).

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 34d71a8909194ed88b00f23a0c25f117c645a605. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/drizzle/0020_sudden_steve_rogers.sql
+++ b/backend/drizzle/0020_sudden_steve_rogers.sql
@@ -1,0 +1,7 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+ALTER TABLE "powersync"."models" ADD COLUMN "accepts_images" integer;--> statement-breakpoint
+ALTER TABLE "powersync"."models" ADD COLUMN "native_file_types" text;--> statement-breakpoint
+ALTER TABLE "powersync"."models" ADD COLUMN "max_file_bytes" integer;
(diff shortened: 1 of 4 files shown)


darkbanjo:
Parking: we've moved to native-first delivery with no client-side capability detection (the model/harness handles what it can; we surface failures + offer convert-to-text). The models.dev-derived capability columns this PR added are no longer needed. Closing unmerged.

PR #1028, opened 2026-06-27
UI style: restyle theme with new fonts and color palette

liviolanda:
- add self-hosted fonts: Mona Sans (body), Mozilla Text (headings), Space Mono (code)
- introduce full light and dark color palette with a cohesive slate dark theme
- fix dark-mode border visibility across form controls (input, select, switch, toggle, etc.)
- make chat-input font size responsive and match its background/outline to slate
- darken the user message bubble slightly in light mode
- remove sidebar section dividers

---

> [!NOTE]
> **High Risk**
> Unresolved merge conflicts in `chat-prompt-input.tsx` and `message-bubbles.tsx` will break the build and may drop attachment UI if the wrong side is kept.
> 
> **Overview**
> **Visual refresh** replaces the oklch-based theme with a cohesive hex palette for light and dark (slate dark), remaps Tailwind `white`/`black` to off-white and slate, and wires **Mona Sans** (UI), **EB Garamond** (headings via base layer), and **Space Mono** (code) through `@fontsource` imports in `index.tsx`. `theme-provider` and provider SVG fills now use the same off-white/dark background colors instead of `#fff` / `#0a0a0a`.
> 
> **Form and shell polish:** shared controls (`input`, `select`, `textarea`, `combobox`, `toggle` outline, `switch` in dark) switch default borders from `border-input` to **`border-border`** for clearer edges; the chat **prompt** overlay/textarea use `var(--font-size-body)` instead of `text-base`. Sidebars drop **separators** and several **top/bottom borders** (header, footer).
> 
> **Chat-specific:** user text bubbles move from `bg-muted` to **`bg-accent`** with **`text-[14px]`** in light/dark.
> 
> **Blocker:** the diff still contains **unresolved `<<<<<<<` / `=======` / `>>>>>>>` markers** in `chat-prompt-input.tsx` (prompt container classes: `bg-card` + dark oklch vs `bg-sidebar`) and `message-bubbles.tsx` (HEAD attachment/`FileCard` flow vs a stripped text-only branch)—the PR cannot build until those are resolved.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd5658805564d495cb86c5353ee89a85bb3ae3c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/chat/message-bubbles.tsx
```diff
@@ -13,7 +13,7 @@ export const MessageBubbles = ({ message }: MessageBubblesProps) =>
   message.parts
     .filter((part) => part.type === 'text')
     .map((part, j) => (
-      <div key={j} className="px-4 rounded-2xl max-w-3/4 bg-muted dark:bg-secondary/60 ml-auto mt-6">
+      <div key={j} className="px-4 rounded-2xl max-w-3/4 bg-[#e8e8e8] dark:bg-secondary/60 ml-auto mt-6">
```
ital0: this hardcodes a raw hex (`#e8e8e8`) for the light-mode bubble. Looks like we have a `bg-accent` token that's `#e7e7e7`. Would it make sense to use it or create a new token for this value?

src/index.tsx
```diff
@@ -6,6 +6,13 @@ import ReactDOM from 'react-dom/client'
 import { App } from './app'
 import './polyfills'
 
+// Self-hosted fonts (offline-capable). Variable fonts ship all weights in one file.
+import '@fontsource-variable/mona-sans'
```
ital0, line 10: The bare `@fontsource-variable/*` import pulls every unicode subset (latin, latin-ext, cyrillic, greek, vietnamese, …) plus the full weight axis into the entry bundle, which loads on the landing path. If we only need latin, importing the scoped subset CSS (e.g. the `latin`/weight-axis file) would ship a lot fewer bytes. Was loading all subsets intended?

src/index.tsx
```diff
@@ -6,6 +6,13 @@ import ReactDOM from 'react-dom/client'
...
+import '@fontsource-variable/dm-sans'
```
ital0: DM Sans is listed only as a fallback after Mona Sans in `--font-sans`, but both are self-hosted. So Mona Sans always loads and DM Sans never actually renders, while still being downloaded. Would it makese to drop the DM Sans import and let `ui-sans-serif` / `system-ui` be the real fallback?

ital0 (changes requested):

src/components/chat/message-bubbles.tsx
liviolanda: create a new token!

src/index.tsx
liviolanda: no, don't know how to fix it. would you help me?

src/index.tsx
liviolanda: Yes!

cjroth:
Closing (adopted in #1095)

PR #1044, opened 2026-07-01
chore(THU-645): update default models

raivieiraadriano92:
Drop Kimi K2.6 and rename DeepSeek V4 Pro → V4 Flash. Row id is preserved so reconciliation upgrades unmodified rows in place; edited rows survive.

---

> [!NOTE]
> **Medium Risk**
> Changes which system models users see and how DeepSeek is served (Tinfoil confidential → Thunderbolt/Fireworks), with migration behavior hinging on soft-deleting the old Pro default rather than reusing its id.
> 
> **Overview**
> Updates the **provided** default model lineup: **Kimi K2.6** and its default profile are removed, and **DeepSeek V4 Pro** (confidential Tinfoil) is replaced by **DeepSeek V4 Flash** as a standard **Thunderbolt** model (`deepseek-v4-flash`, non-confidential).
> 
> **DeepSeek V4 Flash** uses a **new row id** (not V4 Pro’s) so encrypted threads tied to Pro’s confidential row are not stranded by an `isConfidential` flip; existing Pro rows are expected to be **soft-deleted** via default cleanup so those chats show as retired instead of breaking. A matching default model profile is wired to the new model id.
> 
> On the backend, **`deepseek-v4-flash`** is added to inference **`supportedModels`**, routed to **Fireworks** (`accounts/fireworks/models/deepseek-v4-flash`). Tests and fixtures are updated (including Tinfoil proxy examples using **`glm-5-2`** instead of retired DeepSeek Tinfoil ids).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b3019c9fc98545e748b118d3b9c550bd7825b78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/defaults/models.ts
```diff
@@ -58,14 +58,20 @@ export const defaultModelOpus48: Model = {
...
-  name: 'DeepSeek V4 Pro',
-  provider: 'tinfoil',
-  model: 'deepseek-v4-pro',
+  name: 'DeepSeek V4 Flash',
+  provider: 'thunderbolt',
+  model: 'deepseek-v4-flash',
   isSystem: 1,
   enabled: 1,
-  isConfidential: 1,
+  isConfidential: 0,
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 5b3019c9.

src/defaults/models.ts
```diff
@@ -58,14 +58,22 @@ export const defaultModelOpus48: Model = {
...
+/**
+ * Flash ships under a fresh id — not the retired V4 Pro id. Reusing Pro's id
+ * would flip isConfidential 1→0 on threads that were created encrypted
+ * (isEncrypted mirrors the model's isConfidential at creation), stranding them
+ * because the model picker and send guard both enforce isEncrypted === isConfidential.
+ * cleanupRemovedDefaults soft-deletes the Pro row instead, so those threads
+ * surface as "model retired" rather than broken.
+ */
+export const defaultModelDeepseekV4Flash: Model = {
+  id: '019f227e-d640-727d-ba12-d51bd7d0a3d6',
```
ital0, line 70: Love that Flash ships under a fresh id, the reasoning above checks out on every enforcement point I traced. One thing worries me though: any device still on an older build already runs `cleanupRemovedDefaults`, and its defaults array doesn't contain this id. When the Flash row syncs to that device it looks like an unedited system model that's no longer shipped, so the old client soft deletes it. And since `hashModel` includes `deletedAt`, updated clients then treat the row as user modified and never bring it back, so the whole account loses the new default for good. A lagging desktop app next to an updated web session is enough to trigger it. Maybe cleanup should only touch an explicit list of retired ids, or reconcile could resurrect a row whose hash matches the shipped default plus `deletedAt`?

src/defaults/models.ts
```diff
@@ -127,7 +114,6 @@ export const defaultModelGlm52: Model = {
  */
 export const defaultModels: ReadonlyArray<Model> = [
   defaultModelOpus48,
-  defaultModelDeepseekV4Pro,
-  defaultModelKimiK26,
+  defaultModelDeepseekV4Flash,
```
ital0, line 117: Flagging with no strong feelings: user edited Pro or Kimi rows survive by design, still pointing at Tinfoil ids that are presumably going away upstream. Those users will hit raw upstream errors with no hint the model was retired. If that's the accepted tradeoff, all good.

src/defaults/model-profiles.test.ts
```diff
@@ -77,8 +77,8 @@ describe('hashModelProfile', () => {
 })
 
 describe('defaultModelProfiles', () => {
-  test('contains four profiles', () => {
-    expect(defaultModelProfiles).toHaveLength(4)
+  test('contains three profiles', () => {
+    expect(defaultModelProfiles).toHaveLength(3)
```
ital0, line 81: Tiny thought, feel free to ignore: these count tests need a manual bump every time the lineup changes. An assertion that every default profile's `modelId` maps to a default model (and the other way around) would catch real wiring mistakes without the ceremony.

ital0 (changes requested):

raivieiraadriano92:
Superseded by #1049, which bundles this lineup change with the THU-637 version-gate and adds a resurrect guard for the pre-THU-637 cleanup race flagged in the review here. Closing.

PR #1049, opened 2026-07-02
fix(THU-637): version-gate models reconcile + OTA + THU-645 lineup

raivieiraadriano92:
## Fixes THU-637 (models flap) and lands THU-645 (new lineup) atomically

The two changes ship together because the version-gate introduced for THU-637 is the mechanism that makes the THU-645 lineup change safe against multi-device races.

### The race — THU-637
`models` is a PowerSync-synced table, but the shipped defaults live in the client binary. Every launch, `reconcileDefaults` treats rows whose stored `defaultHash` matches their current content as "unedited" and rewrites them with the local bundle's defaults. When two devices (or the same device across app updates) ship different defaults, they ping-pong via sync every boot: a V2 device writes V2 rows → a V1 device reconciles them back to V1 → V2 upgrades again → ...

Even single-device: `waitForInitialSync` timing out (10 s cap) can drop reconcile back to local V1 while newer rows arrive post-reconcile — next launch downgrades.

### The fix — per-table monotonic version gate
Ship `defaultModelsVersion` alongside `defaultModels`. Store the highest ever applied version in the already-synced `settingsTable` under `defaults_version.models`:

```
canOverwrite = pickedVersion > (storedVersion ?? -∞)
```

- `canOverwrite = false` (my bundle is older than what's already applied on this account) → skip in-place updates AND skip `cleanupRemovedDefaults`.
- `canOverwrite = true` → normal reconcile, bump stored version at the end **only when at least one row was actually written**.

Hash-based user-edit protection still runs alongside — user edits survive under both older-bundle and newer-bundle passes.

Only `models` (and paired `model_profiles`) are gated in this PR; other reconciled tables (modes, tasks, skills, settings) keep their current behavior via `canOverwrite: true` defaults. No new synced column — the version marker piggybacks on the existing `settings` sync, so no PowerSync two-PR flow.

Additional guards layered on top of the base gate:
- **Sync-incomplete pessimism** — when \`initialSyncCompleted=false\` and the DB has any rows, cloud state is unknown; skip mutations for that boot.
- **Insert guard on canOverwrite** — models can be legitimately retired under a newer version, so ghost-inserting them is unsafe; profiles opt into \`insertMissing: true\` so the 1:1 model↔profile invariant holds regardless of authority.
- **Marker only advances on writes** — no false-positive "V2 applied" signal to peers whose rows haven't caught up.
- **Cleanup-soft-delete resurrect** — undo a lagging pre-THU-637 client's cleanup soft-delete when the row's content still matches the shipped default's authoring hash. Discriminator is \`defaultHash\`: cleanup preserves it (row is resurrectable), user's \`deleteModel\` scrubs it via \`clearNullableColumns\` (deletion is permanent).

### `/config` as OTA channel for defaults
Backend `/config` now returns:

```ts
defaults: {
  models: { version: N, data: SharedModel[] }
}
```

Frontend picks between server and bundle by `max(version)` with a payload-sanity guard (`Number.isFinite(version) && Array.isArray(data) && data.length > 0`).

- Server ahead → OTA rollout, no client release needed.
- Server behind/equal → bundle wins by strict `>` (no wasted swap).
- Offline → cached `/config` via `zustand/persist`, or bundle on first-ever fresh install.
- Rollback is monotonic: retract a bad server payload by shipping a **higher** version with reverted content.

The picker awaits the `/config` fetch that was already fired at boot Step 0, bounded by its 5 s internal timeout. On cache-hit boots (persisted `/config` present) it uses the cached value immediately and lets the fetch resolve in the background.

### THU-645 — new model lineup
- **Kimi K2.6 retired** (removed from `defaultModels`).
- **DeepSeek V4 Pro** (Tinfoil confidential) → **DeepSeek V4 Flash** (Thunderbolt/Fireworks, non-confidential). Flash ships under a **fresh id** — reusing V4 Pro's id would flip `isConfidential` 1 → 0 on threads created encrypted, stranding them because the picker and send-guard enforce `isEncrypted === isConfidential`. Cleanup soft-deletes the Pro row instead, so encrypted threads bound to it surface as "model retired" rather than broken chats.
- `deepseek-v4-flash` added to backend `supportedModels`, routed to Fireworks.
- `defaultModelsVersion` bumped 1 → 2. Snapshot test pinned to `{ version: 2, hash: … }`.

### Race guard in action for THU-645 rollout
- Combined-PR device applies V2 → syncs Flash + `stored=2`.
- Pre-THU-637 device (still on prod main) syncs Flash — its unconditional `cleanupRemovedDefaults` doesn't recognize Flash's id, so it soft-deletes Flash. Cloud now has Flash deleted.
- Any combined-PR device seeing the deletion: the resurrect branch un-deletes Flash (content still matches Flash's authoring hash → distinguishable from a user delete). Cloud converges back to Flash alive. Flap self-limits as pre-THU-637 devices update.

Without the resurrect branch, Flash would be permanently soft-deleted on any account touched by a lagging pre-THU-637 device — `stored=2` plus `deletedAt`-included hashing would make every future reconcile treat the row as user-edited.

### Version-bump discipline
A colocated snapshot test in `shared/defaults/models.test.ts` fails on any change to `defaultModels` without a matching bump to `defaultModelsVersion`. AGENTS.md documents the convention.

### Also in this PR (ported from #1012)
- `fix(backend): keep pglite in-memory when DATABASE_URL is a connection string` — prevents pglite tests from bootstrapping a spurious data dir when `.env` has `DATABASE_URL=postgresql://…`.
- `fix(tauri): allow localhost:8080 in CSP connect-src` — needed for the local dev backend on 8080.

## Supersedes #1044
PR #1044 shipped the THU-645 lineup change against the pre-THU-637 branch. Its open comments — pre-THU-637 cleanup deleting Flash across the account, retired Tinfoil ids on user-edited rows, count-based profile tests — are addressed here (race by version-gate + resurrect guard; test brittleness by the new structural profile↔model check). PR #1044 can be closed as superseded.

## Test plan
- [x] Frontend + backend `bun tsc --noEmit`: clean
- [x] `bun run test:5x`: 15,470 frontend + 440 shared, 0 fail across 5 iterations
- [x] Backend full suite: 830/830 pass
- [x] Frontend + backend `bun run lint`: 0 errors
- [ ] Manual: wipe local Thunderbolt data → boot on this branch → confirm Flash present, `defaults_version.models = 2` written
- [ ] Manual: on an existing prod-main install (with Pro + Kimi rows), install this build → confirm Flash inserted, Pro + Kimi soft-deleted in place, existing user-edited rows preserved
- [ ] Manual: server serves a higher `defaults.models.version` with a different content payload → confirm client adopts (OTA)
- [ ] Manual: simulate the pre-THU-637 race — soft-delete Flash's row manually (leaving defaultHash intact), reboot → confirm resurrect fires

## Follow-up
Same version-gate + `/config` treatment for `modes` / `tasks` / `skills` / `settings` — pure repetition of the pattern established here. In-app "model retired" hint for user-edited retired rows (ital0's #2 on 1044) also for follow-up.

---

> [!NOTE]
> **High Risk**
> Touches PowerSync-synced models reconciliation, multi-device convergence, and default model IDs/confidentiality—incorrect gating or OTA picking could wipe or flap user-visible models across accounts.
> 
> **Overview**
> Stops multi-device **models** reconcile ping-pong (THU-637) by gating writes on a monotonic `defaults_version.models` in synced settings and only advancing that marker when models/profiles actually change. Reconcile now takes **authority flags** (`canOverwrite`, `canResurrect`, `insertMissing`), skips mutations when initial sync is incomplete but the DB already has rows, can **resurrect** cleanup-shaped soft-deletes, and keeps model↔profile pairing (including `insertMissing` profiles and dropping OTA-only models without bundled profiles).
> 
> **OTA:** Default models move to `shared/defaults/models.ts` with `defaultModelsVersion` (now **2**) and a snapshot test; `GET /config` exposes `defaults.models`, the config store persists it, boot uses `pickModelsDefaults` (sanity checks + bundle floor) before reconcile.
> 
> **Lineup (THU-645):** Kimi retired; DeepSeek V4 Pro → **V4 Flash** on Thunderbolt/Fireworks under a **new id** (avoids flipping `isConfidential` on encrypted threads); `deepseek-v4-flash` added to inference routing; profiles/tests updated.
> 
> Also: shared `hashValues`, PGlite ignores postgres `DATABASE_URL` for in-memory tests, Tauri CSP allows localhost:8080, AGENTS.md documents version-bump discipline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 107513d0ee56157f9959fbfc5d438fa64afd2ae8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/pick-defaults.ts
```diff
@@ -0,0 +1,29 @@
...
+ * bundle wins.
+ *
+ * Rollback semantics are monotonic — a server that regresses its declared
+ * version below what the client already has will not overwrite. To retract a
+ * bad server-published set, ship a *higher* version with the reverted content.
+ */
+export const pickModelsDefaults = (server: ServerModelsDefaults | undefined): ModelsDefaults => {
+  if (server && server.version > defaultModelsVersion) {
+    return { version: server.version, data: server.data }
+  }
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d450d47c.

src/lib/reconcile-defaults.ts
```diff
@@ -167,6 +229,18 @@ export const reconcileDefaults = async (db: AnyDrizzleDatabase) => {
     // Settings
     await reconcileDefaultsForTable(tx, settingsTable, defaultSettings, hashSetting, 'key')
 
+    if (canOverwriteModels) {
+      // Inline upsert: `updateSettings` wraps its writes in its own transaction
+      // and PowerSync's drizzle driver forbids nested transactions. We already
+      // know from `readAppliedVersion` whether the row exists, so branch here.
+      const versionValue = String(modelsSource.version)
+      if (storedModelsVersion === null) {
+        await tx.insert(settingsTable).values({ key: modelsVersionKey, value: versionValue })
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d450d47c.

src/lib/reconcile-defaults.ts
```diff
@@ -143,13 +191,27 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
     // Soft-delete removed system defaults before reconciling current ones.
-    await cleanupRemovedDefaults(tx)
+    await cleanupRemovedDefaults(tx, canOverwriteModels, modelsSource.data)
 
     // AI models
-    await reconcileDefaultsForTable(tx, modelsTable, defaultModels, hashModel)
+    await reconcileDefaultsForTable(tx, modelsTable, modelsSource.data, hashModel, 'id', canOverwriteModels)
 
     // Model profiles (after models, because they reference model IDs)
     await reconcileDefaultsForTable(tx, modelProfilesTable, defaultModelProfiles, hashModelProfile, 'modelId')
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d450d47c.

shared/defaults/models.test.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+const computeSnapshotHash = () =>
+  defaultModels.map((model, index) => `${index}:${model.id}:${hashModel(model)}`).join('|')
+
+const EXPECTED = {
+  version: 1,
+  hash: '0:019af08a-c27b-7074-8aac-95315d1ef3fd:-1vf2pk|1:019e70af-e5b2-76d0-9ede-f22d8265bb14:i4q5q9|2:019e7580-2b0c-77d6-8b99-16a99abe4591:b3y2hj|3:019e7580-2b0e-719c-a43f-d2b56e7f31b4:-g7x2jr',
+}
+
+describe('defaultModels version snapshot', () => {
```
ital0, line 28: Heads up, I don't think this file ever runs in CI. The `test` and `test:5x` scripts in package.json discover shared tests with the glob `shared/*.test.ts`, which doesn't descend into subdirectories, so `shared/defaults/` is invisible to it. Since the whole version bump discipline leans on this test failing loudly, could we add the path to those scripts so it actually gates changes?

src/lib/reconcile-defaults.ts
```diff
@@ -143,13 +191,27 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+
+export const reconcileDefaults = async (db: AnyDrizzleDatabase, overrides?: ReconcileDefaultsOverrides) => {
+  const modelsSource = overrides?.models ?? bundledModelsDefaults
+
   await db.transaction(async (tx) => {
+    // Version gate for models: only overwrite rows when our defaults source
+    // is strictly newer than the highest version ever applied on this account.
+    // Prevents older-bundle devices from downgrading newer synced rows (THU-637).
+    const storedModelsVersion = await readAppliedVersion(tx, modelsVersionKey)
+    const canOverwriteModels = modelsSource.version > (storedModelsVersion ?? Number.NEGATIVE_INFINITY)
```
ital0: One scenario worth acknowledging here: a fresh second device whose initial sync times out reads no stored version, so this ends up true and the device can still overwrite newer rows and regress the marker once. It self heals on the next boot of the newer device, so the infinite ping pong is gone, but the gate is best effort within the initial sync window. Maybe worth a short comment saying so, or treating a timed out first ever sync as skip overwrites for that boot?

src/hooks/use-app-initialization.ts
```diff
@@ -159,9 +161,17 @@ const executeInitializationSteps = async (httpClient?: HttpClient): Promise<Hand
   // the outcome is reported in the app_init_timing event below.
   const initialSyncOutcome = await time('step3_wait_for_initial_sync', () => database.waitForInitialSync())
 
+  // Step 3.5: Settle the /config fetch so reconcile can prefer server-shipped
+  // defaults when they declare a higher version than the bundle. fetchConfig
+  // has an internal 5s timeout and never rejects, so this is bounded; when
+  // offline, the persisted store value from the last successful fetch is used
+  // (or the bundle wins on a fresh install with no cached config).
+  await fetchConfigPromise
```
ital0: Previously this promise settled in parallel and was only awaited after tray and posthog init, so reconcile never waited on the network. Now everything from step 4 on serializes behind it, worst case around 5s when the backend hangs rather than failing fast. Probably fine since it has usually resolved by this point, just checking the tradeoff was intentional. Skipping the await when a persisted config already exists could soften it if boot telemetry ever shows a regression.

src/lib/pick-defaults.ts
```diff
@@ -0,0 +1,29 @@
...
+ * higher declared version. Behaves like an OTA channel: the server can hot-ship
+ * new defaults without a client build; when offline / unreachable / behind, the
+ * bundle wins.
+ *
+ * Rollback semantics are monotonic — a server that regresses its declared
+ * version below what the client already has will not overwrite. To retract a
+ * bad server-published set, ship a *higher* version with the reverted content.
+ */
+export const pickModelsDefaults = (server: ServerModelsDefaults | undefined): ModelsDefaults => {
+  if (server && server.version > defaultModelsVersion) {
```
ital0: Since this is where server data starts winning over the bundle, would a tiny sanity check on the payload be worth it (finite version, array with content)? Today the backend builds it from the same shared constants so nothing can go wrong, but the OTA intent is precisely to hand ship payloads later, and a malformed one would be reconciled into user databases and stamped as applied.

backend/src/api/config.ts
```diff
@@ -20,4 +26,10 @@ export const createConfigRoutes = (settings: Settings) =>
     allowCustomAgents: settings.allowCustomAgents,
     // Omit when unset so the frontend treats it as "no enforcement" without parsing an empty string as semver.
     minAppVersion: settings.minAppVersion || undefined,
+    defaults: {
```
ital0, line 29: Tiny thought: `SharedModel` includes an `apiKey` field and this endpoint is public. Every default ships it as null today so nothing leaks, but nothing structurally stops a future server shipped payload from populating it. Stripping the field here, or a comment noting the constraint, would make that a non issue.

shared/defaults/models.ts
```diff
@@ -2,14 +2,49 @@
...
+  contextWindow: number | null
+  deletedAt: string | null
+  defaultHash: string | null
+  vendor: string | null
+  description: string | null
+  userId: string | null
+  apiKey: string | null
+}
+
+const hashValues = (values: (string | number | null | undefined)[]): string => {
```
ital0: This is now a byte for byte copy of `hashValues` in the frontend utils, and both are load bearing since the stored hashes in user databases depend on the exact output. If someone tweaks one copy and not the other we get silent drift. Could the frontend utils import this one instead, so there's a single source?

shared/defaults/models.ts
```diff
@@ -2,14 +2,49 @@
...
 
-import { hashValues } from '@/lib/utils'
-import type { Model } from '@/types'
+/**
+ * Shape of a shipped model default. Structurally compatible with the
+ * frontend `Model` type (`src/types.ts`) so consumers can pass these values
+ * anywhere a `Model` is expected. Kept self-contained so this file has no
+ * dependency on frontend code and can be imported by backend and shared code.
+ */
+export type SharedModel = {
```
ital0, line 18: The comment promises structural compatibility with `Model`, but nothing enforces it, so a new required field on `Model` would drift silently. A one line assignability check in frontend code or a test would let the compiler stand guard on that promise.

src/lib/reconcile-defaults.ts
```diff
@@ -27,13 +52,18 @@ import { nowIso } from './utils'
...
+ *   Set to false when the caller's defaults version is older than what has
+ *   already been applied on this account (see THU-637 / AGENTS.md).
  */
 export const reconcileDefaultsForTable = async <T extends { defaultHash: string | null }>(
   db: AnyDrizzleDatabase,
   table: SQLiteTableWithColumns<any>,
   defaults: readonly T[],
   hashFn: (item: any) => string,
   keyField: string = 'id',
+  canOverwrite: boolean = true,
```
ital0: Thinking ahead to the follow up where the other tables get the same gate: a sixth positional argument reads fine once, but call sites like `(tx, table, data, hash, 'id', false)` get opaque, and each gated table also threads more parameters through `cleanupRemovedDefaults`. Not blocking at all, but shaping this as a small options object now might save churn when the pattern repeats five more times.

shared/defaults/models.test.ts
```diff
@@ -0,0 +1,35 @@
...
+ *   2. Update `EXPECTED` below to match the actual values from the failure.
+ *
+ * The version is the ordering signal reconcile uses to decide who owns the
+ * newest defaults across devices (THU-637). Changing defaults without bumping
+ * the version breaks that ordering silently.
+ */
+const computeSnapshotHash = () =>
+  defaultModels.map((model, index) => `${index}:${model.id}:${hashModel(model)}`).join('|')
+
+const EXPECTED = {
```
ital0: Small one: house style is camelCase even for module constants, so `expected` here.

backend/src/db/client.ts
```diff
@@ -26,11 +26,32 @@ const postgresUrl = isPglite
...
+// src/index.ts` with mixed env). `new PGlite('postgresql://...')` then
+// treats the connection string as a path and bootstraps a real Postgres
+// data dir into `backend/postgresql:/postgres:postgres@localhost:.../...`.
+// Detect the schema and treat connection-string values as "no path given"
+// (i.e. in-memory PGlite).
+const isPostgresConnectionUrl = (url: string | undefined): boolean =>
+  typeof url === 'string' && /^(?:postgres|postgresql):\/\//.test(url)
+
+const pgliteDataDir =
+  isPglite && process.env.DATABASE_URL && !isPostgresConnectionUrl(process.env.DATABASE_URL)
```
ital0, line 42: Nice guard. One small idea: when the connection string gets ignored we silently fall back to in memory. A single console.warn would save the next person some head scratching if they expected a persistent data dir.

src-tauri/tauri.conf.json
```diff
@@ -23,7 +23,7 @@
     "security": {
       "csp": {
         "default-src": "'self' tauri: asset:",
-        "connect-src": "'self' ipc: http://ipc.localhost https: wss: ws://localhost:8000 http://localhost:8000 http://localhost:11434",
+        "connect-src": "'self' ipc: http://ipc.localhost https: wss: ws://localhost:8000 http://localhost:8000 ws://localhost:8080 http://localhost:8080 http://localhost:11434",
```
ital0, line 26: No strong feelings since port 8000 set the precedent, but these dev origins do ship in the production CSP. If we ever want to tighten it, Tauri supports a separate `devCsp` so dev builds can carry their own policy. Safe to ignore for this PR.

ital0 (changes requested):

src/lib/reconcile-defaults.ts
```diff
@@ -102,10 +146,20 @@ export const reconcileDefaultsForTable = async <T extends { defaultHash: string
...
  * current defaults arrays. The hash-match guard ensures edited rows survive.
  * Skips rows without a `defaultHash` (user-created).
+ *
+ * When `canOverwriteModels` is false, the models scan is skipped: our defaults
+ * are older than what another device already applied, so we don't know which
+ * "unrecognized" rows are legitimate additions vs. genuinely-removed defaults.
+ * The profiles scan still runs but is a natural no-op in this case — no models
+ * were soft-deleted, so every profile's parent stays alive.
  */
-export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in e3c6bf67.

shared/defaults/models.test.ts
raivieiraadriano92: Fixed in cb970f93.

src/lib/reconcile-defaults.ts
raivieiraadriano92: Fixed in e3c6bf67.

src/hooks/use-app-initialization.ts
raivieiraadriano92: Fixed in 50a14b58.

src/lib/pick-defaults.ts
raivieiraadriano92: Fixed in 50a14b58.

backend/src/api/config.ts
raivieiraadriano92: Fixed in 85b9b203.

shared/defaults/models.ts
raivieiraadriano92: Fixed in 85b9b203.

shared/defaults/models.ts
raivieiraadriano92: Fixed in 85b9b203.

src/lib/reconcile-defaults.ts
raivieiraadriano92: Fixed in 12c13eb7.

shared/defaults/models.test.ts
raivieiraadriano92: Fixed in cb970f93.

backend/src/db/client.ts
raivieiraadriano92: Fixed in 50a14b58.

src-tauri/tauri.conf.json
raivieiraadriano92: Deferred to a follow-up per your note.

src/lib/reconcile-defaults.ts
```diff
@@ -165,7 +285,21 @@ export const reconcileDefaults = async (db: AnyDrizzleDatabase) => {
...
+      // row existence — not on parsed version — so a pre-existing row with a
+      // non-numeric value (data corruption, older schema) still routes to
+      // UPDATE instead of hitting a primary-key conflict on INSERT.
+      const versionValue = String(modelsSource.version)
+      if (!stored.exists) {
+        await tx.insert(settingsTable).values({ key: modelsVersionKey, value: versionValue })
+      } else {
+        await tx.update(settingsTable).set({ value: versionValue }).where(eq(settingsTable.key, modelsVersionKey))
+      }
+    }
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 7d7a0f63.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +213,66 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+        return rawCanOverwrite
+      }
+      if (initialSyncCompleted) {
+        return rawCanOverwrite
+      }
+      if (!stored.exists) {
+        return false
+      }
+      return rawCanOverwrite
+    })()
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 7d7a0f63.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +213,66 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+    // Model profiles ship 1:1 with models and mutate together in practice, so
+    // they ride the same gate — otherwise an older-bundle device would revert
+    // profile settings (temperature, tools, addenda) that a newer bundle just
+    // shipped alongside its model changes, reintroducing THU-637 on the profile
+    // side. Insert-of-missing still runs regardless, so orphaned profiles are
+    // impossible even when overwrites are skipped.
+    await reconcileDefaultsForTable(tx, modelProfilesTable, defaultModelProfiles, hashModelProfile, {
+      keyField: 'modelId',
+      canOverwrite: canOverwriteModels,
+    })
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 7d7a0f63.

src/lib/reconcile-defaults.ts
```diff
@@ -11,48 +11,150 @@ import { v7 as uuidv7 } from 'uuid'
...
+    // (see the PR 1044 ital0 discussion / THU-637 rollout race).
+    if (
+      (existing as { deletedAt?: string | null }).deletedAt &&
+      existing.defaultHash &&
+      hashFn({ ...existing, deletedAt: null }) === existing.defaultHash
+    ) {
+      await db.update(table).set({ deletedAt: null }).where(eq(table[keyField], keyValue))
+      mutated = true
+      continue
+    }
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 4ff2a59b.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +261,70 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
   }
 }
 
-export const reconcileDefaults = async (db: AnyDrizzleDatabase) => {
+export type ReconcileDefaultsOverrides = {
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 4ff2a59b.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +261,70 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+    // a newer bundle just shipped alongside its model changes, reintroducing
+    // THU-637 on the profile side. `insertMissing: true` ensures a model
+    // present locally (via sync) always has a profile to pair with, even when
+    // canOverwrite is closed. The rare orphaned-profile edge (bundle contains
+    // a profile whose model the newer version retired) is silent — with no
+    // matching model row, DAL joins won't surface the orphan.
+    const profilesPass = await reconcileDefaultsForTable(
+      tx,
+      modelProfilesTable,
+      defaultModelProfiles,
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in be630ff7.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +283,98 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+          `${droppedOtaModelIds.join(', ')}. OTA can only re-version or retire models this bundle knows; ` +
+          `adding a new model id requires a client build so its profile ships alongside.`,
+      )
+    }
+
     // Soft-delete removed system defaults before reconciling current ones.
-    await cleanupRemovedDefaults(tx)
+    // Cleanup uses the unfiltered OTA set so any id the server still ships
+    // (including new-to-us ones synced from a newer-bundle peer) stays alive.
+    await cleanupRemovedDefaults(tx, canOverwriteModels, modelsSource.data, initialSyncCompleted)
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in a50c7162.

src/lib/reconcile-defaults.ts
```diff
@@ -143,16 +283,98 @@ export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
...
+    // Model profiles ship 1:1 with models and mutate together in practice, so
+    // they ride the same authority gate as models — otherwise an older-bundle
+    // device would revert profile settings (temperature, tools, addenda) that
+    // a newer bundle just shipped alongside its model changes, reintroducing
+    // THU-637 on the profile side. `insertMissing: true` ensures a bundle-
+    // known model always has its bundled profile to pair with, even when
+    // canOverwrite is closed. Bounded to bundle-known model ids: OTA-only-new
+    // ids were dropped from the models pass above (they have no bundled
+    // profile to insert here either).
+    const profilesPass = await reconcileDefaultsForTable(
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d97d297d.

src/hooks/use-app-initialization.ts
```diff
@@ -214,9 +232,8 @@ const executeInitializationSteps = async (httpClient?: HttpClient): Promise<Hand
   // of each other) — run in parallel; each wrapper swallows its own failure.
   const [tray, posthogClient] = await Promise.all([initializeTraySafely(), initializePostHogSafely(client)])
 
-  // Settle step 0 so its duration lands in the timing payload. fetchConfig
-  // never rejects (catch block + 5s timeout) and has typically resolved by now.
-  await fetchConfigPromise
+  // (fetchConfigPromise was already awaited at step 3.5 so its duration lands
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d97d297d.

shared/defaults/models.ts
```diff
@@ -0,0 +1,165 @@
...
+  userId: null,
+}
+
+/**
+ * Flash ships under a fresh id — not the retired V4 Pro id. Reusing Pro's id
+ * would flip `isConfidential` 1 → 0 on threads that were created encrypted
+ * (`isEncrypted` mirrors the model's `isConfidential` at creation), stranding
+ * them because the model picker and send guard both enforce
+ * `isEncrypted === isConfidential`. The retired Pro row is instead
+ * soft-deleted by `cleanupRemovedDefaults`, so encrypted threads bound to it
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d97d297d.

backend/src/db/client.ts
```diff
@@ -26,11 +26,40 @@ const postgresUrl = isPglite
...
+    ? process.env.DATABASE_URL
+    : undefined
+
+if (isPglite && process.env.DATABASE_URL && isPostgresConnectionUrl(process.env.DATABASE_URL)) {
+  console.warn(
+    `[db] DATABASE_DRIVER=pglite but DATABASE_URL is a postgres connection string ` +
+      `(${process.env.DATABASE_URL}) — falling back to in-memory PGlite. ` +
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d97d297d.

src/lib/reconcile-defaults.ts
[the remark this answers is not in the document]
raivieiraadriano92: Deferred to a follow-up.

src/lib/reconcile-defaults.ts
```diff
@@ -95,17 +212,35 @@ export const reconcileDefaultsForTable = async <T extends { defaultHash: string
...
+ *   - `initialSyncCompleted`: guards the profile scan. The profile loop
+ *     trusts `aliveModelIds` derived from local state; that trust is only
+ *     warranted when sync has finished. Otherwise a parent model whose
+ *     alive-state hasn't synced yet could look "missing" locally and cause us
+ *     to soft-delete a profile whose parent is actually still alive on cloud.
  */
-export const cleanupRemovedDefaults = async (db: AnyDrizzleDatabase) => {
+export const cleanupRemovedDefaults = async (
+  db: AnyDrizzleDatabase,
+  canOverwriteModels: boolean = true,
```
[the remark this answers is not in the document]
raivieiraadriano92: Deferred to a follow-up.

PR #1055, opened 2026-07-05
feat(chat): add render_html artifact tool with live preview

cjroth:
- New in-app agent tool `render_html` renders a self-contained HTML page as an inline artifact, with a top-right toggle to a resizable side panel.
- Verifies before display: fast static JS/CSS syntax checks (acorn + css-tree, lazy-imported so they stay out of the entry bundle), then a hidden sandboxed-iframe runtime check (uncaught errors, unhandled rejections, failed subresource loads). Failures return to the model in-turn for self-correction.
- Live scripts-off preview while the HTML streams; scripts run once verified, on-screen (IntersectionObserver, latched), and after a short load-settle delay.
- Sandboxed `allow-scripts` iframe, never `allow-same-origin`; single network knob.

---

> [!NOTE]
> **Medium Risk**
> Runs untrusted model HTML in sandboxed iframes with CSP (documented residual: in-frame navigation); large new surface area in chat rendering and tool execution, though verification gates visible output.
> 
> **Overview**
> Introduces **`render_html`**, a core agent tool (always registered in `getAvailableTools`) that shows self-contained HTML as a first-class chat artifact instead of a collapsed tool row.
> 
> **Verification before display:** lazy-loaded **acorn** / **css-tree** static checks (syntax, blocked external/module imports), then a hidden **`allow-scripts`** iframe pass using a shared **harness** (ready / height / error `postMessage` + per-render nonce). Failed runs stay in the tool group; streaming or verified-ok calls are lifted out for **`ArtifactMessagePart`** / **`InlineArtifactCard`**.
> 
> **Rendering UX:** scripts-off **CSP-only preview** while HTML streams; scripts run after verification when the card is on-screen (`useOnScreen`) and after a one-shot **`useAppSettled`** delay. Users can copy/download HTML (downloads get the offline CSP), open a **side panel** (`ArtifactSidebarContent`), and see runtime errors via **`ArtifactErrorStrip`**. Chat layout tweaks drop the last-message min-height reserve for artifact messages and full-width scroll with centered content.
> 
> **Tests & CI:** broad unit coverage, Playwright **`artifact`** project for real-browser harness behavior, and CI/local **`test:5x`** / backend test steps use a very large per-test timeout so slow PGlite/CI runs don’t flake (step timeouts remain the real limit).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d8f85e5790bbe9704f205c273c1ba853c914c4a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/artifacts/static-check.ts
```diff
@@ -0,0 +1,146 @@
...
+  const scripts = [...doc.querySelectorAll('script')]
+    .filter((el) => {
+      if (el.getAttribute('src') || (el.textContent ?? '').trim().length === 0) {
+        return false
+      }
+      return jsScriptTypes.has((el.getAttribute('type') ?? '').trim().toLowerCase())
+    })
+    .map((el) => ({
+      code: el.textContent ?? '',
+      module: (el.getAttribute('type') ?? '').trim().toLowerCase() === 'module',
```
ital0, line 57: A `<script type="module">` importing a CDN URL passes the static check, fails silently at runtime, and still verifies ok. Could we flag import/export sources as blocked resources too?

src/artifacts/verify-html.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+      if (data.type === 'artifact-ready') {
+        // Handle only the FIRST ready: cancel the hard timeout and open a short grace window
+        // for a late async error. Ignoring repeat 'ready' messages avoids stacking grace timers
+        // and — since the hard timer is now gone — stops a page deferring completion forever by
+        // re-sending 'ready'.
+        if (graceTimer === undefined) {
+          clearTimeout(hardTimer)
+          graceTimer = setTimeout(() => finish({ ok: true, errors: [] }), readyGraceMs)
+        }
```
ital0, line 87: An error arriving after the 250ms grace window still resolves `ok: true`. Intended?

src/components/artifact/artifact-actions.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * Header controls shared by the inline artifact card and the side-panel view:
+ * copy the HTML source (with the same copy→check fade as the preview's copy-URL
+ * button) and download it as a standalone `.html` file. Clicks stop propagation
+ * so they never toggle the surrounding (collapsible) card header.
+ */
+export const ArtifactActions = ({ html, title, buttonClassName }: ArtifactActionsProps) => {
+  const { copy, isCopied } = useCopyToClipboard()
+
+  const handleDownload = () => {
+    const url = URL.createObjectURL(new Blob([html], { type: 'text/html' }))
```
ital0: This downloads the raw HTML, which runs unsandboxed when opened. Did you consider exporting the wrapped (offline CSP) version?

src/components/chat/inline-artifact-card.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <span className="truncate">{runtimeError}</span>
+            </div>
+          )}
+          {showContent && (
+            <SandboxedHtmlFrame
+              html={html}
+              title={title}
+              autoHeight
+              allowScripts={!streaming && active}
+              onReady={() => setRuntimeError(null)}
```
ital0: The harness sends the error before `ready`, so this can wipe an error the user never saw. Maybe only clear on a reload boundary? Same in `artifact-sidebar-content.tsx`.

src/lib/assistant-message.ts
```diff
@@ -90,6 +103,15 @@ export const groupMessageParts = (parts: GroupableUIPart[]): GroupedUIPart[] =>
   // Both typed `tool-<name>` parts and MCP `dynamic-tool` parts group as tool items.
   parts.forEach((part) => {
     if (isToolOrDynamicToolUIPart(part)) {
+      // An artifact renders on its own (live preview while streaming, then the
+      // verified result). Before it has HTML — or if it failed — it stays in the
```
ital0: This comment doesn't match the code anymore: streaming parts lift out regardless of HTML. Mind updating it?

src/components/chat/artifact-message-part.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * Renders a `render_html` tool call as a first-class artifact in the transcript.
+ * From the moment the call starts it shows an inline card that streams into place
+ * (scripts off) and then becomes the interactive artifact once verified. It lives
+ * in exactly one place at a time: inline, OR the side panel — while it's open in
+ * the panel the transcript shows a slim placeholder instead. A finished-but-failed
+ * call renders nothing here (it stays an ordinary tool call in the group).
+ */
+export const ArtifactMessagePart = ({ part }: ArtifactMessagePartProps) => {
+  const artifactId = part.toolCallId
+  const input = (part.input ?? {}) as Partial<RenderHtmlInput>
```
ital0: There are a few of these casts on the tool payload across three files. A shared `isRenderHtmlPart` guard would keep them in sync.

src/artifacts/harness.ts
```diff
@@ -0,0 +1,171 @@
...
+ * an artifact can neither reach the parent origin nor exfiltrate over the network.
+ *
+ * Residual limitation (no clean CSP/sandbox token closes it): a script can still
+ * navigate its OWN frame (`location = ...`, `<meta http-equiv=refresh>`), which
+ * issues an outbound GET the fetch directives never see — so an artifact must not
+ * be trusted with sensitive user-entered input. Set to `null` to lift the network
+ * restriction (e.g. to allow CDN libraries).
+ */
+export const artifactCsp: string | null =
+  "default-src 'none'; script-src 'unsafe-inline' 'unsafe-eval'; style-src 'unsafe-inline'; img-src data: blob:; font-src data: blob:; media-src data: blob:; worker-src blob:; base-uri 'none'; form-action 'none'"
```
ital0, line 35: The `| null` is never used. I'd drop it until a real toggle exists.

package.json
```diff
@@ -156,6 +158,7 @@
     "@testing-library/jest-dom": "^6.9.1",
     "@testing-library/react": "^16.3.0",
     "@types/bun": "^1.2.20",
+    "@types/css-tree": "^2.3.11",
```
ital0, line 161: Types are v2 but runtime is v3. Worth confirming the `parse` signature still matches.

src/components/chat/inline-artifact-card.tsx
```diff
@@ -0,0 +1,158 @@
...
+        className={cn(
+          'grid transition-[grid-template-rows] duration-200 ease-out',
+          open ? 'grid-rows-[1fr]' : 'grid-rows-[0fr]',
+        )}
+      >
+        {/* min-h-0 lets the row collapse to exactly 0 — without it the grid leaves a ~1px
+            sliver of the frame above the card's bottom border when closed. */}
+        <div className="min-h-0 overflow-hidden">
+          {!streaming && runtimeError && (
+            <div className="flex items-center gap-2 border-b border-border bg-destructive/10 px-4 py-1.5 text-xs text-destructive">
```
ital0: Nit: this strip is duplicated in `artifact-sidebar-content.tsx`. Could be a tiny shared component.

src/components/artifact/sandboxed-html-frame.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { formatHarnessError, parseHarnessMessage, wrapArtifactHtml, wrapArtifactPreviewHtml } from '@/artifacts/harness'
+import { cn } from '@/lib/utils'
+import { useEffect, useMemo, useRef, useState } from 'react'
+
+/** Height used before the page reports its own, and the floor/ceiling for the reported height. */
+const defaultAutoHeightPx = 400
+const minAutoHeightPx = 60
+// Ceiling so a page (which knows its own nonce) can't report a huge height and blow out the transcript.
+const maxAutoHeightPx = 20000
```
ital0: Nit: `20_000` for readability.

ital0 (changes requested):
Really solid design overall. A few things I'd like to resolve before merging, plus some smaller suggestions.

src/artifacts/verify-html.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { describe, expect, it, mock } from 'bun:test'
+import { getClock } from '@/testing-library'
+import { runIframeVerification, verifyArtifactHtml } from './verify-html'
+
+const validHtml = '<!doctype html><html><head></head><body><h1>ok</h1></body></html>'
+
+describe('verifyArtifactHtml', () => {
+  it('short-circuits on a static syntax error without running the iframe pass', async () => {
+    const runtime = mock(async () => ({ ok: true, errors: [] }))
+    const result = await verifyArtifactHtml('<script>const x = ;</script>', { runtime })
```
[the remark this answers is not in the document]
cjroth: False positive. `runtime` here is a test stub (a `mock`) and the string is a literal passed to `verifyArtifactHtml`, which never injects it into the DOM in this path — the real render happens in a sandboxed iframe (`allow-scripts`, no `allow-same-origin`). It's already suppressed inline with a `// nosemgrep` annotation on the line above. Resolving as a false positive.

src/components/chat/inline-artifact-card.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const active = settled && onScreen
+  const showContent = useMemo(() => !streaming || hasRenderableBody(html), [streaming, html])
+
+  // While generating the header is inert: no collapse, no hover, not focusable.
+  const interactive = !streaming
+  const toggle = () => setOpen((prev) => !prev)
+  const onHeaderKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+    // Only the header itself toggles — not Enter/Space on a focused action button inside it.
+    if (event.target !== event.currentTarget) {
+      return
```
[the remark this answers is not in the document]
cjroth: Fixed in 00dc05a2. The toggle is now a real `<button>` (flex-1, edge-to-edge padding) with the copy/download/panel buttons as DOM siblings, so there's no button-inside-a-button — screen readers see one toggle, and Enter/Space work natively. Dropped the manual `role`/`tabIndex`/keydown handler.

src/components/ui/expandable.tsx
```diff
@@ -62,7 +62,7 @@ export const Expandable = ({
         <AccordionPrimitive.Header className="flex">
           <AccordionPrimitive.Trigger
             className={cn(
-              'flex flex-1 items-center justify-between gap-2 px-4 py-2 text-left transition-all outline-none min-h-[var(--min-touch-height)]',
+              'flex flex-1 cursor-pointer items-center justify-between gap-2 px-4 py-2 text-left transition-all outline-none min-h-10',
```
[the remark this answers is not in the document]
cjroth: Reverted in 00dc05a2 — back to `min-h-[var(--min-touch-height)]` so desktop keeps its 0 floor everywhere `Expandable` is used (only the `cursor-pointer` addition stays). Thanks for catching the shared-component regression.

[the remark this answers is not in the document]
cjroth: Reverted in 00dc05a2 — back to `min-h-[var(--min-touch-height)]` so desktop keeps its 0 floor everywhere `Expandable` is used (only the `cursor-pointer` addition stays). Thanks for catching the shared-component regression.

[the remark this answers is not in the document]
cjroth: Reverted in 00dc05a2 — back to `min-h-[var(--min-touch-height)]` so desktop keeps its 0 floor everywhere `Expandable` is used (only the `cursor-pointer` addition stays). Thanks for catching the shared-component regression.

src/components/chat/inline-artifact-card.tsx
```diff
@@ -0,0 +1,158 @@
...
+          )}
+          {showContent && (
+            <SandboxedHtmlFrame
+              html={html}
+              title={title}
+              autoHeight
+              allowScripts={!streaming && active}
+              onReady={() => setRuntimeError(null)}
+              onError={streaming ? undefined : setRuntimeError}
+            />
```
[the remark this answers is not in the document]
cjroth: By design, with a small improvement in 00dc05a2. A scripts-off preview iframe is sandboxed without `allow-same-origin`, so the parent physically cannot measure its content height — the harness that reports height only runs when scripts are on. So the streaming preview shows at the default height until scripts activate on verification, at which point it auto-sizes. I did add a reload-boundary height reset in `sandboxed-html-frame` so the preview→active transition re-measures cleanly instead of holding the stale default.

src/components/artifact/sandboxed-html-frame.tsx
```diff
@@ -0,0 +1,106 @@
...
+  title,
+  className,
+  allowScripts = true,
+  autoHeight = false,
+  onReady,
+  onError,
+}: SandboxedHtmlFrameProps) => {
+  const iframeRef = useRef<HTMLIFrameElement>(null)
+  // One nonce per mounted frame; correlates the harness's messages with this iframe.
+  const nonce = useMemo(() => crypto.randomUUID(), [])
```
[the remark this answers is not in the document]
cjroth: Fixed in 00dc05a2 — switched to `useState(() => crypto.randomUUID())`. useMemo is only a hint (React may drop and recompute it), which would regenerate the nonce, reload the iframe, and re-key the message listener; useState is a real stability guarantee.

src/artifacts/static-check.ts
```diff
@@ -0,0 +1,146 @@
...
+  // it renders broken, but nothing throws, so flag them here rather than let verification pass blank.
+  const externalResources = [
+    ...[...doc.querySelectorAll('script[src]')]
+      .map((el) => el.getAttribute('src') ?? '')
+      .filter(isExternalUrl)
+      .map((src) => `<script src="${src}">`),
+    ...[...doc.querySelectorAll('link[rel~="stylesheet"][href]')]
+      .map((el) => el.getAttribute('href') ?? '')
+      .filter(isExternalUrl)
+      .map((href) => `<link rel="stylesheet" href="${href}">`),
```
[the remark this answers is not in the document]
cjroth: Fixed in 00dc05a2. The check now flags every `<script src>` and stylesheet `<link href>` regardless of URL form. The offline CSP (`script-src 'unsafe-inline' 'unsafe-eval'`, `style-src 'unsafe-inline'`) names no scheme or host, so relative, root-relative, protocol-relative, and even `data:` script/stylesheet loads all fail — all are now surfaced instead of silently rendering blank.

src/components/artifact/sandboxed-html-frame.tsx
```diff
@@ -0,0 +1,106 @@
...
+    return () => window.removeEventListener('message', handleMessage)
+  }, [nonce])
+
+  return (
+    <iframe
+      ref={iframeRef}
+      title={title}
+      sandbox={allowScripts ? 'allow-scripts' : ''}
+      srcDoc={srcDoc}
+      style={autoHeight ? { height: contentHeight ?? defaultAutoHeightPx } : undefined}
```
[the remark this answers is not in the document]
cjroth: Fixed in 00dc05a2 — `contentHeight` now resets at each reload boundary (srcDoc change) via a render-time ref, so a document swap or a streaming→active transition re-measures instead of keeping the previous artifact's height.

src/hooks/use-throttled-value.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { useEffect, useRef, useState } from 'react'
+
+/**
+ * Returns a copy of `value` that updates at most once per `intervalMs` (trailing
+ * edge), always settling on the latest value. Useful for feeding rapidly
+ * changing data (e.g. a streaming string) into an expensive sink without
+ * re-rendering it on every change.
+ */
+export const useThrottledValue = <T>(value: T, intervalMs: number): T => {
```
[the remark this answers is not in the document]
cjroth: Agreed — dropped `useThrottledValue` and its test in 00dc05a2 and switched the streaming preview to the existing `useThrottle`.

src/components/chat/assistant-message.tsx
```diff
@@ -177,6 +188,14 @@ export const AssistantMessage = memo(
       [message.parts],
     )
 
+    // An artifact is a tall block of its own, so it doesn't need the last-message
+    // height reserve — and applying it would leave empty space that pushes the
+    // artifact toward the middle of the viewport ("floating" too far down).
+    const hasArtifact = useMemo(
+      () => message.parts.some((part) => isToolOrDynamicToolUIPart(part) && getToolName(part) === renderHtmlToolName),
```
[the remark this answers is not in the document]
cjroth: Fixed in 00dc05a2 — `hasArtifact` now reuses `artifactRendersStandalone` (via the shared `isRenderHtmlPart` guard), so it flips true only for the streaming/verified-ok cases that actually lift the card out. A last message whose only `render_html` call failed verification now keeps the normal height reserve.

src/artifacts/static-check.ts
cjroth: Fixed in 00dc05a2. `checkScripts` now walks the acorn AST and flags every module specifier — static `import`/`export … from`, `export * from`, and dynamic `import('…')` with a literal — as a blocked resource, since none of them resolve in an offline artifact (CDN, bare, or relative).

src/artifacts/verify-html.ts
cjroth: Intended. Runtime verification is a bounded, one-shot check: "does the page load and initialize without throwing?" We wait for `ready` plus a 250ms grace for an immediate async error, then settle. Errors that surface later (e.g. on user interaction) aren't verification failures — the *visible* frame carries the same harness and surfaces them live via `onError` (the red error strip on the card/panel). Keeping the window open indefinitely would just delay or hang verification for every artifact.

src/components/artifact/artifact-actions.tsx
cjroth: Good call — fixed in 00dc05a2. The download now wraps the HTML with the same offline CSP the app renders it under (`wrapArtifactPreviewHtml`), so opening the saved file in a browser can't run it unsandboxed with network access. Copy still yields the raw author markup for editing/re-use.

src/components/chat/inline-artifact-card.tsx
cjroth: Fixed in 00dc05a2. Dropped the clear-on-`ready` (the harness reports load-time errors *before* `ready`, so it wiped errors the user never saw). Both the inline card and the side panel now clear the error only at a reload boundary (the document changed), via a render-time ref.

src/lib/assistant-message.ts
cjroth: Updated the comment in 00dc05a2 — it now says a `render_html` part lifts out from the moment the call starts (a scripts-off preview streams in even before any HTML arrives), and only a finished-but-failed or errored call stays in the group as an ordinary tool call.

src/components/chat/artifact-message-part.tsx
cjroth: Done in 00dc05a2 — added `isRenderHtmlPart` plus typed `renderHtmlInput`/`renderHtmlOutput` accessors in `render-html-tool.ts`, and routed all three call sites (`artifact-message-part.tsx`, `assistant-message.ts`, `assistant-message.tsx`) through them, removing the hand casts.

src/artifacts/harness.ts
cjroth: Dropped in 00dc05a2 — `artifactCsp` is now just `string` and `cspMetaTag` always emits the tag. Removed the "set to null" escape hatch until something actually needs it.

package.json
cjroth: Checked — `@types/css-tree` has no v3 release published (latest is 2.3.11), so v2 is the newest available. The only signature we rely on is `parse(source, { onParseError })`, which is unchanged between css-tree 2 and 3 (`onParseError` has existed since 2.x), and our `css-tree-parser.d.ts` shim borrows exactly that type for the lean `css-tree/parser` entry we import. Type-check and the static-check tests both pass against the v3 runtime. Will bump the types if/when a v3 is published.

src/components/chat/inline-artifact-card.tsx
cjroth: Done in 00dc05a2 — extracted `ArtifactErrorStrip` and used it in both the inline card and the side panel.

src/components/artifact/sandboxed-html-frame.tsx
cjroth: Done in 00dc05a2 — `20_000`.

src/artifacts/static-check.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  it('flags a relative module import too (nothing resolves offline)', async () => {
+    const html = page('', '<script type="module">import { x } from "./x.js"; console.log(x)</script>')
+    const issues = await staticCheckHtml(html)
+    expect(issues.some((i) => i.source === 'resource')).toBe(true)
+  })
+
+  it('flags a dynamic import() with a string literal specifier', async () => {
+    const html = page('', '<script>import("https://cdn.example.com/lib.js").then(() => {})</script>')
+    const issues = await staticCheckHtml(html)
+    expect(issues.some((i) => i.source === 'resource' && i.message.includes('cdn.example.com'))).toBe(true)
```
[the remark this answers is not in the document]
cjroth: False positive — this is a test assertion, not URL sanitization. The `.includes(...)` was checking that the flagged error message quotes the import specifier; it never gates any security decision. Rewritten in 6d8f85e5 to assert via the `toContain` matcher on the found issue's message instead, which is both clearer and doesn't read as host validation. Resolving.

PR #1058, opened 2026-07-07
perf(bundle): lazy-load posthog-js so analytics leaves the entry chunk

cjroth:
## What & why
The PostHog SDK (~62KB gzip) shipped in the **landing entry chunk** — `src/lib/posthog.tsx` imported `posthog-js` statically and `trackEvent` is used across the chat critical path. Analytics is post-paint by definition, so it has no business blocking first paint.

## The fix (behavior-preserving)
- `import('posthog-js')` dynamically inside the already-async `initPosthog()`.
- Replace the `posthog-js/react` `PostHogProvider`/`usePostHog` with a tiny in-house context (`usePostHogClient()`) carrying the nullable client.
- Non-React alias call sites use a `getPosthogClient()` getter — the `alias` identity-stitching for `anonymous_user_promoted` is unchanged (same client, same args/timing).
- Init options, `before_send` sanitization, and consent (`opt_out_capturing_by_default`) unchanged.

## Before / after
| metric | before | after | Δ |
| --- | --- | --- | --- |
| entry chunk (gzip) | 788.1 KB | 726.0 KB | **−62 KB** |

posthog-js now lives in its own async chunk, loaded after first paint.

## Verified
- `tsc` + lint green; analytics + anonymous-promotion tests pass (20/0).
- `grep "from 'posthog-js'"` (excluding `import type`) → empty; no static SDK import remains.
- chat-landing focus run (headless Chromium): **0 console errors, 0 page errors** — deferral throws nothing.

## Reviewer checklist
- [x] Single concern (analytics deferral); no behavior change.
- [x] Consent + alias identity-stitching preserved.
- [x] Real before/after (build measured).
- [x] No unrelated files.


src/lib/analytics/use-anonymous-promotion-analytics.ts
```diff
@@ -23,7 +22,7 @@ export const pendingAnonIdKey = 'thunderbolt_pending_anon_id'
 // If the client was never initialized (self-hosted, no key), alias() is a silent
 // no-op because posthog-js handles uninitialized state.
 const defaultAlias = (newUserId: string, anonId: string) => {
-  posthog.alias(newUserId, anonId)
+  getPosthogClient()?.alias(newUserId, anonId)
```
ital0, line 26: Tiny thing: the comment just above still says the silent skip happens because the SDK handles uninitialized state, but after this change it happens because getPosthogClient() returns null before init.

src/lib/posthog.tsx
```diff
@@ -75,6 +81,7 @@ export const initPosthog = async (httpClient?: HttpClient): Promise<HandleResult
     const apiHost = `${cloudUrl}/posthog`
 
     if (!posthogClient) {
+      const { default: posthog } = await import('posthog-js')
```
ital0, line 84: Just checking the tradeoff here: app init awaits this step before setting initData, so the SDK chunk now downloads on the startup critical path and only after the config request returns. If that shows up in app_init_timing, kicking off the import in parallel with the config fetch would overlap the two network trips. Fine to keep as is if the numbers look good.

src/lib/posthog.tsx
```diff
@@ -130,13 +137,24 @@ const TelemetryAvailableContext = createContext(false)
  */
 export const useTelemetryAvailable = () => useContext(TelemetryAvailableContext)
 
+const PostHogClientContext = createContext<PostHog | null>(null)
```
ital0, line 131: Now that the client itself travels through context, useTelemetryAvailable is derivable from it, since telemetry available just means client is not null. Two parallel contexts carrying the same information feels like a bit of duplication. No strong feelings, just flagging in case you want to fold them into one.

ital0 (dismissed):

src/lib/analytics/use-anonymous-promotion-analytics.ts
cjroth: Done in 1a45dce2 — reworded to: null client (self-hosted with no key, or before initPosthog resolves) → the optional chain makes alias() a silent no-op.

src/lib/posthog.tsx
cjroth: Done in 1a45dce2 — dropped `TelemetryAvailableContext` entirely and made `useTelemetryAvailable` derive from the client context (`usePostHogClient() !== null`), so there is a single source of truth. `PostHogProvider` now wraps just the one context. Public API (`useTelemetryAvailable`, `usePostHogClient`) is unchanged, so no consumer changes.

src/lib/posthog.tsx
cjroth: Good catch on the tradeoff — I kept it serial on purpose. Kicking off `import('posthog-js')` in parallel with the config fetch would download the SDK chunk *before* we know whether an API key exists, so self-hosted deployments with no key would pull the SDK for nothing — which undercuts the exact property this PR establishes (the SDK never loads when analytics is off). And this step runs a beat after first paint (analytics isn't awaited before the UI renders), so the serial cost is off the paint-blocking path. If app_init_timing shows the config→import hop mattering for the hosted app, we can speculatively prefetch the chunk only in the hosted build. Left as-is for now.

PR #1059, opened 2026-07-07
perf(bundle): lazy-load KaTeX and drop the duplicate markdown parser

cjroth:
## What & why
`memoized-markdown.tsx` renders **every** chat message and carried two entry-chunk costs most messages never use.

## The fix
1. **Lazy-load KaTeX (~80KB).** `rehype-katex` + `remark-math` + the katex CSS were static (loaded for everyone). Now loaded via a module-cached dynamic `import()`, gated per-block by the file's existing math-span regexes. Non-math blocks never request the chunk; the first math block shows raw `$…$` briefly, then re-renders into KaTeX. `remark-gfm` stays static.
2. **Drop `marked` (a duplicate parser).** `marked` was used only to split messages into top-level blocks; react-markdown already parses with remark/micromark. Replaced `marked.lexer` with the same remark pipeline (`unified` + `remark-parse` + `remark-gfm`), slicing top-level nodes by mdast position offsets — a 1:1 behavioral port.
3. **Bug fix.** The shared global math-span regexes leaked `lastIndex` between blocks, intermittently corrupting currency-escaping of digit-leading math (`$3.14$`). Reset before + after probing.

## Before / after
| metric | before | after | Δ |
| --- | --- | --- | --- |
| entry chunk (gzip) | 726.0 KB | 633.9 KB | **−92 KB** |

KaTeX is now its own chunk; `marked` is gone from the entry.

## Verified
- All **18** memoized-markdown tests pass (adapted to `await` the now-async lazy load; no assertions weakened).
- `tsc` green. Headless Chromium: a math message renders `.katex` + requests the katex chunk; a plain message makes **0** katex requests.

## Follow-up (intentionally not in this PR)
`bun remove marked && bun add unified remark-parse` to formalize the direct deps (both already present via react-markdown 10) — kept out to avoid entangling the harness's unrelated uncommitted `package.json`/`bun.lock` tooling changes.

## Reviewer checklist
- [x] Single concern (slim the markdown renderer).
- [x] Hot-path correctness: full test suite passes; math still renders (headless-verified).
- [x] Real before/after (build measured).
- [x] No unrelated files.


src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,104 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+type MathPlugins = { remark: PluggableList[number]; rehype: PluggableList[number] }
+
+// remark-math + rehype-katex + the KaTeX stylesheet make up their own chunk,
+// loaded once (module-level cached promise) and reused across every block, and
+// only when a block needs math — so the ~70KB never ships in the entry for the
+// many messages that render no equations. The CSS is imported here so the
+// stylesheet lands in the katex chunk rather than the entry.
+let mathPluginsPromise: Promise<MathPlugins> | null = null
+const loadMathPlugins = (): Promise<MathPlugins> => {
+  mathPluginsPromise ??= Promise.all([
```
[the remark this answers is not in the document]
cjroth: Fixed in 63549c97 — `loadMathPlugins` now clears `mathPluginsPromise` in a `catch` before rethrowing, so a flaky/stale chunk load can retry on a later block instead of staying broken for the session.

src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,104 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+    import('remark-math'),
+    import('rehype-katex'),
+    import('katex/dist/katex.min.css'),
+  ]).then(([remark, rehype]) => ({ remark: remark.default, rehype: rehype.default }))
```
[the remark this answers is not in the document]
cjroth: Done in 63549c97 — `loadMathPlugins` is now an `async` function using `await Promise.all([...])`, which also gives the try/catch a natural home for the failure reset above.

src/components/chat/memoized-markdown.tsx
```diff
@@ -2,25 +2,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { marked } from 'marked'
```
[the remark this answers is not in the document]
cjroth: Removed `marked` from package.json and the lockfile (`bun remove marked`) in 63549c97 — confirmed nothing else imports it.

src/components/chat/memoized-markdown.tsx
```diff
@@ -2,25 +2,25 @@
...
-import { type CSSProperties, memo, useMemo } from 'react'
+import { type CSSProperties, memo, useEffect, useMemo, useState } from 'react'
 import type { Components } from 'react-markdown'
 import ReactMarkdown from 'react-markdown'
-import rehypeKatex from 'rehype-katex'
 import remarkGfm from 'remark-gfm'
-import remarkMath from 'remark-math'
+import remarkParse from 'remark-parse'
```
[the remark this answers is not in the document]
cjroth: Good call — added `unified@^11.0.5` and `remark-parse@^11.0.0` to `dependencies` in e645a98e (pinned to the versions already resolved via react-markdown, so no duplicate in the tree). No more phantom dependency.

src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,114 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+// block renders without math (a brief FOUC of raw `$…$`, then a re-render into
+// KaTeX). The cancelled flag guards against a setState after unmount.
+const useMathPlugins = (enabled: boolean): MathPlugins | null => {
+  const [plugins, setPlugins] = useState<MathPlugins | null>(null)
+  useEffect(() => {
+    if (!enabled || plugins) {
+      return
+    }
+    let cancelled = false
+    loadMathPlugins().then((loaded) => {
```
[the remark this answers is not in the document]
cjroth: Fixed the unhandled rejection in e645a98e — the load now has a `.catch` that logs the failure. I intentionally didn't add an in-place retry: the effect has no back-off, so retrying on every render would busy-loop against a persistently-failing chunk. Recovery still happens because `loadMathPlugins` clears its cached promise on failure, so any freshly-mounted block re-attempts the import. Happy to add an explicit error-state + manual retry button later if we see this in practice.

src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,114 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+// (already-normalized) content — the same spans `escapeCurrencyDollars` uses, so
+// detection agrees with what remark-math would render. `.test()` on a global
+// regex advances (and leaks) `lastIndex`, and `matchAll` reads that `lastIndex`
+// at creation — so a leftover value would corrupt the currency-escaping spans on
+// the next block. Reset both before and after probing to keep the shared regexes
+// clean and detection stable.
+const blockHasMath = (content: string): boolean => {
+  displayMathSpan.lastIndex = 0
+  inlineMathSpan.lastIndex = 0
+  const hasMath = displayMathSpan.test(content) || inlineMathSpan.test(content)
```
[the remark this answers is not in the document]
cjroth: Closed in e645a98e — `parseMarkdownIntoBlocks` now returns `{ content, isCode }` and the renderer skips `blockHasMath` for `code` nodes, so a fenced block that only *shows* `$$…$$` no longer pulls the KaTeX chunk. Rendered output is unchanged (remark-math renders nothing in code anyway).

src/components/chat/memoized-markdown.test.tsx
```diff
@@ -51,30 +71,35 @@ describe('MemoizedMarkdown — LaTeX', () => {
...
     expect(container.querySelector('.katex')).not.toBeNull()
     expect(container.textContent).not.toContain('\\(')
   })
 
-  it('renders LaTeX `\\[…\\]` display delimiters as a KaTeX display block', () => {
+  it('renders LaTeX `\\[…\\]` display delimiters as a KaTeX display block', async () => {
     const { container } = renderMarkdown('\\[ x = \\frac{-b \\pm \\sqrt{b^2 - 4ac}}{2a} \\]')
+    await flushLazyLoad()
     expect(container.querySelector('.katex-display')).not.toBeNull()
     expect(container.textContent).not.toContain('\\[')
```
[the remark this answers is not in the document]
cjroth: Fixed in a0fdce25 — reworded the comment to "the block is code, so math detection is skipped and the plugins never load." Good catch; the old wording was left over from before the `isCode` short-circuit landed.

src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,129 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+// block renders without math (a brief FOUC of raw `$…$`, then a re-render into
+// KaTeX). The cancelled flag guards against a setState after unmount.
+const useMathPlugins = (enabled: boolean): MathPlugins | null => {
+  const [plugins, setPlugins] = useState<MathPlugins | null>(null)
+  useEffect(() => {
+    if (!enabled || plugins) {
+      return
+    }
+    let cancelled = false
+    loadMathPlugins()
```
ital0: Tiny style point: house rules prefer async/await over `.then` and `.catch` chains. An inner async function called from the effect would match the convention, though I get that effects nudge you toward this shape. Fine to keep as is if you prefer it.

src/components/chat/memoized-markdown.tsx
```diff
@@ -136,30 +136,129 @@ const normalizeDisplayMath = (markdown: string): string => {
...
+// renderer skip math detection on fenced/indented code, so a block that merely
+// *shows* `$$…$$` as source never pulls the KaTeX chunk (remark-math renders
+// nothing there anyway).
+type Block = { content: string; isCode: boolean }
+
+// Slice each top-level mdast node back out of `source` by its position offsets,
+// reproducing the exact block strings without a second markdown parser.
+const sliceTopLevelBlocks = (source: string): Block[] =>
+  markdownParser.parse(source).children.map((node) => ({
+    content: source.slice(node.position?.start.offset ?? 0, node.position?.end.offset ?? 0),
```
ital0: If a node ever came back without position data, the `?? 0` fallbacks here and in `parseMarkdownIntoBlocks` would silently produce an empty block and drop content, which would be painful to debug. Since remark always sets positions at parse time, failing loudly might be safer than a quiet wrong value. Flagging with no strong feelings.

src/components/chat/memoized-markdown.test.tsx
```diff
@@ -51,30 +71,36 @@ describe('MemoizedMarkdown — LaTeX', () => {
...
-  it('leaves a `$$…$$` line inside a fenced code block as literal source', () => {
+  it('leaves a `$$…$$` line inside a fenced code block as literal source', async () => {
     const { container } = renderMarkdown('```\n$$E = mc^2$$\n```')
-    // Not rendered as math, and the source text is preserved verbatim (not
-    // rewritten with extra newlines).
+    // The whole message parses to a single `code` node, so math detection is
+    // skipped (`isCode` short-circuits `hasMath`) and the plugins never load.
+    // Even if they did, remark-math never treats code-fence contents as math —
+    // so no `.katex` is emitted and the source text is preserved verbatim.
+    await flushLazyLoad()
```
ital0, line 111: Nice that this negative test flushes before asserting nothing rendered. The other negative tests (lone dollar sign, two currency amounts, inline code span) still assert `.katex` is null without flushing, so they would keep passing even if math detection regressed and KaTeX rendered after the async load. Worth adding `await flushLazyLoad()` there too so they keep proving what they used to.

src/components/chat/memoized-markdown.test.tsx
```diff
@@ -2,44 +2,64 @@
...
     <ExternalLinkDialogProvider>
       <MemoizedMarkdown id="test" content={content} />
     </ExternalLinkDialogProvider>,
   )
 
+// KaTeX is lazy-loaded (its ~70KB chunk ships only when a block has math), so
+// math renders after an async re-render rather than synchronously. Flush the
+// pending dynamic imports + setState by draining the fake clock inside `act`.
+const flushLazyLoad = async () => {
+  for (let i = 0; i < 5; i++) {
```
ital0, line 32: Small question: why 5 iterations?

ital0 (changes requested):

src/components/chat/memoized-markdown.tsx
cjroth: Done in 37c38b7f — the effect now runs an inner `async` `load()` with `try/catch` instead of the `.then/.catch` chain, matching the house async/await convention.

src/components/chat/memoized-markdown.tsx
cjroth: Done in 37c38b7f — replaced the `?? 0` fallbacks with a `nodeOffsets` helper that throws if a node ever lacks position offsets, so a missing position fails loudly instead of silently slicing an empty block and dropping content. Used at both slice sites.

src/components/chat/memoized-markdown.test.tsx
cjroth: Done in 37c38b7f — added `await flushLazyLoad()` to the negative tests (lone dollar sign, inline code span, two currency amounts, and the empty `\( \)` case) so a detection regression that loaded KaTeX after the async import would now fail them.

src/components/chat/memoized-markdown.test.tsx
cjroth: Explained in 37c38b7f — added a comment: the load is a chain of async boundaries (effect → loadMathPlugins → Promise.all of three dynamic imports → setState → re-render), and each drained tick can schedule the next, so one pass is not enough. 5 is a comfortable upper bound for the ~3–4-deep chain.

PR #1068, opened 2026-07-08
refactor(THU-653): make PowerSync JWT kid env-driven

raivieiraadriano92:
## Summary

Makes the JWT `kid` in `deploy/config/powersync-config.yaml` env-driven (`!env PS_JWT_KID`) so the same PowerSync image can validate JWTs signed with any backend's existing `POWERSYNC_JWT_KID` value. Unblocks the operational migration of production PowerSync from PowerSync Cloud to a self-hosted service on Render — the two Render resources (`powersync` Web Service + `powersync-storage` Postgres) will be provisioned manually via the Render dashboard, no repo changes needed.

Also removes an orphan `mcp_servers` sync rule from `deploy/config/powersync-config.yaml` and `deploy/k8s/templates/configmaps.yaml` — it was never in `shared/powersync-tables.ts` or the backend schema.

## Behavior preservation

- Preview + enterprise Pulumi stacks continue to work: `services.ts` and `shared.ts` now inject `PS_JWT_KID: enterprise-powersync` (the value the yaml previously hardcoded).
- The enterprise `docker-compose.yml` does the same for the local self-host stack.
- The k8s Helm chart is unaffected — its config already uses a Helm-templated `kid` (`{{ .Values.powersync.jwt.kid }}`).

## Test plan

- Local enterprise `docker compose up postgres powersync` — clean boot, sync rules load, replication starts, JWT validation works end-to-end (401 on unauth/malformed, valid HS256 JWT with matching kid reaches the post-auth Fastify layer).
- Locally-built `deploy/docker/powersync.Dockerfile` image tested identically.

### DIFF
--- a/deploy/pulumi/src/shared.ts
+++ b/deploy/pulumi/src/shared.ts
@@ -492,6 +492,9 @@ export const createSharedStack = (args: SharedStackArgs): SharedStackOutputs => 
             name: 'POWERSYNC_JWT_KEY_BASE64',
             value: args.powersyncJwtSecret.apply((s) => Buffer.from(s).toString('base64')),
           },
+          // Label paired with the JWT secret. Must match the backend's
+          // POWERSYNC_JWT_KID so PowerSync can find the right key when validating.
+          { name: 'PS_JWT_KID', value: 'enterprise-powersync' },
         ],
         portMappings: [{ containerPort: 8080 }],
         logConfiguration: logConfig('powersync'),
--- a/deploy/config/powersync-config.yaml
+++ b/deploy/config/powersync-config.yaml
@@ -30,7 +30,6 @@ sync_rules:
         data:
           - SELECT * FROM powersync.chat_messages WHERE user_id = bucket.user_id
           - SELECT * FROM powersync.tasks WHERE user_id = bucket.user_id
-          - SELECT * FROM powersync.mcp_servers WHERE user_id = bucket.user_id
           - SELECT * FROM powersync.prompts WHERE user_id = bucket.user_id
           - SELECT * FROM powersync.skills WHERE user_id = bucket.user_id
           - SELECT * FROM powersync.triggers WHERE user_id = bucket.user_id
@@ -42,15 +41,14 @@ client_auth:
   jwks:
     keys:
       - kty: oct
-        # Base64 of POWERSYNC_JWT_SECRET. Sourced from PS_JWT_KEY_BASE64 at runtime so
-        # the value can rotate per environment without rebuilding the image. The PS_
-        # prefix is mandatory — PowerSync's YAML loader rejects !env substitutions for
-        # variables that don't start with it (security guardrail).
-        # Local default: base64("enterprise-thunderbolt-powersync-jwt-default-secret").
-        # Preview stacks (Pulumi) inject base64 of the per-stack random JWT secret.
+        # Both k (secret) and kid (label) are env-driven so the same image can serve
+        # multiple environments without a rebuild. Must match the values the backend
+        # signs with: k = base64(POWERSYNC_JWT_SECRET), kid = POWERSYNC_JWT_KID.
+        # The PS_ prefix is mandatory — PowerSync's YAML loader rejects !env
+        # substitutions for variables that don't start with it (security guardrail).
         # Backend validates the underlying secret is >=32 chars when decoded.
         k: !env PS_JWT_KEY_BASE64
         alg: HS256
-        kid: enterprise-powersync
+        kid: !env PS_JWT_KID
   telemetry:
     disable_telemetry_sharing: true
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -93,6 +93,8 @@ services:
       # Default value below = base64("enterprise-thunderbolt-powersync-jwt-default-secret"),
       # which matches the default backend POWERSYNC_JWT_SECRET.
       PS_JWT_KEY_BASE64: ZW50ZXJwcmlzZS10aHVuZGVyYm9sdC1wb3dlcnN5bmMtand0LWRlZmF1bHQtc2VjcmV0
+      # Label paired with the JWT secret. Must match the backend's POWERSYNC_JWT_KID.
+      PS_JWT_KID: enterprise-powersync
     volumes:
       - ./config/powersync-config.yaml:/config/config.yaml
     ports:
--- a/deploy/k8s/templates/configmaps.yaml
+++ b/deploy/k8s/templates/configmaps.yaml
@@ -38,7 +38,6 @@ data:
             data:
               - SELECT * FROM powersync.chat_messages WHERE user_id = bucket.user_id
               - SELECT * FROM powersync.tasks WHERE user_id = bucket.user_id
-              - SELECT * FROM powersync.mcp_servers WHERE user_id = bucket.user_id
               - SELECT * FROM powersync.prompts WHERE user_id = bucket.user_id
               - SELECT * FROM powersync.skills WHERE user_id = bucket.user_id
               - SELECT * FROM powersync.triggers WHERE user_id = bucket.user_id
--- a/deploy/pulumi/src/services.ts
+++ b/deploy/pulumi/src/services.ts
@@ -271,6 +271,9 @@ export const createServices = (args: ServiceArgs) => {
           // The PS_ prefix is mandatory: PowerSync's YAML loader rejects !env vars
           // that don't start with it.
           { name: 'PS_JWT_KEY_BASE64', value: args.secrets.powersyncJwtSecret.apply((s) => Buffer.from(s).toString('base64')) },
+          // Label paired with PS_JWT_KEY_BASE64. Must match the backend's
+          // POWERSYNC_JWT_KID so PowerSync can find the right key when validating.
+          { name: 'PS_JWT_KID', value: 'enterprise-powersync' },
         ],
         portMappings: [{ containerPort: 8080 }],
         logConfiguration: logConfig('powersync'),


deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+     ```bash
+     printf '%s' "$POWERSYNC_JWT_SECRET" | base64
+     ```
+     (`printf`, not `echo` — avoid a trailing newline.)
+   - `PS_JWT_KID` — the backend's `POWERSYNC_JWT_KID` value verbatim.
+4. Wait for the `powersync` service to reach healthy. Watch logs for
+   `PostgresLockManager.init` to complete, followed by replication start-up
+   and LSN catch-up. Confirm bucket state populates:
+   ```
+   psql "$PS_STORAGE_URI" -c "SELECT count(*) FROM bucket_data"
```
ital0: (suggestion) Two things make this verify step fail as pasted, the storage URI is a dashboard only var holding an internal URL, so it is not in the operator's shell, and the storage tables live in a powersync schema, so the query likely wants `powersync.bucket_data`. Maybe show it as run from a Render shell with the schema spelled out.

deploy/render/render.yaml
```diff
@@ -0,0 +1,51 @@
...
+services:
+  - type: web
+    name: powersync
+    runtime: image
+    plan: standard
+    region: ohio
+    autoDeployTrigger: "off"
+    healthCheckPath: /probes/liveness
+    image:
+      url: ghcr.io/thunderbird/thunderbolt-powersync:0.1.106
```
ital0: I don't think this image path can pull. The publish workflow tags everything under `ghcr.io/thunderbird/thunderbolt`, so the package actually lives at `ghcr.io/thunderbird/thunderbolt/thunderbolt-powersync` (the same nested path Pulumi uses in shared.ts:464). As written the service can never start.

deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+The frontend connects to `powersync.<domain>` directly — the URL is served by
+`backend/src/api/powersync.ts` in `/powersync/token`, so no frontend rebuild is
+needed to cut over.
+
+## One-off source-Postgres prep
+
+Runs against the existing consumer Postgres (the app's primary DB). Not a
+Drizzle migration — this is server-side plumbing.
+
+1. In the Render dashboard for the source Postgres, enable **logical
```
ital0: Heads up on a Render gotcha: logical replication is not something we can enable ourselves, it takes a support ticket, and `CREATE PUBLICATION ... FOR ALL TABLES` needs superuser rights that only Render support has (render.com/docs/postgresql-logical-replication). Steps 1 and 3 probably need reworking around that ticket.

deploy/pulumi/src/shared.ts
```diff
@@ -492,6 +492,9 @@ export const createSharedStack = (args: SharedStackArgs): SharedStackOutputs =>
             name: 'POWERSYNC_JWT_KEY_BASE64',
             value: args.powersyncJwtSecret.apply((s) => Buffer.from(s).toString('base64')),
           },
+          // Label paired with the JWT secret. Must match the backend's
+          // POWERSYNC_JWT_KID so PowerSync can find the right key when validating.
+          { name: 'PS_JWT_KID', value: 'enterprise-powersync' },
```
ital0, line 497: While you're in this env block: the config baked into the image reads the key from `!env PS_JWT_KEY_BASE64`, but this container sets `POWERSYNC_JWT_KEY_BASE64` (line 492), so the key never resolves and the new kid can't help the shared stack. Predates this PR, but it's a single rename, and the comment above misquotes the config too.

AGENTS.md
```diff
@@ -97,8 +97,8 @@ See [docs/architecture/e2e-encryption.md](docs/architecture/e2e-encryption.md) f
 
 **Deploying new synced tables (two-PR process):**
 
-1. **PR 1 (backend-only):** Backend schema, Drizzle migration, `shared/powersync-tables.ts`, and `config.yaml` sync rule. Merge → run migration → update PowerSync Cloud dashboard rules.
-2. **PR 2 (frontend + everything else):** Frontend schema, DAL, defaults, reconciliation, and any UI/logic. Merge only after PR 1's dashboard rules are live.
+1. **PR 1 (backend + sync rules):** Backend schema, Drizzle migration, `shared/powersync-tables.ts`, and both sync-rule configs (`powersync-service/config/config.yaml` and `deploy/config/powersync-config.yaml`). Merge → run migration → wait for `images-publish.yml` to publish the new `ghcr.io/thunderbird/thunderbolt-powersync` image → roll the Render PowerSync service to the new tag.
```
ital0: Tiny path fix here (and in powersync-account-devices.md lines 59 and 69): the published package is `ghcr.io/thunderbird/thunderbolt/thunderbolt-powersync`. The repo segment is easy to drop by accident when copying.

deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+To rotate later:
+
+1. Generate a fresh secret (`openssl rand -base64 32`).
+2. Update backend env `POWERSYNC_JWT_SECRET` on Render. Restart backend.
+3. Update PowerSync service env `PS_JWT_KEY_BASE64` on Render = base64 of the
+   new secret. The service auto-redeploys.
+
+Both restarts should complete within ~1 min of each other; brief mismatch
+just causes clients to retry — SDK handles it. If you also change the kid,
```
ital0: I think this step contradicts the nice change this PR just made. The kid now comes from `!env PS_JWT_KID` and render.yaml lets you edit it in the dashboard, so rotating it is just the two env updates, no merge or image rebuild.

docs/architecture/powersync-account-devices.md
```diff
@@ -54,21 +54,24 @@ See [docs/composite-primary-keys-and-default-data.md](composite-primary-keys-and
 2. **Backend schema**: Add only a `user_id` index: `index('idx_[table]_user_id').on(table.userId)`. Do not add composite foreign keys or other indexes (see above).
 3. Register in [src/db/powersync/schema.ts](../src/db/powersync/schema.ts) (`drizzleSchema`).
 4. Add the table name and query keys in [shared/powersync-tables.ts](../shared/powersync-tables.ts) (`POWERSYNC_TABLE_NAMES` and `powersyncTableToQueryKeys`).
-5. Update [powersync-service/config/config.yaml](../powersync-service/config/config.yaml): add a line under `sync_rules.content` → `bucket_definitions.user_data.data` (e.g. `- SELECT * FROM my_table WHERE my_table.user_id = bucket.user_id`).
+5. Update **both** sync-rule configs so local, preview, and prod stay in parity:
```
ital0: There's actually a third live copy of the sync rules: the k8s configmap template embeds its own bucket definitions (it still has the `mcp_servers` rule this PR removes, and it already tracks skills and agents, so it's maintained). Might be worth saying "all three" here so enterprise k8s doesn't silently drift.

deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+     ```
+     (`printf`, not `echo` — avoid a trailing newline.)
+   - `PS_JWT_KID` — the backend's `POWERSYNC_JWT_KID` value verbatim.
+4. Wait for the `powersync` service to reach healthy. Watch logs for
+   `PostgresLockManager.init` to complete, followed by replication start-up
+   and LSN catch-up. Confirm bucket state populates:
+   ```
+   psql "$PS_STORAGE_URI" -c "SELECT count(*) FROM bucket_data"
+   ```
+5. Point Cloudflare DNS record `powersync.<thunderbolt-domain>` at the
```
ital0: One more step needed here: Render only routes hostnames that were added and verified as custom domains on the service, so the Cloudflare record alone won't cut it (render.com/docs/custom-domains).

deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+   ```
+
+   `FOR ALL TABLES` mirrors the container init script; PowerSync filters via
+   sync rules regardless.
+
+## Blueprint bootstrap
+
+Both the JWT secret (`k`) and its label (`kid`) in `deploy/config/powersync-config.yaml` are env-driven (`!env PS_JWT_KEY_BASE64` and `!env PS_JWT_KID`). We plug the *existing* backend values straight in — no backend change, no client refresh.
+
+1. `render blueprint launch deploy/render/render.yaml` (or apply via the
```
ital0: The CLI only ships `render blueprints validate`, there's no launch command, and actual deploys go through the dashboard. Maybe flip the parenthetical and make the dashboard the primary path.

deploy/render/README.md
ital0: This one won't work as pasted: the storage URI only exists as an env var on the service (and it's the internal URL), and the tables live in the `powersync` schema. Showing it as `powersync.bucket_data` run from a Render shell would make it something you can paste directly.

AGENTS.md
ital0: Small operational trap: the workflow tags by package.json version, so a sync rules merge overwrites `0.1.106` in place rather than minting a new tag, and a later unrelated merge can overwrite it again before the manual roll. A version bump per sync rules change, or pinning by digest, would make the redeploy deterministic.

docs/architecture/powersync-account-devices.md
```diff
@@ -10,7 +10,7 @@ This document consolidates documentation for:
 
 ## 1. PowerSync Overview
 
-PowerSync provides offline-first sync between the backend (PostgreSQL) and clients (SQLite). Data is scoped by `user_id` from the JWT. The backend issues PowerSync JWTs and can apply client uploads (PUT/PATCH/DELETE) to Postgres. Production uses PowerSync Cloud; local development uses the Docker stack in `powersync-service/`.
+PowerSync provides offline-first sync between the backend (PostgreSQL) and clients (SQLite). Data is scoped by `user_id` from the JWT. The backend issues PowerSync JWTs and can apply client uploads (PUT/PATCH/DELETE) to Postgres. Production runs the self-hosted PowerSync service on Render (see [deploy/render/README.md](../../deploy/render/README.md)); local development uses the Docker stack in `powersync-service/`. The frontend never hard-codes the PowerSync URL — the backend returns it in `/powersync/token`, so URL changes are transparent to clients.
```
ital0: A few sibling docs still point at the PowerSync Cloud dashboard: docs/multi-device-sync.md:73, docs/architecture/multi-device-sync.md:73, docs/self-hosting/configuration.md:61, and the powersync-sync-reviewer agent (line 36), plus the Cloud labels in the sync middleware diagrams. Worth sweeping them into this PR's docs pass?

docs/architecture/powersync-account-devices.md
```diff
@@ -54,21 +54,24 @@ See [docs/composite-primary-keys-and-default-data.md](composite-primary-keys-and
 2. **Backend schema**: Add only a `user_id` index: `index('idx_[table]_user_id').on(table.userId)`. Do not add composite foreign keys or other indexes (see above).
 3. Register in [src/db/powersync/schema.ts](../src/db/powersync/schema.ts) (`drizzleSchema`).
 4. Add the table name and query keys in [shared/powersync-tables.ts](../shared/powersync-tables.ts) (`POWERSYNC_TABLE_NAMES` and `powersyncTableToQueryKeys`).
```
ital0, line 56: The header comment in `shared/powersync-tables.ts` (the file that calls itself the single source of truth) still lists only the local config in its checklist for adding a table. Probably wants the same "both configs" treatment this PR gives everywhere else.

deploy/render/render.yaml
```diff
@@ -0,0 +1,51 @@
...
+#
+# Region coupling: `region:` on BOTH resources below AND the source Postgres
+# (declared outside this blueprint) must be identical. Cross-region breaks
+# `fromDatabase` internal-URL wiring, forces sslmode=require, and adds
+# per-packet WAN latency. Update both `region:` values together if the
+# source DB ever moves.
+
+databases:
+  - name: powersync-storage
+    plan: basic-256mb
```
ital0: Was the smallest plan deliberate for prod bucket storage? Given we're migrating because Cloud's storage ran out, the defaults here are a fixed 15 GB disk and 256 MB RAM with no autoscaling. Just want to make sure that's sized on purpose.

deploy/render/render.yaml
```diff
@@ -0,0 +1,51 @@
...
+    region: ohio
+
+services:
+  - type: web
+    name: powersync
+    runtime: image
+    plan: standard
+    region: ohio
+    autoDeployTrigger: "off"
+    healthCheckPath: /probes/liveness
```
ital0: Curious if liveness over readiness was deliberate. It avoids restarts during a long LSN catch up, but deploys will look healthy before replication is actually ready. Happy either way, just checking it's intentional.

deploy/render/README.md
```diff
@@ -0,0 +1,173 @@
...
+   `powersync` service:
+   - `PS_PG_URI` — internal URL of the *source* Postgres, using the
+     `powersync_role` credentials from the prep step, e.g.
+     `postgresql://powersync_role:<pw>@<source-internal-host>:5432/<db>`.
+   - `PS_JWT_KEY_BASE64` — base64 of the backend's raw secret:
+     ```bash
+     printf '%s' "$POWERSYNC_JWT_SECRET" | base64
+     ```
+     (`printf`, not `echo` — avoid a trailing newline.)
+   - `PS_JWT_KID` — the backend's `POWERSYNC_JWT_KID` value verbatim.
```
ital0: Flagging just in case: the backend's prod default for the kid is an empty string (settings.ts:191), so "verbatim" only works if prod explicitly set one. A quick note to confirm the value is not empty would save someone a confusing debug session.

docs/architecture/powersync-account-devices.md
```diff
@@ -54,21 +54,24 @@ See [docs/composite-primary-keys-and-default-data.md](composite-primary-keys-and
...
-5. Update [powersync-service/config/config.yaml](../powersync-service/config/config.yaml): add a line under `sync_rules.content` → `bucket_definitions.user_data.data` (e.g. `- SELECT * FROM my_table WHERE my_table.user_id = bucket.user_id`).
+5. Update **both** sync-rule configs so local, preview, and prod stay in parity:
+   - [powersync-service/config/config.yaml](../powersync-service/config/config.yaml) — local docker-compose.
```
ital0: These two new links resolve one directory short. From docs/architecture they need `../../`, like the link you added at line 13 (the rest of the file has the same older issue, so just these two).

deploy/render/render.yaml
```diff
@@ -0,0 +1,51 @@
...
+    image:
+      url: ghcr.io/thunderbird/thunderbolt-powersync:0.1.106
+    envVars:
+      # Source Postgres — the app's primary DB, provisioned outside this blueprint.
+      # Set via the Render dashboard to the *internal* connection URL of the source
+      # Postgres, using a REPLICATION-privileged role (see README.md § Source DB).
```
ital0: Tiny one: the README doesn't have a "Source DB" section, the heading is "One-off source-Postgres prep".

docs/architecture/powersync-account-devices.md
ital0: Really nice migration design, by the way. Serving the URL from the token endpoint while keeping the secret and kid stable makes cutover a single env flip with a real glide window. I traced the whole path and it holds together.

ital0 (changes requested):

deploy/render/render.yaml
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

AGENTS.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

docs/architecture/powersync-account-devices.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

docs/architecture/powersync-account-devices.md
raivieiraadriano92: Fixed in 17edf2c5.

docs/architecture/powersync-account-devices.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/README.md
raivieiraadriano92: Fixed in 17edf2c5.

docs/architecture/powersync-account-devices.md
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/render.yaml
raivieiraadriano92: Fixed in 17edf2c5.

deploy/render/render.yaml
raivieiraadriano92: Fixed in 17edf2c5 — bumped to basic-1gb + storageAutoscalingEnabled.

deploy/render/render.yaml
raivieiraadriano92: Kept liveness intentionally; added a comment explaining why. Fixed in 17edf2c5.

deploy/pulumi/src/shared.ts
raivieiraadriano92: Pre-existing — filing a separate fix, out of scope for this PR.

AGENTS.md
raivieiraadriano92: Accepting the risk for now; will pin-by-digest in a follow-up if it bites.

docs/architecture/powersync-account-devices.md
raivieiraadriano92: Thanks!

docs/architecture/powersync-account-devices.md
```diff
@@ -54,21 +54,24 @@ See [docs/composite-primary-keys-and-default-data.md](composite-primary-keys-and
...
+   - [deploy/config/powersync-config.yaml](../deploy/config/powersync-config.yaml) — baked into the `ghcr.io/thunderbird/thunderbolt-powersync` image; used by preview stacks (Pulumi) and by prod on Render.
+   Add a line under `sync_rules.content` → `bucket_definitions.user_data.data` in each (e.g. `- SELECT * FROM powersync.my_table WHERE user_id = bucket.user_id`).
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 17edf2c5.

PR #1073, opened 2026-07-10
fix(THU-676): request advertised scopes in MCP OAuth flow

darkbanjo:
## Problem

Thunderbolt's MCP OAuth flow never requested any OAuth scopes — it didn't read `scopes_supported` from the server's protected-resource metadata, didn't set `scope` on DCR registration, and didn't pass `scope` to `startAuthorization`. Every `scope` reference in the auth code only *stored* the token's returned scope; nothing ever *requested* one.

For MCP servers that don't scope-gate, a scope-less token still has full access, so this was invisible. But for servers that gate each tool behind a granular OAuth scope, the issued token is authorized for nothing → `tools/list` returns `[]` even though the connection, auth, and `capabilities.tools` are all fine.

### Repro (hosted Metabase MCP)

- OAuth completes; token is attached on `initialize` (verified on the wire).
- `initialize` → 200 with `capabilities.tools: { listChanged: true }`.
- `tools/list` → `{"tools":[]}`; the session id decodes to `{"v":1,"ui":false}` (unprivileged).
- Metabase's protected-resource metadata advertises 13 required `agent:*` scopes; our token carried none of them.

## Fix

`src/lib/mcp-auth/web-oauth-flow.ts`:
- `discoverServer` now reads `scopes_supported` from the protected-resource metadata and returns it space-delimited.
- `prepareAuthorization` passes that `scope` to **both** `registerClient` (DCR) and `startAuthorization`.
- When a server advertises no scopes, `scope` is `undefined` and the request is unchanged — no regression for non-gated servers.

Per the MCP auth spec (RFC 8707 / 9728) the client SHOULD request scopes from `scopes_supported`, so this is a correctness fix that unblocks any scope-gated OAuth MCP server, not just Metabase.

## Testing

- Added 2 unit tests: scope-gated server requests the advertised scopes on registration + authorization; non-gated server omits `scope`.
- `bun test src/lib/mcp-auth/web-oauth-flow.test.ts` → 35/35 pass.
- `bun run type-check` → clean.

Note: this requests all advertised scopes (incl. write/execute for Metabase). A read-only variant is a possible follow-up.

Linear: THU-676

### DIFF
--- a/src/lib/mcp-auth/web-oauth-flow.test.ts
+++ b/src/lib/mcp-auth/web-oauth-flow.test.ts
@@ -17,7 +17,7 @@ import {
   type WebOAuthDeps,
 } from './web-oauth-flow'
 import type { LoopbackCallbackParams } from './mcp-oauth-loopback'
-import { getMcpOAuthState, setMcpOAuthState } from './mcp-oauth-state'
+import { clearMcpOAuthState, getMcpOAuthState, setMcpOAuthState } from './mcp-oauth-state'
 
 const serverId = 'srv-1'
 const serverUrl = 'https://mcp.example.com'
@@ -55,6 +55,14 @@ beforeAll(async () => {
 
 afterAll(async () => {
   await teardownTestDatabase()
+})
+
+// Bun runs test files in one worker, so a fresh pending handshake left in
+// localStorage would leak into the next file and make isMcpOAuthCallback claim
+// callbacks that aren't ours (e.g. the deep-link routing test). Clear it after
+// every test so nothing escapes this file.
+afterEach(() => {
+  clearMcpOAuthState()
 })
 
 describe('isOAuthServer', () => {
--- a/src/lib/mcp-auth/web-oauth-flow.test.ts
+++ b/src/lib/mcp-auth/web-oauth-flow.test.ts
@@ -207,6 +207,60 @@ describe('startMcpOAuthFlow', () => {
     expect(redirectedTo).toBe(`${authServerUrl}/authorize?x=1`)
   })
 
+  it('requests the resource-advertised scopes_supported on registration and authorization', async () => {
+    const db = getDb()
+    let registeredScope: string | undefined
+    let authorizedScope: string | undefined
+
+    await startMcpOAuthFlow(
+      { db, serverId, serverUrl, fetchFn: noFetch, origin, isBackendConnected: () => false },
+      {
+        // A scope-gated resource (like Metabase) advertises the scopes its tools require.
+        discoverOAuthProtectedResourceMetadata: async () =>
+          ({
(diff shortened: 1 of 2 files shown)


src/lib/mcp-auth/web-oauth-flow.test.ts
```diff
@@ -207,6 +207,60 @@ describe('startMcpOAuthFlow', () => {
     expect(redirectedTo).toBe(`${authServerUrl}/authorize?x=1`)
   })
 
+  it('requests the resource-advertised scopes_supported on registration and authorization', async () => {
```
ital0, line 218: These flow tests finish with a fresh pending handshake still sitting in localStorage, since the suite only clears it in `beforeEach`. Bun runs test files in the same worker, so whatever the last test leaves behind can leak into the next file. I suspect that's what the failing CI test is showing: the deep link routing test received `/settings/mcp-servers`, which is exactly what `isMcpOAuthCallback` returns when it finds a leftover fresh handshake. Clearing the MCP OAuth state in an `afterEach` here should take care of it.

src/lib/mcp-auth/web-oauth-flow.ts
```diff
@@ -135,7 +135,14 @@ const discoverServer = async (
     throw new Error('Authorization server does not support PKCE S256.')
   }
 
-  return { authorizationServerUrl, metadata }
+  // RFC 9728: request the scopes the resource advertises (space-delimited). A
+  // scope-gated server issues a token authorized for nothing without this — e.g.
+  // Metabase gates every MCP tool behind an `agent:*` scope, so an unscoped token
+  // makes `tools/list` return empty. `undefined` when none are advertised, which
+  // omits the `scope` parameter and preserves behavior for non-gated servers.
+  const scope = prm.scopes_supported?.join(' ') || undefined
```
ital0, line 143: one thought: the SDK gives the scope hint from the 401 `WWW-Authenticate` header precedence over `scopes_supported`. We start from discovery so it never sees that header today, but if a server ever advertises narrower required scopes there, it would over request. 

Fine to leave as is, just flagging.

ital0 (dismissed):

src/lib/mcp-auth/web-oauth-flow.test.ts
darkbanjo: ⚡ Good call — exactly right. The last test in this file left a fresh pending handshake in localStorage, and since Bun runs test files in one worker isMcpOAuthCallback in the next file (use-deep-link-listener.test.ts) claimed the callback and returned /settings/mcp-servers. Added a top-level afterEach(clearMcpOAuthState) so nothing escapes this file. Verified both files now pass when run together.

PR #1075, opened 2026-07-13
fix(THU-678): bypass proxy for Custom providers on local URLs

### LINKED ISSUE #976: Local LLM using LM Studio fails to test and add models
Hi there 🙂

Thunderbolt locally using Docker Compose with a LM Studio instance for local models does not seem to work.  
When trying to pick a model from the list from LM Studio I get all the models, as expected, but pressing "Test Model" and "Add model" fails.  

Logs from Docker gives me this result:
```
backend-1    | {"level":30,"time":1781208638333,"pid":1,"hostname":"5f7077265e9b","event":"proxy_request","method":"POST","target_host":"localhost","status":400,"duration_ms":2,"bytes_in":0,"bytes_out":0,"user_id":"FNU1tnhElgpycSl4yIbjaIPg9tSv4RFS","request_id":"af50457f-0c3b-4e17-8368-6524f9e4302a","error_type":"ssrf"}
frontend-1   | 172.19.0.1 - - [11/Jun/2026:20:10:38 +0000] "POST /v1/proxy HTTP/1.1" 400 68 "http://localhost:3000/settings/models" "Mozilla/5.0 (X11; Linux x86_64; rv:151.0) Gecko/20100101 Firefox/151.0" "-"
``` 

The API is clearly reachable since all models are pulled in correctly, but none can be added or used.

Tested both LM Studio endpoint `/api/v1/` and OpenAI `/v1/`

## Details

Provider URL: http://localhost:1234/v1

Commit: f95c9cdb786a16e46cc609be32475c5ea9876bce
OS: Linux
Docker Compose Version: 5.1.4
Docker Version:
```
Client:
 Version:           29.5.2
 API version:       1.54
 Go version:        go1.26.3-X:nodwarf5
 Git commit:        79eb04c7d8
 Built:             Mon Jun  1 15:47:11 2026
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          29.5.2
  API version:      1.54 (minimum version 1.40)
  Go version:       go1.26.3-X:nodwarf5
  Git commit:       568f755ebe
  Built:            Mon Jun  1 15:47:11 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.3.1
  GitCommit:        64b425cf570b3b8dd1d4cc46da7c1fce65c6651a.m
 runc:
  Version:          1.4.2
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

## Screenshots

Image

Image

raivieiraadriano92:
## Summary

- Under docker-compose, LM Studio at `localhost:1234` was unreachable because the universal proxy sits in a container where `localhost` points at the container itself, not the host. Routes local Custom-provider URLs direct from the browser instead, so `localhost` means what the user expects.
- Public Custom URLs keep going through the proxy — preserves the THU-424 CORS fix for endpoints that don't set CORS headers.
- Strict hostname classifier (`localhost`, `host.docker.internal`, `::1`, RFC1918 IPv4 literals) so a domain like `localhost.evil.com` cannot trick the dispatcher into a bypass.

Closes thunderbird/thunderbolt#976

## Test plan

- [ ] Docker-compose stack: add a Custom provider with URL `http://localhost:1234/v1` (or plain `http://localhost:1234`), Load Models populates, Test Model succeeds.
- [ ] Same test with `http://127.0.0.1:1234/v1`, `http://host.docker.internal:1234`, `http://192.168.X.Y:PORT` — all should route direct.
- [ ] Public Custom URL (e.g. a hosted OpenAI-compat endpoint) still routes through backend proxy — check backend logs for a `proxy_request` line with the expected `target_host`.
- [ ] OpenAI / Anthropic / OpenRouter / Tinfoil providers unaffected — chat, model list, Test Model all still work through the proxy.
- [ ] Tauri desktop with `proxy_enabled=false` (default): Custom providers keep working as before (direct via Tauri fetch).


src/lib/is-local-url.ts
```diff
@@ -0,0 +1,46 @@
...
+ * local. A domain like `localhost.evil.com` does NOT match — the string
+ * `localhost` must be the entire hostname — so an attacker can't trick the
+ * dispatcher into bypassing the proxy for a public target. Common IPv4
+ * private ranges are covered (RFC 1918 + loopback); rare setups (IPv6 unique
+ * local, link-local, `::ffff:` mapped private) fall through to the proxy
+ * path, which is the safer default anyway.
+ */
+const exactLocalHostnames = new Set(['localhost', 'host.docker.internal'])
+
+const privateIpv4Patterns = [
```
ital0: These patterns match hostname prefixes, not IP literals, so a real DNS name like `10.evil.com` or `127.0.0.1.evil.com` classifies as local and bypasses the proxy. That's exactly the trick the docstring above says can't happen, and the tests only cover the `localhost.example.com` case. We already have an octet validated version of this check in `src/lib/mcp-url-validation.ts` (`isLocalOrPrivateHost`), so reusing it here (plus `host.docker.internal`) would fix this and avoid two divergent "is this local" policies. Adding a `10.evil.com` style case to the tests would lock it in.

src/lib/is-local-url.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Returns true when the URL points at the user's own machine or private
+ * network — the case where the browser can reach the target without our
+ * backend acting as a CORS intermediary. Used by the Custom-provider dispatch
+ * to decide whether to skip the universal proxy (see `src/ai/fetch.ts`).
+ *
+ * Strict hostname matching: only exact names and literal IP ranges count as
+ * local. A domain like `localhost.evil.com` does NOT match — the string
+ * `localhost` must be the entire hostname — so an attacker can't trick the
+ * dispatcher into bypassing the proxy for a public target. Common IPv4
+ * private ranges are covered (RFC 1918 + loopback); rare setups (IPv6 unique
```
ital0: Small thought: mDNS names like `mymac.local` are a fairly common way to reach a LAN model server, and they'll still route through the proxy (so still broken under docker compose). If that's a deliberate fall through it might be worth listing here next to the IPv6 cases.

src/lib/is-local-url.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const privateIpv4Patterns = [
+  /^127\./, // loopback (127/8)
+  /^10\./, // RFC 1918
+  /^172\.(1[6-9]|2\d|3[01])\./, // RFC 1918
+  /^192\.168\./, // RFC 1918
+]
+
+export const isLocalUrl = (rawUrl: string): boolean => {
+  let hostname: string
```
ital0: House style nit: `URL.canParse(rawUrl)` with an early return would avoid the `let` and the try/catch, same shape as `mcp-url-validation.ts`.

src/ai/fetch.ts
```diff
@@ -365,11 +367,22 @@ export const createModel = async (modelConfig: Model, getProxyFetch: () => Fetch
...
+      // Custom URLs (LM Studio at localhost:1234, Ollama, LAN model servers,
+      // etc.) skip the universal proxy so `localhost` means what the browser
+      // sees, not what the backend container sees — this fixes docker-compose
+      // where the backend can't reach the host's LM Studio. Public Custom
+      // URLs keep going through the proxy because THU-424 established that
+      // most public OpenAI-compat endpoints don't set CORS headers and would
+      // fail a direct browser fetch. `baseFetch` is the plain browser fetch
+      // (or Tauri fetch when opted in); it doesn't attach session
+      // credentials, so the local server never sees our auth.
+      const providerFetch: typeof baseFetch = isLocalUrl(modelConfig.url) ? baseFetch : getProxyFetch()
```
ital0: Also wondering about the LAN IP branch on hosted web: from an https origin, a direct fetch to `http://192.168.x.x` or `http://10.x.x.x` is blocked as mixed content (only localhost is exempt), plus Chrome's private network access checks. So the RFC1918 cases probably can't succeed from the web app, while they lose the proxy path that works when a self hosted backend shares the LAN. Limiting the bypass to loopback plus `host.docker.internal` (the actual docker compose bug) might be safer.

src/ai/fetch.ts
```diff
@@ -365,11 +367,22 @@ export const createModel = async (modelConfig: Model, getProxyFetch: () => Fetch
...
       const openaiCompatible = createOpenAICompatible({
         name: 'custom',
-        baseURL: modelConfig.url,
+        baseURL: normalizeOpenAiBaseUrl(modelConfig.url),
         apiKey: modelConfig.apiKey || undefined,
-        fetch: getProxyFetch(),
+        fetch: providerFetch,
```
ital0: `proxy-fetch.ts` documents that `typeof fetch` shapes are ambiguous across Bun/DOM libs and that `FetchFn` exists to pin that down. Since `FetchFn` is already imported in this file, typing `providerFetch` as `FetchFn` would be a bit more robust than `typeof baseFetch`.

src/lib/openai-base-url.ts
```diff
@@ -0,0 +1,17 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Canonicalises a user-supplied OpenAI-compatible base URL so the `/v1` suffix
+ * is present and there is no trailing slash. The AI SDK constructs upstream
+ * URLs as `${baseURL}${path}` where `path` is e.g. `/chat/completions`; a
+ * baseURL without `/v1` sends the request to the wrong path (404), and a
+ * baseURL with a trailing slash produces `//chat/completions` which some
+ * servers reject. Applying the same normalisation to the `Load Models` fetch
+ * and the runtime provider keeps both paths in lockstep.
```
ital0: Doc nit: the two paths share the URL normalization but not the transport. Load Models always fetches directly while the runtime proxies public Custom URLs, so a public endpoint can behave differently in settings vs chat. Softening this to "same URL normalisation" would keep the comment accurate.

ital0 (changes requested):

src/lib/is-local-url.ts
```diff
@@ -0,0 +1,46 @@
...
+const exactLocalHostnames = new Set(['localhost', 'host.docker.internal'])
+
+const privateIpv4Patterns = [
+  /^127\./, // loopback (127/8)
+  /^10\./, // RFC 1918
+  /^172\.(1[6-9]|2\d|3[01])\./, // RFC 1918
+  /^192\.168\./, // RFC 1918
+]
+
+export const isLocalUrl = (rawUrl: string): boolean => {
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d8871506.

src/ai/fetch.ts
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in d8871506.

src/lib/is-local-url.ts
raivieiraadriano92: Fixed in d8871506.

src/lib/is-local-url.ts
raivieiraadriano92: Fixed in d8871506.

src/lib/is-local-url.ts
raivieiraadriano92: Fixed in d8871506.

src/ai/fetch.ts
raivieiraadriano92: Fixed in d8871506.

src/ai/fetch.ts
raivieiraadriano92: Fixed in d8871506.

src/lib/openai-base-url.ts
raivieiraadriano92: Fixed in d8871506.

src/ai/fetch.ts
```diff
@@ -429,11 +431,26 @@ export const createModel = async (modelConfig: Model, getProxyFetch: () => Fetch
...
+      // `baseFetch` is the plain browser fetch (or Tauri fetch when opted in);
+      // it doesn't attach session credentials, so the local server never sees
+      // our auth.
+      const hostname = URL.canParse(conn.baseURL) ? new URL(conn.baseURL).hostname : ''
+      const providerFetch: FetchFn = isLoopbackHost(hostname) ? baseFetch : conn.fetch
+
       const openaiCompatible = createOpenAICompatible({
         name: 'custom',
-        baseURL: conn.baseURL,
+        baseURL: normalizeOpenAiBaseUrl(conn.baseURL),
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 417e56c7.

src/ai/fetch.ts
```diff
@@ -429,11 +431,26 @@ export const createModel = async (modelConfig: Model, getProxyFetch: () => Fetch
...
+      // Everything else — including RFC1918 LAN IPs, `host.docker.internal`,
+      // mDNS `.local` names, and true public endpoints — keeps going through
+      // the proxy. Loopback is the only cross-origin `http://` target an
+      // https-served frontend can reach anyway (mixed-content blocking), and
+      // public Custom endpoints rely on the proxy for CORS bypass (THU-424).
+      // `baseFetch` is the plain browser fetch (or Tauri fetch when opted in);
+      // it doesn't attach session credentials, so the local server never sees
+      // our auth.
+      const hostname = URL.canParse(conn.baseURL) ? new URL(conn.baseURL).hostname : ''
+      const providerFetch: FetchFn = isLoopbackHost(hostname) ? baseFetch : conn.fetch
```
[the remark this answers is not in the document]
raivieiraadriano92: Fixed in 417e56c7.

PR #1076, opened 2026-07-13
fix(THU-680): exclude libwayland-client from Linux AppImage

raivieiraadriano92:
## Summary

- Fixes `EGL_BAD_PARAMETER` when running the Linux AppImage on a Wayland compositor (Arch, some Fedora setups). The bundled `libwayland-client.so` is often incompatible with the user's compositor; excluding it forces the AppImage to load the system copy at runtime, which matches the compositor by construction.
- Sets `LINUXDEPLOY_EXCLUDED_LIBRARIES='libwayland-client.so*'` on the Linux build. Also pre-places the current linuxdeploy `continuous` release in Tauri's cache — Tauri's pinned binary predates the env-var feature (added upstream in linuxdeploy@98f393c4, 2024-06-13), so pre-placing the newer binary makes Tauri's `if !exists() { download }` guard skip the outdated download and the env var actually takes effect.
- Mirrors both changes in `test-build.yml`, plus fixes two pre-existing pipeline gaps that blocked validating this: missing Linux deps (`librsvg2-dev`, `squashfs-tools`, `libfuse2`) and an upload glob (`target/*/release/bundle/**`) that never matched native Linux builds.

Closes thunderbird/thunderbolt#955

## Test plan

- [x] `test-build` workflow (linux) produces an AppImage with `libwayland-client.so*` **absent** from `AppDir/usr/lib/` — verified locally via `unsquashfs`; sibling `libwayland-*` libs still present (not targeted).
- [ ] Post the artifact URL on #955 for @PACHAKUTlQ (Arch/Wayland) to confirm the app launches without `EGL_BAD_PARAMETER`.
- [ ] Optional local reproduction: boot into a Wayland session in UTM Debian 12 Rosetta or Fedora 38 and run the AppImage.

PR #1082, opened 2026-07-13
ci(THU-681): pin + verify linuxdeploy via shared composite action

raivieiraadriano92:
## Summary

- Follow-up to [PR #1076](https://github.com/thunderbird/thunderbolt/pull/1076) review — addresses two bot comments flagging (1) supply-chain risk from pulling linuxdeploy from the rolling `continuous` tag without a checksum, and (2) duplication of that setup between `desktop-release.yml` and `test-build.yml`.
- Extracts the linuxdeploy pre-populate step + `LINUXDEPLOY_EXCLUDED_LIBRARIES` env var into a shared composite action at `.github/actions/setup-linuxdeploy/action.yml`. Both workflows now call `uses: ./.github/actions/setup-linuxdeploy`.
- The action pins to a specific `continuous` asset (2026-07-01, upstream commit `a9f929ff`), retries on transient failures, downloads to a temp file, verifies sha256 **before** placing in Tauri's cache, and exports the exclusion env var via `$GITHUB_ENV` so downstream `bun tauri build` steps inherit it.

Closes THU-681. Stacked on [PR #1076](https://github.com/thunderbird/thunderbolt/pull/1076) — will auto-retarget to `main` once that lands.

## Test plan

- [ ] Re-run the `test-build` (linux) workflow off this branch; artifact should still have `libwayland-client.so*` absent from the AppImage (identical bytes-worth of behavior to #1076).
- [ ] Deliberate-mismatch check: temporarily change `EXPECTED_SHA` to a wrong value, confirm the workflow fails with `sha256sum: WARNING: 1 computed checksum did NOT match`.

### DIFF
--- a/.github/actions/setup-linuxdeploy/action.yml
+++ b/.github/actions/setup-linuxdeploy/action.yml
@@ -1,0 +1,61 @@
+name: Set up pinned linuxdeploy
+description: >
+  Download a checksum-verified linuxdeploy release into Tauri's cache dir so
+  `bun tauri build` uses it instead of Tauri's outdated binary-releases pin
+  (which lacks LINUXDEPLOY_EXCLUDED_LIBRARIES env-var support). Also exports
+  the exclusion env var so downstream `bun tauri build` steps in the same job
+  drop libwayland-client from the AppImage bundle. Fixes EGL_BAD_PARAMETER on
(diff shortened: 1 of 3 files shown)


raivieiraadriano92:
Closing — folding into PR #1076 instead.

PR #1030, opened 2026-06-27
feat(settings): add tooltips to models page actions

Sdaswani:
## Summary
- Add hover tooltips on disabled edit/remove buttons for built-in models explaining they can't be edited or removed
- Add "Add model" tooltip on the header + button (icon-only control)
- Wrap disabled action buttons so tooltips still receive pointer events

## Test plan
- [x] `bun test src/settings/models/index.test.tsx`
- [x] Open Settings → Models and hover the + button — shows "Add model"
- [x] Hover disabled pen/trash on built-in models (Opus 4.8, Tinfoil models) — shows built-in restriction messages
- [x] Hover pen/trash on a user-added model — shows "Edit model" / "Remove model"

Made with [Cursor](https://cursor.com)

PR #1088, opened 2026-07-15
feat(settings): add tooltips to Agents, MCP, and Skills add buttons

Sdaswani:
## Summary
- Follow-up to #1030: add hover tooltips on the icon-only **+** buttons on **Agents**, **MCP Servers**, and **Skills**
- Match the Models page pattern (`aria-label` + tooltip); wrap the Agents button so the tooltip still works when disabled
- Labeled empty-state buttons (e.g. “Add Server”) are left alone — text already conveys the action

## Test plan
- [x] Settings → Agents — hover header **+** → “Add custom agent”
- [x] Settings → MCP Servers — hover header **+** → “Add MCP server”
- [x] Settings → Skills — hover header **+** → “Create skill”
- [x] Empty-state labeled buttons still work without needing a tooltip

Made with [Cursor](https://cursor.com)


ital0:
Hello @Sdaswani :D 

Sorry for the wait on this.

I'm going to close this without merging. The change lands in `PageCreateAction`, so it would affect every page that uses the header "+" (Agents, Skills, Connections, Models, Tasks, Projects), not just the three in the PR title. More importantly, tooltips on those buttons aren't consistent with how similar actions work elsewhere in the app. UI changes like this need to go through our design review first so we don't end up with one pattern in settings and another everywhere else.

Thanks for putting this together, and for following up on #1030. If we want tooltips on these actions, we'll tackle it as part of a broader UI pass.

Appreciate the contribution.

Issue #956, opened 2026-06-05
Desktop appimage fails to initialize


PACHAKUTlQ:
Arch Linux, Wayland. thunderbolt desktop appimage version v0.1.101

The downloaded AppImage fails to initialize. (the appimage fails with EGL_BAD_PARAMETER error, but setting system wayland lib temporarily fixes this, see https://github.com/thunderbird/thunderbolt/issues/955)

<img width="1259" height="662" alt="Image" src="https://github.com/user-attachments/assets/3dc40f9b-757c-4a59-9b24-ccb4665efd92" />

same for extracting appimage and running the AppRun.

PACHAKUTlQ:
The issues persist in latest version. Any updates?

chymian:
also with deb as of today  Thunderbolt_0.1.105_amd64.deb

Valinwolf:
I can confirm Thunderbolt-0.1.105-1.x86_64.rpm holds the same issue.

mrdev023:
I can confirm too with Thunderbolt_0.1.106_amd64.AppImage

raivieiraadriano92:
Thanks for reporting! This is a duplicate of https://github.com/thunderbird/thunderbolt/issues/955 (same underlying `EGL_BAD_PARAMETER`). Closing this one — follow #955 for the fix.

PACHAKUTlQ:
Hi! @raivieiraadriano92 thx for the fix for #955, but this is not a duplicate of #955. #955 cannot create window, while this failure of initialization is within the app. The initialization still fails with v0.1.108.

This is why I created two issues in the first place.

Please have a look at the screenshot above.

raivieiraadriano92:
Hi @PACHAKUTlQ — sorry, you're right, this isn't a dup of #955. Reopening!

To dig in, could you share:

1. Terminal output: ./Thunderbolt_0.1.108_amd64.AppImage 2>&1 | tee thunderbolt.log
2. DevTools Console tab — hit Ctrl+Shift+I in the app window and screenshot it
3. Your compositor (sway / hyprland / gnome-shell / etc.)

Thanks

PACHAKUTlQ:
There is nothing in the terminal and only one warning in the console tab. btw all the requests succeed in the network tab.

<img width="1369" height="749" alt="Image" src="https://github.com/user-attachments/assets/25b7d8a4-85f0-4fad-a718-6fc666edb712" />

btw there is this in terminal when opening devtool:
```
[...]

PR #1089, opened 2026-07-16
fix(linux): resolve WebKitGTK OOM and OPFS-unavailable sync failure causing crash on Fedora 44 desktop app

chryzmo:
- Fall back to IDBBatchAtomicVFS when OPFS is unavailable, since WebKitGTK (Tauri on Linux) doesn't implement navigator.storage.getDirectory and OPFSCoopSyncVFS's worker would otherwise throw on init.
- Disable WebKitGTK's WASM OMG JIT tier, which leaks native memory when repeatedly recompiling wa-sqlite's module and eventually OOM-kills the webview process (upstream bug webkit.org/show_bug.cgi?id=319572).

Note that the WebKit flag appears to be required due to a bug in WebKit that causes constant spiked CPU and a moderately quick memory leak.  I have confirmed this WebKit bug exists in the version of WebKit provided with Fedora 44, but have not tested on other distros or versions of Fedora

It looks like this could be the same crash that https://github.com/thunderbird/thunderbolt/issues/956 is experiencing, so perhaps this could help there as well.

I have also filed a bug with WebKit that includes an example script to replicate the CPU spike and memory leak outside without Thunderbolt context:  https://bugs.webkit.org/show_bug.cgi?id=319572


raivieiraadriano92 (changes requested):
hey @chryzmo thanks for this bugfix.

Could you handle with the conflict in the file `lib.rs` ? I'll be happy to merge this PR after that.

chryzmo:
@raivieiraadriano92 Thanks for the review!  I've accepted both sets of changes and it looks like tests are passing again.  Let me know if I can provide anything else.

raivieiraadriano92 (approved):
Amazing - thanks for the bugfix @chryzmo

PR #1148, opened 2026-07-28
feat(THU-683): acoustic barge-in for voice mode

darkbanjo:
## What

Adds **acoustic barge-in** to voice mode: start talking and the assistant stops. Previously voice was half-duplex (the mic was gated off while the assistant thought/spoke), so long rambling replies couldn't be interrupted.

## How

- **`vad.ts`** — keeps energy VAD running, but fires `onSpeechStart` only after `bargeInFrames` (~130 ms) of **sustained** speech, not on the first frame. That's the echo-robustness guard: a brief residual-echo blip from the assistant's own playback can't self-trigger. Kept below `minSpeechFrames` so barge-in leads the utterance commit.
- **`session.ts`** — the mic/VAD now runs during `thinking`/`speaking` (only `idle` silences it). On `onSpeechStart` it aborts the turn (stops STT/LLM/TTS), flushes the playback queue (<100 ms), and drops back to `listening`. The VAD keeps the same utterance buffered, so it commits as the next turn — no lost onset.

## Reliability caveat ⚠️

Running the mic during playback reintroduces the echo problem we deliberately avoided when going half-duplex. This leans entirely on the browser's echo cancellation (already enabled in `getUserMedia`, with playback routed through the AEC reference) to stop the assistant's own audio from self-triggering a barge-in. That's **unreliable without headphones**, especially cross-browser/device.

The tuning knobs if it interrupts itself:
- `bargeInFrames` (raise → needs longer sustained speech)
- `speechRmsThreshold` (raise → needs louder speech)

**Needs real-device testing to dial in** — I couldn't validate audio behavior in CI. If it proves flaky in the wild, options are: raise the thresholds, gate barge-in behind a setting, or fall back to a manual tap-to-interrupt.

## Tests

- `session.test.ts`: barge-in during `speaking` aborts the turn (no assistant transcript committed) and returns to `listening`; `onSpeechStart` is a no-op when idle/listening. Full suite green (53 tests), typecheck clean.


darkbanjo:
Follow-up on the latest `thunder-deep-review` notes:

- **🔧 Nit (stale `bargeInFrames` ref):** fixed — the session doc now points at `minSpeechFrames` in the endpointer, which is the real knob (the endpointer already named it correctly).
- **📐 Convention (global voice flag blast radius):** intentionally keeping the tab-global flag for now — the only realistic overlap is title-generation during a live voice session (composer is inert during voice, so no interactive overlap), and the impact is cosmetic. Documented the trade-off and the per-request escalation path (`sendMessage(..., { body: { voiceMode: true } })` → prompt builder) in `voice-mode.ts` so it's a deliberate decision, not an oversight. Happy to scope it per-request if it ever actually bites.

PR #1150, opened 2026-07-28
fix(makefile): lowercase path segments in COMPOSE_PROJECT_NAME

vdbrown2:
## Summary
- `make up` fails with `invalid project name "...": must consist only of lowercase alphanumeric characters, hyphens, and underscores...` whenever the repo's parent directory contains an uppercase letter (e.g. `~/Projects/thunderbolt`).
- The auto-generated `COMPOSE_PROJECT_NAME` (`<parent>-<repo>`) already sanitized special characters via `sed`, but never lowercased the segments, so Docker Compose rejected the result.
- Adds `tr '[:upper:]' '[:lower:]'` before sanitizing each path segment so the generated project name is always valid.

## Test plan
- [x] Reproduced the original failure with a capitalized parent directory and confirmed `make up` errored with the exact message from the issue.
- [x] Applied the fix and confirmed the derived name (`fake-parent-dir-thunderbolt-test`) is fully lowercase for a mixed-case path.
- [x] Ran `make up` / `make down` in the real repo (parent dir `~/Projects`) with no `COMPOSE_PROJECT_NAME` override — containers started successfully as `projects-thunderbolt-*`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ COMPOSE ?= $(shell command -v podman-compose > /dev/null 2>&1 && podman info > /
 # override with `COMPOSE_PROJECT_NAME=foo make up` if you want a fixed name.
 # Sanitize each segment so paths with spaces or other special characters (e.g. "~/My Projects/thunderbolt")
 # produce a valid Docker Compose project name.
-COMPOSE_PROJECT_NAME ?= $(shell basename "$$(cd .. && pwd)" | sed 's/[^a-zA-Z0-9._-]/-/g')-$(shell basename "$$(pwd)" | sed 's/[^a-zA-Z0-9._-]/-/g')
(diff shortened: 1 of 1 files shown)


vdbrown2:
Pushed a follow-up commit adding a code comment for the \`sed\` character class change, in case it stands out in review: it was narrowed from \`[^a-zA-Z0-9._-]\` to \`[^a-z0-9._-]\` since the new \`tr\` step already lowercases both segments before \`sed\` runs. Keeping \`A-Z\` in the allowed set would've been dead code, and dropping it makes \`sed\` assert "this is already lowercase" rather than just stripping stray characters — so a future reordering of the pipe that reintroduces uppercase would get caught instead of silently passing through.

ital0:
hey @vdbrown2 :]

Thanks for this, and sorry for the slow review.

The uppercase bug is real. I reproduced the Docker Compose rejection on names like `Projects-thunderbolt`, and the `tr` + narrowed `sed` fix looks correct. Paths that are already lowercase keep the same project name, so no volume migration concern there.

I'm going to merge as-is.

Nice catch.

ital0 (approved):

PR #1152, opened 2026-07-29
feat(models): Ollama capabilities catalog + chat Thinking chip

Sdaswani:
## Summary

- When adding a **Custom** provider whose URL looks like Ollama (port `11434` or hostname contains `ollama`), prefer native `GET /api/tags` so tools / thinking / vision / context window prefills are richer than sparse `/v1/models`.
- Soft-fail the probe (network error, non-Ollama JSON, empty catalog) and fall back to `/v1/models` so non-Ollama custom endpoints keep working.
- Persist advertised capabilities into the model row (`toolUsage`, `startWithReasoning`, `contextWindow`) on **create and edit**, show them on model detail, and add a per-conversation **Thinking** chip in the composer for models with `startWithReasoning === 1`.
- Chip off does not mutate the stored model (so the chip stays visible); it overrides the send path (`startWithReasoning` view + `reasoningEffort: 'none'` for OpenAI-compat / Ollama, and Pi `thinkingLevel: 'off'` in the built-in adapter).

Durable contract is the model row + session override — not “we are permanently bound to Ollama.” Autodetection is opportunistic.

## Heuristic / failover notes

- Probe gate: `isLikelyOllamaBaseUrl` — intentional incomplete match so llama.cpp / corporate OpenAI-compat URLs are not hit with an extra failed `/api/tags`.
- Ollama on an uncommon port without `ollama` in the host still works via `/v1/models` (no rich caps).
- `/api/tags` body is zod-validated (`ollamaTagsResponseSchema` envelope + per-tag `ollamaTagModelSchema`); malformed tags are skipped; empty/invalid → soft-fail.

## Test plan

- [x] Unit: `ollama-catalog` origin/heuristic/mapping/schema
- [x] Unit: `fetchModelsForProvider` prefers `/api/tags`; fails over on throw and non-Ollama body
- [x] Unit: add-model form persists Ollama-advertised flags
- [x] Unit: `thinking-session` helpers + `openaiCompatThinkingProviderOptions`
- [x] Unit: `withThinkingSessionOverride` (ACP / Pi path)
- [x] Unit: `ThinkingChip` a11y pressed/toggle
- [ ] Manual: Custom URL `http://localhost:11434/v1` → catalog shows tools/thinking/context; create model; chip appears in chat
- [ ] Manual: toggle Thinking off → send does not stream reasoning (Ollama thinking model)
- [ ] Manual: Custom URL non-Ollama (e.g. llama.cpp other port) → no `/api/tags` probe failure in UX; `/v1/models` path works
- [ ] Manual: Edit model → change Thinking / Tools / Context window → detail + chip reflect updates
- [ ] Manual: switch to a non-thinking model → chip hidden; prior chat’s `thinkingEnabled` does not leak (reset on model switch)

## Reviewer notes

Addressed prior review:

1. Dropped ALL_CAPS `REASONING_OFF` (inlined `0`)
2. Removed dead `contextLengthFromModelInfo`; wired `ollamaTagsResponseSchema` into the fetch path (envelope stays `unknown[]` for per-tag skip)
3. Removed `as never` casts on `stubJsonResponse` mocks
4. Spread `ResolvedPiModel` consistently in both `withThinkingSessionOverride` branches

Made with [Cursor](https://cursor.com)


src/settings/models/ollama-catalog.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/** Maps Ollama capability strings + detail fields onto Thunderbolt catalog fields. */
+export const mapOllamaTagToAvailableModel = (tag: OllamaTagModel): AvailableModel => {
+  const capabilities = new Set((tag.capabilities ?? []).map((capability) => capability.toLowerCase()))
+  const contextWindow = tag.details?.context_length
+  return {
+    id: tag.name,
+    name: tag.name,
+    supports_tools: capabilities.has('tools'),
+    supports_thinking: capabilities.has('thinking'),
```
ital0: Current `GET /api/tags` rows do not contain `capabilities` or `details.context_length`; Ollama exposes those through `POST /api/show`. With a real tags response, this maps every model to tools, thinking, and vision `false`, then the add form persists `toolUsage: 0`. Please enrich the list from `/api/show`, or keep missing fields as unknown instead of converting them to false.

src/settings/models/model-catalog.ts
```diff
@@ -119,6 +126,15 @@ export const fetchModelsForProvider = async ({ provider, apiKey, url }: CatalogR
       .sort((left, right) => left.id.localeCompare(right.id))
   }
 
+  // Local Ollama (default Custom URL) exposes tools / thinking / vision /
+  // context_length on `/api/tags`. Prefer that over sparse `/v1/models`.
+  if (provider === 'custom') {
+    const ollamaCatalog = await tryFetchOllamaCatalog(url)
```
ital0: The new form consumers distinguish false from unknown, but the generic mapping later in this function always writes a boolean for `supports_tools`. When OpenAI or fallback `/v1/models` omits that field, this function returns false instead of undefined and add or edit disables tools. Please preserve all three states and add a sparse OpenAI response test.

src/settings/models/use-add-model-form.ts
```diff
@@ -180,18 +197,35 @@ export const useAddModelForm = ({ isOpen, onClose, onMutationStart }: UseAddMode
...
   }
 
+  const applyCatalogCapabilities = (modelId: string) => {
+    const selected = catalog.models.find((candidate) => candidate.id === modelId)
+    form.setValue('contextWindow', selected?.context_window ?? defaultCapabilityValues.contextWindow, {
+      shouldValidate: false,
+    })
+    form.setValue(
+      'toolUsage',
+      selected?.supports_tools === undefined ? defaultCapabilityValues.toolUsage : selected.supports_tools,
```
ital0: `fetchModelsForProvider` always turns a missing `supports_tools` from OpenAI or generic `/v1/models` into false, so the `undefined` fallback here never runs. Selecting a normal OpenAI row such as `{ id: 'gpt-5' }` now saves `toolUsage: 0`, while this flow previously saved `1`. Please preserve unknown catalog values and default them to tools on.

src/settings/models/use-edit-model-form-state.ts
```diff
@@ -101,13 +115,30 @@ export const useEditModelFormState = (model: Model) => {
     void fetchCatalog(catalogRequest)
   }
 
+  const applyCatalogCapabilities = (modelId: string) => {
+    const selected = catalog.models.find((candidate) => candidate.id === modelId)
+    if (!selected) {
+      return
+    }
+    form.setValue('contextWindow', selected.context_window ?? null, { shouldDirty: true, shouldValidate: false })
+    form.setValue('toolUsage', selected.supports_tools === undefined ? true : selected.supports_tools, {
```
ital0: This also overwrites existing settings when catalog metadata is sparse. Reselecting an OpenAI or custom model can turn tools off and clear `startWithReasoning` even though the provider never reported either capability. Please leave each current form value unchanged when that catalog field is unknown.

src/ai/fetch.ts
```diff
@@ -737,9 +748,18 @@ export const aiFetchStreamingResponse = async ({
...
+        // option means “treat all content as reasoning until </think>”.
+        // Ollama thinking models emit a native `reasoning` field and plain
+        // answer `content` with no closing tag; wiring the chip flag through
+        // here classifies the answer as reasoning forever and never emits
+        // reasoning-end (stuck spinner with the reply stuck in the muted
+        // Thinking pane).
         extractReasoningMiddleware({
           tagName: 'think',
-          startWithReasoning: Boolean(model.startWithReasoning),
+          startWithReasoning: false,
```
ital0: `startWithReasoning` was originally added as the `extractReasoningMiddleware` parser option for models that begin with untagged reasoning. Hardcoding this to false stops honoring existing rows, while the same synced column is now written as a capability and older clients will interpret that new value as parser behavior. Please use a separate capability field and roll that synced model column out with the documented two-PR process.

src/ai/fetch.ts
```diff
@@ -761,6 +781,8 @@ export const aiFetchStreamingResponse = async ({
       // rather than relying solely on the profile — custom OpenAI models may not have a profile.
       ...(model.vendor === 'openai' && { systemMessageMode: 'developer' as const }),
       ...profile?.providerOptions,
+      // Thinking chip off → ask OpenAI-compat endpoints (incl. Ollama) for no reasoning.
+      ...openaiCompatThinkingProviderOptions(thinkingDisabled),
```
ital0: `reasoningEffort: 'none'` is not a universal disable signal. This branch also adds it for Anthropic, and OpenAI o3, o4-mini, and several GPT-5 families reject `none`, so toggling the visible chip can do nothing or fail with a 400. Please define supported off behavior per provider and model, then test the actual payload.

src/acp/built-in-adapter.ts
```diff
@@ -468,11 +491,12 @@ const fetchViaHarness = async (
...
     telemetry: context.telemetry,
   })
-  const resolved = resolvePiModel(agentCore, context, config.profile)
-  if (!resolved) {
+  const resolvedBase = resolvePiModel(agentCore, context, config.profile)
+  if (!resolvedBase) {
     context.telemetry?.setDimensions({ engine: 'legacy' })
     return fallback()
   }
+  const resolved = withThinkingSessionOverride(resolvedBase, context.selectedModel, context.thinkingEnabled)
```
ital0: For user-added OpenAI and custom models, `config.profile` is normally null, so `resolvedBase.descriptor.reasoning` is already false. Pi then clamps the fallback thinking level off, which makes chip on and chip off send the same request, while legacy sends `reasoning_effort: none` only for off.

src/components/chat/chat-prompt-input.tsx
```diff
@@ -188,8 +189,15 @@ export const ChatPromptInput = forwardRef<ChatPromptInputRef, ChatPromptInputPro
       id: chatThreadId,
       selectedAgent,
       selectedModel,
+      thinkingEnabled,
     } = useCurrentChatSession()
+    const updateSession = useChatStore((state) => state.updateSession)
 
+    const thinkingOn = thinkingEnabled !== false
+    const showThinkingChip = selectedModel.startWithReasoning === 1
```
ital0: The model picker already hides itself unless `selectedAgent.type === 'built-in'`, but this new chip does not have that gate. A remote or managed ACP chat can therefore show a control that its adapter ignores. Please gate the chip on the built-in agent and cover the ACP case.

src/chats/chat-store.ts
```diff
@@ -51,6 +51,11 @@ export type ChatSession = {
   retriesExhausted: boolean
   selectedAgent: Agent
   selectedModel: Model
+  /**
+   * Per-conversation override for models that advertise thinking
+   * (`startWithReasoning === 1`). `undefined` means “use the model default” (on).
+   */
+  thinkingEnabled?: boolean
```
ital0: This is described as a conversation setting, but it only lives in the in-memory session and hydration never restores it. Reloading a chat silently turns Thinking back on. Was that intentional, or should this be persisted with the chat?

src/ai/thinking-session.ts
```diff
@@ -0,0 +1,37 @@
...
+ * the composer chip remains visible).
+ */
+export const withThinkingDisabledForSend = <T extends Pick<Model, 'startWithReasoning'>>(
+  model: T,
+  thinkingEnabled: boolean | undefined,
+): T => {
+  if (!isThinkingDisabledForSend(model, thinkingEnabled)) {
+    return model
+  }
+  return { ...model, startWithReasoning: 0 }
```
ital0: After `fetch.ts` hardcodes the middleware option to false, no legacy code reads this changed field from the copied model. The actual request change uses the separate `thinkingDisabled` boolean, so this helper and its test are dead code. Please remove it and keep one source for send behavior.

src/settings/models/model-catalog.test.ts
```diff
@@ -57,3 +65,82 @@ describe('model catalog policy', () => {
...
+describe('fetchModelsForProvider Ollama path', () => {
+  it('prefers /api/tags capabilities when the Custom URL looks like Ollama', async () => {
+    const getSpy = spyOn(http, 'get').mockImplementation((url: string) => {
+      if (String(url).includes('/api/tags')) {
+        return stubJsonResponse({
+          models: [
+            {
+              name: 'qwen2.5:14b',
+              details: { context_length: 32_768 },
+              capabilities: ['completion', 'tools'],
```
ital0: This fixture includes fields the real `/api/tags` contract does not return, so the test passes while production persists false capabilities. Please use a documented tags row and cover whatever `/api/show` enrichment supplies the capabilities and context. That would reproduce the main failure instead of hiding it.

src/settings/models/ollama-catalog.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  })
+})
+
+describe('mapOllamaTagToAvailableModel', () => {
+  it('maps tools + context from a Qwen 2.5 style tag', () => {
+    expect(
+      mapOllamaTagToAvailableModel({
+        name: 'qwen2.5:14b',
+        details: { context_length: 32_768, family: 'qwen2' },
+        capabilities: ['completion', 'tools'],
```
ital0: Please add a mapping case with a real tags row that has no `capabilities` and no `context_length`. The expected result should preserve unknown values, not assert that missing data means the model cannot use tools or thinking. This is the edge case users hit by default.

src/settings/models/use-add-model-form.test.tsx
```diff
@@ -133,6 +142,68 @@ describe('useAddModelForm', () => {
...
+          contextWindow: 40_960,
+          toolUsage: true,
+          startWithReasoning: true,
+        })
+        await getClock().runAllAsync()
+      })
+
+      const created = useChatStore.getState().models.find((candidate) => candidate.model === 'qwen3:1.7b')
+      expect(created?.contextWindow).toBe(40_960)
+      expect(created?.toolUsage).toBe(1)
```
ital0: This create test only covers an ideal catalog row with explicit true capabilities. Please add the OpenAI flow with `/v1/models` returning `{ id: 'gpt-5' }`, select it, submit it, and assert `toolUsage` stays `1`. That test would catch the current regression.

src/settings/models/model-catalog.test.ts
```diff
@@ -57,3 +65,82 @@ describe('model catalog policy', () => {
...
+      expect(getSpy.mock.calls.some((call) => String(call[0]).endsWith('/api/tags'))).toBe(true)
+    } finally {
+      getSpy.mockRestore()
+    }
+  })
+
+  it('falls back to /v1/models when /api/tags fails for an Ollama-like URL', async () => {
+    const getSpy = spyOn(http, 'get').mockImplementation((url: string) => {
+      if (String(url).includes('/api/tags')) {
+        throw new Error('connection refused')
```
ital0: This intentionally triggers `console.error` in the fallback path, but the test leaves it unsuppressed. `docs/development/testing.md` asks tests that expect an error to install a `console.error` spy, so this will add noise to the suite. Please suppress and restore it for these cases.

src/settings/models/ollama-catalog.ts
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Fetches the native Ollama model list.
+ *
+ * Returns `null` when the probe should soft-fail (network error, non-2xx via
+ * http client, or body that is not Ollama-shaped) so callers fall back to
+ * `/v1/models`. Soft-fail is intentional for an optional capability probe —
+ * loud throw would break Custom-provider add flows for every non-Ollama URL
+ * that still matched the heuristic.
+ */
+export const fetchOllamaTagsCatalog = async (origin: string): Promise<AvailableModel[] | null> => {
```
ital0: The new catalog function imports the shared HTTP singleton, which forces these tests to replace `http.get` globally. Both testing guides prefer an injected client or fetch seam to avoid cross-test pollution. Could this function accept an injected dependency with the production client as its default?

ital0:
@Sdaswani, thanks for the time you put into this, and for working through the earlier review rounds.

After analyzing your PR, I'm not going to move forward with it. 

Our focus right now is on managed and enterprise deployments, and capability discovery that only triggers for Ollama URLs sits outside that. The Thinking chip and the new form fields are generic in principle, but in practice Ollama is the only provider that feeds them, so most of what this PR adds would go unused by the people we're building for.

There's a second, practical reason. Even if the scope were right for us, the change writes to a synced model column and reshapes the shared catalog mapping, and doing that safely for every provider would take more design and review than we can commit this cycle. That's a roadmap constraint on our side, not a judgment on the quality of your work.

Thanks again for the contribution.

PR #1186, opened 2026-08-04
refactor(THU-739): drop legacy modes table

arienemaiara:
## What & why

THU-739, PR 2 of 2. The backend teardown for the removed chat-modes feature, now that PR #1181 stopped syncing it. Drops the `modes` table and the dead `chat_threads.mode_id` column (a FK-like reference to it; nothing reads or writes it). PR 1 kept the backend defs so the postgres objects survived until the PowerSync service rolled off the `modes` sync rule; this PR removes them.

## Changes

- **`backend/src/db/powersync-schema.ts`** — removed the `modesTable` definition and the `chat_threads.modeId` column.
- **`backend/drizzle/0026_overconfident_silvermane.sql`** — one migration for both drops:
  ```sql
  DROP TABLE "powersync"."modes" CASCADE;
  ALTER TABLE "powersync"."chat_threads" DROP COLUMN "mode_id";
  ```
- **`_journal.json` / snapshot** — migration metadata (idx 26).

## ⚠️ Merge gate

Run this migration only **after** the Render `powersync` service has been rolled onto the image from PR #1181 (sync rules no longer reference `powersync.modes`). Dropping the table while the service still replicates it breaks replication.

## Testing

- `bun db generate` → "No schema changes" (snapshot consistent, no drift).
- Backend type-check clean (no dangling `modesTable` / `modeId` refs).
- `powersync.test.ts` → 25 pass (test DB applied the migration; the legacy accept-and-ignore case still passes).

### DIFF
--- a/backend/drizzle/0026_overconfident_silvermane.sql
+++ b/backend/drizzle/0026_overconfident_silvermane.sql
@@ -1,0 +1,6 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+DROP TABLE "powersync"."modes" CASCADE;--> statement-breakpoint
+ALTER TABLE "powersync"."chat_threads" DROP COLUMN "mode_id";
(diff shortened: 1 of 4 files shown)


backend/drizzle/0026_overconfident_silvermane.sql
```diff
@@ -0,0 +1,5 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
raivieiraadriano92, line 1: could we merge these two migration files into a single one?

PR #1201, opened 2026-08-06
test: stop depending on another test file's leaked indexedDB global

njbrake:

Removes a cross-file dependency that makes `sidebar-footer.test.tsx` fail depending on test order.

Its "sync retry flow" block enables sync, which puts the encryption config on the render path. That reads the bare `indexedDB` global, which happy-dom does not implement, so the tests only passed when some earlier test *file* had already stubbed it. With `--randomize`, an ordering that runs this file first fails with `ReferenceError: indexedDB is not defined`.

`use-app-initialization.test.tsx` was the accidental supplier, and its restore was itself wrong: it captured `globalThis.indexedDB` at module load (undefined, since happy-dom has none) and wrote that back in `afterAll`, leaving the property defined-but-undefined rather than absent.

Adds `stubIndexedDb`/`restoreIndexedDb`, which delete the global when it did not exist before, and has both files opt in explicitly. The absence is load-bearing elsewhere, since the boot pipeline's storage pre-flight is supposed to report STORAGE_UNAVAILABLE without it, so this stays opt-in rather than going into the global preload.

I saw this fail on a randomized run of `main`. I could not pin a seed that reproduces it on `main` itself, but it reproduces deterministically once any single extra test file changes the shuffle.

### DIFF
--- a/src/components/sidebar-footer.test.tsx
+++ b/src/components/sidebar-footer.test.tsx
@@ -11,6 +11,7 @@ import { MemoryRouter } from 'react-router'
 
 import { setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
 import { createMockAuthClient } from '@/test-utils/auth-client'
+import { restoreIndexedDb, stubIndexedDb } from '@/test-utils/indexed-db'
 import { createTestProvider } from '@/test-utils/test-provider'
 import { forceMobileViewport, restoreViewport } from '@/test-utils/viewport'
 
@@ -225,12 +226,20 @@ describe('sync retry flow', () => {
   // The test database is bun-sqlite (no PowerSync instance), so usePowerSyncStatus
   // reports 'not-configured' — with sync enabled that is exactly the
   // "needs attention" state that surfaces the Retry button.
+  //
+  // Enabling sync also puts the encryption config on the render path, which reads
+  // `indexedDB` — absent in happy-dom. These tests used to pass only when an
+  // earlier test FILE happened to leave a stub on the global, so a shuffle that
+  // put this file first failed with `ReferenceError: indexedDB is not defined`.
+  // Stubbing it here makes the block self-contained.
   beforeEach(() => {
+    stubIndexedDb()
     useLocalSettingsStore.getState().setLocalSetting('syncEnabled', true)
   })
 
   afterEach(() => {
     useLocalSettingsStore.getState().setLocalSetting('syncEnabled', false)
+    restoreIndexedDb()
   })
 
   const openAccountMenu = async () => {
--- a/src/components/sidebar-footer.test.tsx
+++ b/src/components/sidebar-footer.test.tsx
@@ -4,7 +4,7 @@
 
 import '@testing-library/jest-dom'
 import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react'
-import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from 'bun:test'
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, mock } from 'bun:test'
 import { Cloud, CloudAlert, CloudOff, Loader2 } from 'lucide-react'
 import { type ReactElement, type ReactNode } from 'react'
 import { MemoryRouter } from 'react-router'
@@ -217,6 +217,26 @@ describe('SyncStateIcon', () => {
   })
 })
 
+// `pending-device-modal.test.tsx` and `use-pending-device-notification.test.tsx`
+// register `mock.module('@/db/encryption', { isEncryptionEnabled: () => true })`.
+// Bun's module mocks are worker-global and permanent — they also rewrite the
+// binding `needsSyncSetupWizard` calls inside `./config`, so resetting the config
+// store is not enough to switch encryption back off. Re-provide both modules with
+// the real store-backed implementation, the same defense `db/encryption/config.test.ts`
+// uses. Without it, a shuffle that runs either of those files first sends the sync
+// retry tests into key storage and fails them on the missing `indexedDB` global.
+const realEncryptionConfig = await import('@/db/encryption/config')
+mock.module('@/db/encryption/config', () => ({
+  ...realEncryptionConfig,
+  isEncryptionEnabled: () => useConfigStore.getState().config.e2eeEnabled === true,
+}))
+
+const realEncryption = await import('@/db/encryption')
+mock.module('@/db/encryption', () => ({
+  ...realEncryption,
+  isEncryptionEnabled: () => useConfigStore.getState().config.e2eeEnabled === true,
+}))
+
 describe('sync retry flow', () => {
   const loggedInAuthClient = () =>
     createMockAuthClient({
@@ -229,10 +249,10 @@ describe('sync retry flow', () => {
   //
   // Enabling sync also puts the encryption config on the render path: with E2EE
   // on, `useSyncEnabledToggle` reads the Content Key out of IndexedDB — absent in
-  // happy-dom — and silently turns sync back off when there is none. The config
-  // store is a persisted module global that another test FILE can leave switched
-  // on, so pin it off here. That keeps sync enabled for the whole block and keeps
-  // the encryption path (and IndexedDB with it) off the render path entirely.
+  // happy-dom, where key storage reads the bare global and throws
+  // `ReferenceError: indexedDB is not defined`. Keeping E2EE off keeps that path
+  // (and IndexedDB with it) out of the render entirely, which is what the two
+  // guards above the block and the store reset below exist to guarantee.
   beforeEach(() => {
     useConfigStore.setState({ config: {} })
     useLocalSettingsStore.getState().setLocalSetting('syncEnabled', true)
--- a/src/components/sidebar-footer.test.tsx
+++ b/src/components/sidebar-footer.test.tsx
@@ -9,9 +9,9 @@ import { Cloud, CloudAlert, CloudOff, Loader2 } from 'lucide-react'
 import { type ReactElement, type ReactNode } from 'react'
 import { MemoryRouter } from 'react-router'
 
+import { useConfigStore } from '@/api/config-store'
 import { setupTestDatabase, teardownTestDatabase } from '@/dal/test-utils'
 import { createMockAuthClient } from '@/test-utils/auth-client'
-import { restoreIndexedDb, stubIndexedDb } from '@/test-utils/indexed-db'
 import { createTestProvider } from '@/test-utils/test-provider'
 import { forceMobileViewport, restoreViewport } from '@/test-utils/viewport'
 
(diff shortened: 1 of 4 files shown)


src/components/sidebar-footer.test.tsx
```diff
@@ -225,12 +226,20 @@ describe('sync retry flow', () => {
...
   // reports 'not-configured' — with sync enabled that is exactly the
   // "needs attention" state that surfaces the Retry button.
+  //
+  // Enabling sync also puts the encryption config on the render path, which reads
+  // `indexedDB` — absent in happy-dom. These tests used to pass only when an
+  // earlier test FILE happened to leave a stub on the global, so a shuffle that
+  // put this file first failed with `ReferenceError: indexedDB is not defined`.
+  // Stubbing it here makes the block self-contained.
   beforeEach(() => {
+    stubIndexedDb()
```
ital0: These tests do not control `e2eeEnabled`. In this path, the real `needsSyncSetupWizard()` only calls `getCK()` when encryption is enabled. This stub returns no key, so the hook disables sync after render. The assertions can therefore observe the retry UI before the async migration check finishes, instead of the stable state described by the test.

src/hooks/use-app-initialization.test.tsx
```diff
@@ -28,35 +29,18 @@ const mockPostHogConfig = {
...
   beforeAll(async () => {
-    stubWorkingIndexedDb()
+    stubIndexedDb()
     await setupTestDatabase()
   })
 
   afterAll(async () => {
     await teardownTestDatabase()
-    Object.defineProperty(globalThis, 'indexedDB', { value: realIndexedDb, configurable: true, writable: true })
+    restoreIndexedDb()
```
ital0, line 43: This cleanup fixes a real leak. The old code could replace a missing global with an `indexedDB` property whose value was `undefined`. However, no test asserts that this restore removes the property, so the same regression could return without any test failing.

src/test-utils/indexed-db.ts
```diff
@@ -0,0 +1,87 @@
...
+ * A stub left behind makes a "no IndexedDB" test pass for the wrong reason, and
+ * `defineProperty(..., { value: undefined })` is NOT a valid restore — it leaves
+ * the property defined-but-undefined, which turns a would-be `ReferenceError`
+ * into a `TypeError` and quietly changes what the code under test does.
+ */
+
+/** Present so `restoreIndexedDb` can tell "was absent" from "was something". */
+const absent = Symbol('absent')
+
+let previous: IDBFactory | typeof absent = absent
```
ital0: `previous` is shared by every call. If `stubIndexedDb()` is called twice before restoring, the second call overwrites the first snapshot, and two restores still leave a stub installed. The helper can therefore create the same global leak it is meant to prevent.

src/test-utils/indexed-db.ts
```diff
@@ -0,0 +1,87 @@
...
+  const factory = {
+    open: () => {
+      const request = {
+        onsuccess: null as (() => void) | null,
+        onerror: null as (() => void) | null,
+        onupgradeneeded: null as (() => void) | null,
+        onblocked: null as (() => void) | null,
+        result: {
+          close: () => {},
+          // Enough of a database for key-storage's read/write round trip to be
```
ital0: This says the fake supports a read/write round trip, but it does not store values, implement `delete()`, or fire `transaction.oncomplete`. I verified that `storeCK()` and `getKeyPair()` never settle, while `clearCK()` throws. The type assertion presents this incomplete object as a full `IDBFactory`, so future tests can hang or fail when they use the advertised operations.

ital0 (changes requested):
I can see one concrete benefit here: `use-app-initialization.test.tsx` now restores a missing `indexedDB` global correctly. The old cleanup left the property defined with an `undefined` value.

I could not reproduce the reported failure. I ran the affected tests alone and together, then ran the full randomized suite five times on the parent commit and five times on this PR. Every run completed with 4,312 passing tests and no failures. That does not prove the flake cannot happen, but this PR has no regression test that fails before the change and passes after it.

My main concern is the sidebar test. It reaches key storage only when the E2EE state sends it there, but the test does not control that state. The new IndexedDB stub reports that no key exists, which makes the hook disable sync after render. The test can therefore assert the retry UI before the async migration check finishes.

The shared helper also presents itself as an `IDBFactory` that supports reads and writes, but several key-storage operations do not work with it. It can also leave the stub installed when calls are nested.

As written, the PR fixes one cleanup bug, but it does not demonstrate the reported failure or add coverage. It also introduces a shared fake that can leave global state behind and can make key-storage calls hang or throw.

src/components/sidebar-footer.test.tsx
```diff
@@ -225,7 +226,15 @@ describe('sync retry flow', () => {
...
   // "needs attention" state that surfaces the Retry button.
+  //
+  // Enabling sync also puts the encryption config on the render path: with E2EE
+  // on, `useSyncEnabledToggle` reads the Content Key out of IndexedDB — absent in
+  // happy-dom — and silently turns sync back off when there is none. The config
+  // store is a persisted module global that another test FILE can leave switched
+  // on, so pin it off here. That keeps sync enabled for the whole block and keeps
+  // the encryption path (and IndexedDB with it) off the render path entirely.
   beforeEach(() => {
+    useConfigStore.setState({ config: {} })
```
ital0, line 257: This reset does not isolate the test from leaked module mocks. `pending-device-modal.test.tsx` registers a global `mock.module('@/db/encryption')`. When that file runs first, this reset does not remove the mock, and the two sync retry tests fail with `ReferenceError: indexedDB is not defined`. The test still depends on file order.

src/test-utils/indexed-db.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { beforeEach, describe, expect, it } from 'bun:test'
+
+import { restoreIndexedDb, stubIndexedDb } from './indexed-db'
+
+const hasIndexedDb = () => 'indexedDB' in globalThis
+
+describe('stubIndexedDb / restoreIndexedDb', () => {
+  // happy-dom ships no IndexedDB, so "absent" is the real baseline — assert
+  // against it explicitly rather than trusting whatever ran before this file.
+  beforeEach(() => {
```
ital0: This deletes the global before each test, but cleanup still happens inside the individual test bodies. If an assertion fails before `restoreIndexedDb()`, both the global and the helper's private snapshot stay dirty and can affect later randomized tests.

src/test-utils/indexed-db.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    Reflect.deleteProperty(globalThis, 'indexedDB')
+  })
+
+  it('installs an openable factory and removes it again', async () => {
```
ital0, line 30: This file never exercises a key-storage read or write. The old fake that hung on transaction events would still pass all four tests here, so the new claim that unsupported operations fail instead of hanging is not covered.

src/test-utils/indexed-db.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    stubIndexedDb()
+    const opened = await new Promise<boolean>((resolve) => {
+      const request = globalThis.indexedDB.open('probe')
+      request.onsuccess = () => resolve(true)
+      request.onerror = () => resolve(false)
+    })
+    expect(opened).toBe(true)
+    restoreIndexedDb()
+    // Deleted, not set to undefined: code reading the bare global must still
+    // see a ReferenceError, which is what the boot pre-flight relies on.
```
ital0: This comment is not accurate. The boot preflight reads `globalThis.indexedDB` and returns `false` both when the property is absent and when its value is `undefined`. It does not rely on a `ReferenceError`. That error comes from key storage reading the bare global.

src/test-utils/indexed-db.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(hasIndexedDb()).toBe(false)
+  })
+
+  it('restores a pre-existing global rather than deleting it', () => {
+    const original = { marker: 'real' } as unknown as IDBFactory
```
ital0: This double assertion bypasses the type system and conflicts with the repository rule against `as unknown as`. It also treats a marker object as a complete `IDBFactory` even though it does not implement that contract.

src/test-utils/indexed-db.ts
```diff
@@ -0,0 +1,74 @@
...
+let previous: IDBFactory | typeof absent | typeof unstubbed = unstubbed
+
+/**
+ * Installs a minimal `indexedDB` whose `open()` always succeeds on a microtask.
+ * That covers availability probes (`isIndexedDbAvailable`) and nothing more —
+ * the returned database has no `transaction`, so key-storage reads and writes
+ * fail loudly rather than hanging on an event this fake never fires. Extend it
+ * here, with a test, if a case genuinely needs a backing store.
+ *
+ * Nested calls keep the first snapshot, so a stub can never restore over a stub.
```
ital0: These calls are not safe as nested pairs. After an outer and inner `stubIndexedDb()`, the inner `restoreIndexedDb()` removes the outer caller's stub. The test named `nested` also calls restore only once, so it verifies repeated setup rather than nested ownership.

ital0 (changes requested):
Hey! Not sure if you're done with the fixes... But I took another look and the follow-up fixes several issues from the first review.

The main order dependency is still reproducible. Running `pending-device-modal.test.tsx` before `sidebar-footer.test.tsx` leaves a global `@/db/encryption` module mock in place. Resetting the Zustand config does not remove that mock. In that order, two sidebar retry tests fail with `ReferenceError: indexedDB is not defined`.

The full randomized suite passed with 4,316 tests, and the three directly affected files passed with 39 tests. The specific two-file order above still failed with 25 passing tests and 2 failures. This means the general run can miss the exact ordering problem the PR is meant to remove.

The remaining test-isolation and documentation problems are noted inline.

njbrake:
@ital0 thanks for the review! I'll take another look, thanks for your patience 🙏 !

src/components/sidebar-footer.test.tsx
```diff
@@ -216,6 +217,26 @@ describe('SyncStateIcon', () => {
...
+mock.module('@/db/encryption/config', () => ({
+  ...realEncryptionConfig,
+  isEncryptionEnabled: () => useConfigStore.getState().config.e2eeEnabled === true,
+}))
+
+const realEncryption = await import('@/db/encryption')
+mock.module('@/db/encryption', () => ({
+  ...realEncryption,
+  isEncryptionEnabled: () => useConfigStore.getState().config.e2eeEnabled === true,
+}))
```
ital0, line 238: I may be missing a Bun detail here, but `realEncryption` can already be the namespace registered by an earlier `mock.module()`. `use-sync-enabled-toggle.test.ts` replaces `needsSyncSetupWizard`, and this spread carries that replacement forward because this guard only overwrites `isEncryptionEnabled`. I reproduced that order with E2EE off: `needsSyncSetupWizard()` still returned `true`, the hook rendered with sync enabled, and the migration effect then disabled it. The sidebar assertions pass before that effect settles, so this still leaves an order-dependent false positive.

These two registrations are also worker-global and permanent. That is the same behavior called out in `docs/development/testing.md` under "Avoid mock.module() for Shared Modules."

src/components/sidebar-footer.test.tsx
```diff
@@ -216,6 +217,26 @@ describe('SyncStateIcon', () => {
...
 
+// `pending-device-modal.test.tsx` and `use-pending-device-notification.test.tsx`
+// register `mock.module('@/db/encryption', { isEncryptionEnabled: () => true })`.
+// Bun's module mocks are worker-global and permanent — they also rewrite the
+// binding `needsSyncSetupWizard` calls inside `./config`, so resetting the config
+// store is not enough to switch encryption back off. Re-provide both modules with
+// the real store-backed implementation, the same defense `db/encryption/config.test.ts`
+// uses. Without it, a shuffle that runs either of those files first sends the sync
+// retry tests into key storage and fails them on the missing `indexedDB` global.
+const realEncryptionConfig = await import('@/db/encryption/config')
```
ital0, line 228: Small convention note: these `await import(...)` calls are not tied to a documented circular dependency. `AGENTS.md` asks us to prefer top-level imports in that case. Here the dynamic ordering seems to exist only to support the module-mock workaround.

src/test-utils/indexed-db.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/** happy-dom ships no IndexedDB, so "absent" is the real baseline. Reset to it on
+ *  both sides: an assertion that fails mid-test would otherwise leave the global
+ *  and the helper's snapshot dirty for whatever the shuffle runs next. */
+const resetToAbsent = () => {
+  restoreIndexedDb()
+  Reflect.deleteProperty(globalThis, 'indexedDB')
+}
+
+const openStub = async () =>
```
ital0, line 19: Small style note: `openStub` is a new test utility without JSDoc. `AGENTS.md` asks new utility functions to carry a short JSDoc comment.

ital0 (changes requested):
Thanks for the follow-up. I took another pass at the latest commit and reran the failing file pairs. The IndexedDB helper is in much better shape now: cleanup runs in `afterEach`, repeated stubbing restores the original global, unsupported key-storage use fails instead of hanging, and the comments now match the implementation.

The two known pending-device orders now pass. I found one more leaked encryption export, though. If `use-sync-enabled-toggle.test.ts` runs first, its mocked `needsSyncSetupWizard` survives the new sidebar guards because the spread comes from an already mocked module and only `isEncryptionEnabled` is replaced. In that order, the sidebar first renders the retry state, then the migration effect disables sync. The existing assertions run before that effect settles, so the tests remain green while observing a transient state.

I reproduced this against the current `main` integration and left the details inline. I also noted two smaller repository-convention issues in the new code.

PR #1202, opened 2026-08-06
fix(auth): keep SSO working when the app and API are on different ori…

njbrake:
…gins

Fixes two SSO failures that only appear when the app and the API are served from different origins. Both are invisible in the Compose and ALB setups, where one origin proxies both.

**Sign-in returns a bare `NOT_FOUND`.** Better Auth resolves its error redirect as `onAPIError.errorURL || ${baseURL}/error`, and `baseURL` is the API origin, which serves no HTML. Any auth failure lands on a blank 404 instead of the app. Now sets `onAPIError.errorURL` to `APP_URL`.

**The OAuth state cookie is rejected.** `__Secure-better-auth.state` is set by the API origin and read back during a redirect initiated from the app origin, so a cross-site `SameSite=Lax` cookie is dropped and sign-in fails. `account.skipStateCookieCheck` is now enabled only when `APP_URL` and the backend origin actually differ, so single-origin deployments keep the check.

PKCE still protects the exchange, and the state parameter is still round-tripped through the provider. The cookie is a second layer that a split-origin deployment cannot rely on.


ital0 (changes requested):
Thanks for digging into this one. The split-origin failure mode is subtle, and the write-up in the description made it easy to follow. The errorURL change looks good to me as is.

A few things I'd like before merging:

**skipStateCookieCheck**

This is my main concern. The state cookie is what ties the OAuth callback to the browser that started the flow. With the check skipped, someone with an account on the same IdP could start a login, capture their own callback URL (code + state) without consuming it, and hand it to a victim. The victim's browser completes the flow and ends up logged into the attacker's account, and anything the victim types lands in the attacker's data. PKCE doesn't cover this, since the verifier lives server-side in the verification row: it binds the code exchange to the server, not to the initiating browser.

I know the alternative today is SSO being fully broken on split-origin deploys, so I don't want to block on perfection. Two ideas:

1. If our split-origin deploys share a registrable domain (app.example.com + api.example.com), they're actually same-site and SameSite=Lax cookies still work. `advanced.crossSubDomainCookies` might let us keep the check there and reserve the skip for genuinely cross-site setups.
2. The current condition compares origins, which is broader than SameSite. localhost:1420 vs localhost:8000 is cross-origin but same-site, so we also drop the check in dev and in the whole test suite. Worth tightening, or at least calling out.

One wording suggestion: the comment says the cookie was "only an additional binding check", which understates the login CSRF window a bit. Naming the trade-off explicitly will help future readers.

**Possible redirect loop on persistent errors**

The frontend never reads `?error=`. In SSO mode the flow becomes: callback error → appUrl/?error=... → auth gate → /sso-redirect → auto-starts SSO on mount → IdP → error again. Recoverable errors (stale state) fix themselves on retry, which is great, but persistent ones (unlinked account, IdP misconfiguration) loop forever with no message. Before this change the bare 404 at least stopped the cycle. Could we point errorURL at a dedicated error route, or have SsoRedirect check for `?error=` and show the retry UI it already has?

**Tests**

One side effect worth knowing: createTestSettings uses localhost:1420 and localhost:8000, so every existing OIDC/SAML integration test now runs with the skip active, and the cookie-check path loses coverage. A small unit test on the shape of the createAuth options (skip present only when origins differ, errorURL set) would document the contract.

njbrake:
You were right about the state cookie, so I dropped skipStateCookieCheck entirely rather than tightening the predicate.

Login CSRF. 

better-auth 1.6.9's state.mjs confirms your read: under storeStateStrategy: "database" the verification row makes state single-use, but the signed cookie is the only browser binding. New sso-callback-state.test.ts drives the real callback: a fresh state with no cookie, and an attacker's state replayed against a victim's cookie, both stop before the token exchange. I checked that the test discriminates by re-adding the skip, at which point both reach the code exchange and the tests fail.

crossSubDomainCookies. 

It doesn't help the deployment I need, which is Railway. up.railway.app is on the Public Suffix List, so web-x.up.railway.app and api-x.up.railway.app are different registrable domains, not siblings. That also means the session cookie (SameSite=Lax, 7 day) would never reach the app even with state fixed, so the original diff would not have finished the job. I'm moving those services onto subdomains of one custom apex, which makes everything same-site and needs no code.

Redirect loop. 

errorURL now targets a dedicated /auth-error route instead of the app root. Verified in a browser against a production bundle built with VITE_AUTH_MODE=sso: /?error=... redirects into /sso-redirect and drops the error param, while /auth-error holds. The page reads both ?error= and ?state=, since one better-auth path reports state_not_found on the latter, and renders error_description as text.

Tests. 

Those 6 plus 4 on the resolved option shape. Test settings stay cross-origin, which no longer costs coverage now that the skip is gone.

Two things from the source dive. 

SAML never checked this cookie, since parseRelayState passes skipStateCookieCheck: true internally, so this only ever affected OIDC. And the state cookie's Max-Age is 300 while the verification row lives 10 minutes, so an IdP login slower than 5 minutes fails the same way. Pre-existing, so I left it.

The same-site requirement is now written up in docs/self-hosting/configuration.md.

> Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's.

backend/src/auth/sso-callback-state.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    const app = new Elysia({ prefix: '/v1' }).mount(createAuth(testEnv.db).handler)
+    handle = (request) => app.handle(request)
+  }, 60_000)
+
+  afterAll(async () => {
+    getSettingsSpy?.mockRestore()
+    globalThis.fetch = savedFetch
+    process.env.TRUSTED_ORIGINS = savedOrigins
+    if (cleanup) {
+      await cleanup().catch(() => {})
```
ital0: The repo pattern is one transaction per test with `cleanup()` in `afterEach` (backend/docs/testing.md is pretty explicit about it). The shared `beforeAll` transaction works here since each `startFlow()` writes fresh rows, but worth either matching the pattern or a short comment on why not. Also, `.catch(() => {})` hides rollback failures silently.

backend/src/auth/auth-options.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(options.onAPIError?.errorURL).not.toBe('https://app.example.com')
+    expect(options.onAPIError?.errorURL).toEndWith('/auth-error')
+  })
+
+  it('never skips the OAuth state cookie check', () => {
+    // The signed state cookie is the only thing binding a callback to the browser
+    // that started the flow — the verification row makes state single-use but not
+    // browser-bound. Skipping the check reopens login CSRF, so split-origin
+    // deployments must keep the app and API same-site instead.
+    const sameSite = optionsFor({ appUrl: 'https://app.example.com', betterAuthUrl: 'https://api.example.com' })
```
ital0: This runs in the default `consumer` mode, so `options.account` is undefined and `not.toHaveProperty` passes without exercising anything. Pass `authMode: 'oidc'` so the test would actually catch a SSO-conditional skip coming back. (The callback suite still catches it, so this is belt-and-suspenders, but right now this test is green for free.)

src/components/auth-error.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+import '@testing-library/jest-dom'
+import { render, screen } from '@testing-library/react'
+import { afterEach, describe, expect, it, mock } from 'bun:test'
+import { MemoryRouter, Route, Routes } from 'react-router'
+import AuthError from './auth-error'
+
+const replace = mock(() => {})
+const originalLocation = window.location
+
+Object.defineProperty(window, 'location', {
```
ital0: `window.location` is replaced globally and never restored; the `afterEach` only clears the mock. The fake object leaks into other test files running in the same worker. Restoring the original in an `afterAll` would fix it.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * `onAPIError.errorURL`. Deliberately does not re-enter the sign-in flow on
+ * mount: a persistent failure (unlinked account, IdP misconfiguration) would
+ * otherwise bounce between the app and the identity provider forever.
+ */
+const AuthError = () => {
+  const [searchParams] = useSearchParams()
+  // Better Auth reports the code as `?error=` on most paths, but a missing state
+  // parameter arrives as `?state=state_not_found`.
+  const code = searchParams.get('error') || searchParams.get('state') || 'unknown_error'
+  const description = searchParams.get('error_description')
```
ital0, line 46: this pageview goes to PostHog with the full query string as `$current_url` (`usePageTracking` appends `location.search`, and `sanitizeUrl` only rewrites the `/chats/:id` pathname). With telemetry enabled, an IdP-supplied `error_description` lands in analytics. Stripping the query for this route in `sanitizeUrl`, or clearing the params here after reading them, should cover it.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <div className="flex max-w-md flex-col items-center gap-8 px-6 text-center">
+        <div className="flex items-center gap-1.5 text-sm text-muted-foreground">
+          <AppLogo size={16} />
+          <span>Thunderbolt</span>
+        </div>
+
+        <div className="flex flex-col items-center gap-2">
+          <h1 className="text-4xl font-semibold tracking-tight">Sign-in failed</h1>
+          <p className="text-muted-foreground">{describeAuthError(code)}</p>
+          <p className="text-[length:var(--font-size-xs)] text-muted-foreground">{description || code}</p>
```
ital0: When `error_description` is present, the raw code disappears from the page, though the JSDoc above says the code is always shown so support tickets keep the real signal. Showing both (labeled) would match the comment. Confirmed on the preview deploy: `?error=idp_exploded&error_description=Client+not+registered` renders only the description.

docs/self-hosting/configuration.md
```diff
@@ -118,6 +118,14 @@ Tested with BetterStack, Jaeger, Zipkin, New Relic, Grafana Cloud, and any OTLP-
...
 | `MONITORING_TOKEN` | —                       | Shared secret for authenticated `/health` checks                     |
 
+### Serving the app and API on separate hostnames
+
+The simplest deployments put the frontend and the backend behind one origin (Compose, the ALB, a k8s ingress). If you split them, `APP_URL` and `BETTER_AUTH_URL` must stay **same-site**: two hostnames under one registrable domain, such as `app.example.com` and `api.example.com`. Different ports or subdomains of a shared parent are fine.
+
+Hostnames on different registrable domains break sign-in. The auth cookies are `SameSite=Lax`, so a cross-site deployment loses both the OAuth `state` cookie (the callback fails with `state_security_mismatch` after a successful IdP login) and the session cookie (`/get-session` returns 401 from the app origin). Watch out for PaaS hostnames that look like subdomains but are not: `up.railway.app`, `vercel.app`, and similar are on the [Public Suffix List](https://publicsuffix.org/list/), which makes `web.up.railway.app` and `api.up.railway.app` cross-site. Attach custom domains under one apex instead.
```
ital0: The redirect actually delivers `?error=state_mismatch` (the new callback tests assert exactly that), so an operator searching logs for `state_security_mismatch` won't find it. Worth using the delivered code here.

ital0 (changes requested):
Nice work on the follow-up.

A few small things that I believe worth to take a look.

njbrake:
Thanks @ital0 ! I think Claude and I took care of those remaining issues. And please let me know if you would like the PRs presented in a different way, I know I sent a whole batch of them at once, if you want them submitted more gradually of if you want more or less AI involved in these PRs, let me know :) I saw and used a handful of the `thunder` skills you have in the repo so I have the feeling you all are AI friendly but just let me know if you want any different communication methods 🙏

backend/src/auth/auth.ts
```diff
@@ -157,6 +157,20 @@ export const createAuth = (database: typeof DbType, emailDeps: AuthEmailDeps = {
...
+    // the API domain instead of a page the user can retry from. Only observable
+    // when the app and API are on separate origins; where one origin proxies
+    // both, the SPA's catch-all already absorbed this.
+    //
+    // Points at /auth-error rather than the app root on purpose: in SSO mode the
+    // root is behind the auth gate, which sends unauthenticated users to
+    // /sso-redirect, which starts a new IdP round-trip on mount. A persistent
+    // failure (unlinked account, IdP misconfiguration) would loop through the
+    // IdP forever. /auth-error is unguarded and terminal.
+    onAPIError: { errorURL: `${settings.appUrl}/auth-error` },
```
ital0: Could we cover the SAML sibling before treating this as the terminal SSO error target? In `@better-auth/sso` 1.6.9, `processSAMLResponse()` ignores `relayState.errorURL` and sends validation or linking failures to `relayState.callbackURL`. Thunderbolt sets that to `/`, so `AuthGate` starts SSO again and the loop remains for SAML. A SAML failure test plus a flow-specific error callback would protect this path.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+
+        <div className="flex flex-col items-center gap-2">
+          <h1 className="text-4xl font-semibold tracking-tight">Sign-in failed</h1>
+          <p className="text-muted-foreground">{describeAuthError(code)}</p>
+          {description && <p className="text-[length:var(--font-size-xs)] text-muted-foreground">{description}</p>}
+          <p className="text-[length:var(--font-size-xs)] text-muted-foreground">Error code: {code}</p>
+        </div>
+
+        <Button onClick={() => window.location.replace(isSsoMode() ? '/sso-redirect' : '/')}>Try again</Button>
```
ital0: Could the desktop flow get its own failure callback instead of landing on this web retry? `desktop-initiate` gives Better Auth only the loopback success URL. On an OIDC failure, the system browser opens `/auth-error` while `startSsoFlowLoopback()` waits five minutes; clicking this button starts a normal web flow and never settles the desktop promise. Forwarding the error to the loopback would let the app fail and retry immediately.

src/lib/posthog.tsx
```diff
@@ -44,6 +51,10 @@ export const sanitizeUrl = (url: string): string => {
     }
   })()
 
+  if (queryStrippedRoutes.some((route) => route === pathname)) {
```
ital0: Could we close the remaining analytics paths rather than only rewriting selected event properties? PostHog also adds `$session_entry_url`, and in same-origin deployments analytics requests carry the full page URL in `Referer`, which our proxy forwards. The `error_description` can therefore still reach PostHog. Removing the sensitive query before PostHog initializes, sanitizing every SDK URL field, or dropping `Referer` at the proxy would make the guarantee here hold.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <div className="flex max-w-md flex-col items-center gap-8 px-6 text-center">
+        <div className="flex items-center gap-1.5 text-sm text-muted-foreground">
+          <AppLogo size={16} />
+          <span>Thunderbolt</span>
+        </div>
+
+        <div className="flex flex-col items-center gap-2">
+          <h1 className="text-4xl font-semibold tracking-tight">Sign-in failed</h1>
+          <p className="text-muted-foreground">{describeAuthError(code)}</p>
+          {description && <p className="text-[length:var(--font-size-xs)] text-muted-foreground">{description}</p>}
```
ital0: I may be missing a reason we need the raw provider text, but could we avoid rendering it verbatim? `/auth-error` is public, so anyone can craft an app URL that displays arbitrary instructions under Thunderbolt's branding. React escapes markup, but it does not prevent misleading copy. Keeping mapped text and the stable code, or clearly labeling and tightly limiting the description, would avoid that. It would also prevent a missing IdP description from showing the literal word `undefined`.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { AppLogo } from '@/components/app-logo'
+import { Button } from '@/components/ui/button'
+import { isSsoMode } from '@/lib/auth-mode'
+import { useSearchParams } from 'react-router'
+
+/** Codes that mean the flow went stale rather than that it can never succeed. */
+const staleFlowCodes = new Set([
+  'invalid_state',
+  'please_restart_the_process',
+  'state_mismatch',
```
ital0: Could we take `state_mismatch` out of the stale bucket? The callback tests show it is returned when the signed cookie is missing or belongs to another browser, and the docs say a cross-site setup produces it immediately. In those cases the login did not take too long, and retrying cannot fix the configuration.

src/components/auth-error.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * Unmapped codes get generic copy; the code is always rendered alongside it so a
+ * support ticket carries the real signal even when the provider also sent a
+ * description.
+ */
+const describeAuthError = (code: string): string => {
+  if (staleFlowCodes.has(code)) {
+    return 'Your sign-in took too long to complete. Starting over usually fixes this.'
+  }
+  if (code === 'account_not_linked') {
+    return 'This email is already registered with a different sign-in method. Ask your administrator to link the accounts.'
```
ital0: Could we change this recovery advice? Thunderbolt does not expose an admin account-linking flow, and trusted SSO providers should link matching emails automatically. If this code appears, asking the user to sign in with the original method or contact the administrator to check the identity mapping seems more accurate than asking them to link accounts.

docs/self-hosting/configuration.md
```diff
@@ -118,6 +118,14 @@ Tested with BetterStack, Jaeger, Zipkin, New Relic, Grafana Cloud, and any OTLP-
 | `SWAGGER_ENABLED`  | `false`                 | Expose `/v1/swagger` with the full OpenAPI spec (don't in production) |
 | `MONITORING_TOKEN` | —                       | Shared secret for authenticated `/health` checks                     |
 
+### Serving the app and API on separate hostnames
+
+The simplest deployments put the frontend and the backend behind one origin (Compose, the ALB, a k8s ingress). If you split them, `APP_URL` and `BETTER_AUTH_URL` must stay **same-site**: two hostnames under one registrable domain, such as `app.example.com` and `api.example.com`. Different ports or subdomains of a shared parent are fine.
```
ital0: Could we make the complete split-origin requirement explicit here? SameSite is schemeful, so the URLs need the same scheme and registrable domain. The app origin also needs `CORS_ORIGINS` and `TRUSTED_ORIGINS`, credentialed CORS, and an absolute `VITE_THUNDERBOLT_CLOUD_URL` pointing at the API. Same-site hostnames alone are not enough.

docs/self-hosting/configuration.md
```diff
@@ -118,6 +118,14 @@ Tested with BetterStack, Jaeger, Zipkin, New Relic, Grafana Cloud, and any OTLP-
...
+
+Hostnames on different registrable domains break sign-in. The auth cookies are `SameSite=Lax`, so a cross-site deployment loses both the OAuth `state` cookie (after a successful IdP login the callback redirects to `APP_URL/auth-error?error=state_mismatch`, and the backend logs `state_security_mismatch`) and the session cookie (`/get-session` returns 401 from the app origin). Watch out for PaaS hostnames that look like subdomains but are not: `up.railway.app`, `vercel.app`, and similar are on the [Public Suffix List](https://publicsuffix.org/list/), which makes `web.up.railway.app` and `api.up.railway.app` cross-site. Attach custom domains under one apex instead.
```
ital0: Small correction: Better Auth 1.6.9 returns `200 null` from `/get-session` when the session cookie is absent, not 401. The app still sees no session, but the actual status matters when an operator is debugging this setup.

backend/src/auth/sso-callback-state.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+describe('SSO callback state binding', () => {
+  let handle: (request: Request) => Promise<Response>
+  let cleanup: () => Promise<void>
+  let getSettingsSpy: ReturnType<typeof spyOn>
+  const savedFetch = globalThis.fetch
+  const savedOrigins = process.env.TRUSTED_ORIGINS
+
+  beforeAll(() => {
+    globalThis.fetch = Object.assign(stubbedFetch, { preconnect: () => {} }) as unknown as typeof fetch
```
ital0: Small convention note: could we type the composed fetch stub as `typeof fetch` at its declaration instead of using `as unknown as`? The chained assertion bypasses the contract at the exact boundary this test is supposed to model.

backend/src/auth/auth.ts
ital0, line 180: `onAPIError.errorURL` does not cover the SAML result path in `@better-auth/sso` 1.6.9. `processSAMLResponse()` ignores `relayState.errorURL` and sends validation or linking failures to `relayState.callbackURL`. Thunderbolt sets that to `/`, so `AuthGate` starts SSO again and the redirect loop remains for SAML.

src/components/auth-error.tsx
ital0, line 90: Desktop OIDC failures land on this web page while `startSsoFlowLoopback()` continues waiting for its localhost callback for up to five minutes. This button starts a normal web flow in the system browser and never settles the original desktop promise, leaving the app pending until timeout.

src/lib/posthog.tsx
ital0: This only sanitizes selected event properties. PostHog also records `$session_entry_url`, and same-origin analytics requests carry the full page URL in `Referer`, which the backend proxy forwards. `error_description` can still reach PostHog despite the test asserting that it never does.

src/components/auth-error.tsx
ital0: `/auth-error` is public and renders `error_description` verbatim under Thunderbolt branding. A crafted URL can display arbitrary instructions. React escapes markup, but it does not prevent misleading text. When an IdP omits the description, the SSO plugin can also forward the literal value `undefined`.

src/components/auth-error.tsx
ital0, line 20: `state_mismatch` does not mean the flow went stale. The callback tests use it for a missing or different-browser cookie, and the docs use it for persistent cross-site configuration. The page therefore reports a timeout and recommends retry for failures that happened immediately and cannot be fixed by retry.

src/components/auth-error.tsx
ital0: Thunderbolt has no administrator account-linking flow, while trusted SSO providers already link matching emails automatically. This message directs the user to an unavailable action and hides the more likely original-sign-in or identity-mapping problem.

docs/self-hosting/configuration.md
ital0: This split-origin recipe is incomplete. SameSite is schemeful, so the URLs need the same scheme as well as the same registrable domain. Split origins also depend on `CORS_ORIGINS`, `TRUSTED_ORIGINS`, credentialed CORS, and an absolute `VITE_THUNDERBOLT_CLOUD_URL` pointing at the API. A deployment can follow the current paragraph and still fail sign-in.

docs/self-hosting/configuration.md
ital0: Better Auth 1.6.9 returns `200 null` from `/get-session` when the session cookie is absent, not 401. The documented status points operators at the wrong troubleshooting symptom.

backend/src/auth/sso-callback-state.test.ts
ital0: `as unknown as typeof fetch` bypasses the fetch contract. The test can continue compiling after the stub stops matching Bun's real fetch signature, which weakens this process-wide boundary.

ital0 (commented):
Thanks for the careful follow-up. The final direction resolves the state-cookie and automatic redirect-loop concerns from my earlier reviews, and the callback tests add useful coverage.

I did one more line-by-line pass and left focused comments on two sibling flows that still fall outside the new landing page, SAML and desktop loopback, plus the remaining trust and privacy boundary around error query data. The other notes are smaller copy, test, and self-hosting documentation corrections.

One general housekeeping item: `main` now contains broader URL sanitization from [#1210](https://github.com/thunderbird/thunderbolt/pull/1210), so please preserve that implementation while resolving the current conflict.

The overall direction makes sense to me. I think addressing the sibling flows and query-data handling will make the fix hold across the deployment modes this project supports.

ital0 (changes requested):

backend/src/auth/sso-desktop-callback.ts
```diff
@@ -78,10 +78,17 @@ export const createSsoDesktopCallbackRoutes = (settings: Settings) => {
...
+          // /auth-error page), which leaves this flow's loopback server waiting
+          // out its five-minute timeout with nothing to receive. The port lives
+          // in the path because the SSO plugin appends `?error=` with a literal
+          // `?`, which would corrupt an existing query string.
+          const errorCallbackURL = `${baseUrl}/v1/api/auth/sso/desktop-error/${port}`
           const ssoResponse = await fetch(`${baseUrl}/v1/api/auth/sign-in/sso`, {
             method: 'POST',
             headers: { 'Content-Type': 'application/json' },
-            body: JSON.stringify({ providerId: 'sso', callbackURL }),
+            body: JSON.stringify({ providerId: 'sso', callbackURL, errorCallbackURL }),
```
ital0, line 91: Thanks for adding the desktop error callback. I traced this through Better Auth and `@better-auth/sso` 1.6.9, and two branches still bypass it. An OIDC state failure happens before Better Auth can recover `errorCallbackURL`, so it falls back to the web `/auth-error`. SAML stores this value in RelayState, but `processSAMLResponse()` redirects with `callbackURL` instead. Neither branch reaches the loopback server, leaving the desktop flow pending until its five-minute timeout.

ital0 (changes requested):
Thanks for the careful follow-up. The latest changes address most of my earlier comments.

I left one remaining inline note on the desktop error callback. In the versions pinned by the backend, OIDC state failures and SAML failures still bypass the loopback error path, leaving the desktop pending until the five-minute timeout.

PR #1203, opened 2026-08-06
fix(backend): wire THUNDERBOLT_INFERENCE_URL through the inference proxy

njbrake:

Makes `THUNDERBOLT_INFERENCE_URL` work. It is documented in `docs/self-hosting/configuration.md` as the way to point a self-hosted backend at an OpenAI-compatible gateway, but nothing read it: the variable was absent from `settings.ts` and from the inference client, so setting it had no effect.

Adds the setting, a `thunderbolt-inference` provider in the inference client, and routing in `inference/routes.ts` for model ids the built-in table does not cover.

Models are **discovered** from the endpoint's `GET /models` rather than listed by hand, so a gateway that gains a model needs no redeploy. Discovery is cached for five minutes and refreshed when the app fetches `/config`. If the endpoint is unreachable its models are omitted and the built-ins keep working, so a gateway outage cannot break boot.

`THUNDERBOLT_INFERENCE_MODELS` stays optional and now does two things when set: restricts which discovered models are exposed, and supplies display names.

Requests are proxied through the backend, so the gateway key stays server side and the gateway needs no CORS configuration. Capabilities are advertised conservatively because the endpoint cannot be introspected.

Verified against a live gateway: 16 discovered models, and server logs confirming `provider="thunderbolt-inference"` with a 200 on a real completion.


backend/src/inference/gateway-models.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Publishable row for a gateway model.
+ *
+ * Capabilities are deliberately conservative, because a gateway can front any
+ * model and we cannot introspect it: `vendor: null` means images are stripped
+ * (`vendorSupportsImages` does not guess), `contextWindow: null` falls back to
+ * the transport default, and parallel tool calls are off since advertising them
+ * to a model that rejects them turns every tool call into an upstream error.
+ */
+const toSharedModel = (spec: GatewayModelSpec): SharedModel => ({
+  id: uuidV5(spec.id, gatewayIdNamespace),
```
ital0, line 281: Could I be missing a step here? These generated IDs have no bundled model profile. The frontend reconciler explicitly drops OTA model IDs without one, so the discovered models never reach the database or picker.

backend/src/api/config.ts
```diff
@@ -17,8 +18,29 @@ import { Elysia } from 'elysia'
...
+const buildModelDefaults = async (settings: Settings) => {
+  // Refresh discovery here rather than at boot: /config is fetched on every app
+  // start, so the catalogue is picked up without a redeploy, and a warm cache
+  // makes this a no-op.
+  await ensureGatewayModels(settings)
+  const gatewayModels = getGatewaySharedModels(settings)
+  if (gatewayModels.length === 0) {
+    return { version: defaultModelsVersion, data: defaultModels }
+  }
+  return { version: defaultModelsVersion + 1, data: [...defaultModels, ...gatewayModels] }
```
ital0: This version is unchanged for every non-empty discovery result. Once a client records `defaultModelsVersion + 1`, the strict version gate ignores later model additions, removals, and label changes from the gateway.

backend/src/api/config.ts
```diff
@@ -17,8 +18,29 @@ import { Elysia } from 'elysia'
...
+ * `pickDefaults` requires before it will trust a server payload.
+ */
+const buildModelDefaults = async (settings: Settings) => {
+  // Refresh discovery here rather than at boot: /config is fetched on every app
+  // start, so the catalogue is picked up without a redeploy, and a warm cache
+  // makes this a no-op.
+  await ensureGatewayModels(settings)
+  const gatewayModels = getGatewaySharedModels(settings)
+  if (gatewayModels.length === 0) {
+    return { version: defaultModelsVersion, data: defaultModels }
```
ital0: I think an empty catalog and a failed discovery become the same state here. If the higher gateway version was already applied, the client will not roll back to this bundled version, so stale gateway models remain visible while the backend no longer recognizes them.

backend/src/inference/client.ts
```diff
@@ -252,6 +260,50 @@ const getAnthropicClient = (options: InferenceClientOptions = {}): OpenAI | Post
...
+    return thunderboltInferenceClient
+  }
+
+  const settings = getSettings()
+
+  if (!settings.thunderboltInferenceUrl) {
+    throw new Error('Thunderbolt inference URL not configured')
+  }
+
+  if (!settings.thunderboltInferenceApiKey) {
```
ital0: Discovery allows a gateway with no API key, and the self-hosting docs mention Ollama and llama.cpp, but the inference client rejects that same configuration here. The models can be listed and then fail before a completion request is sent.

backend/src/inference/gateway-models.ts
```diff
@@ -0,0 +1,230 @@
...
+    })
+    if (!response.ok) {
+      console.warn(`Inference gateway model discovery failed: HTTP ${response.status}`)
+      return []
+    }
+    const body = (await response.json()) as ModelsResponse
+    const ids = (body.data ?? []).map((entry) => entry?.id).filter((id): id is string => typeof id === 'string' && !!id)
+
+    // A non-empty allowlist narrows discovery; an empty one accepts everything.
+    const allowed = overrides.size > 0 ? ids.filter((id) => overrides.has(id)) : ids
```
ital0, line 177: A non-empty allowlist containing only invalid entries produces an empty map, which this branch treats as no allowlist. In that case every model advertised by the gateway is exposed, although the operator attempted to restrict the list.

backend/src/inference/routes.ts
```diff
@@ -125,7 +127,20 @@ export const createInferenceRoutes = (options: CreateInferenceRoutesOptions) =>
...
+      // built-in id by listing it in THUNDERBOLT_INFERENCE_MODELS. Gateway ids
+      // pass through unchanged: the gateway knows them by their own name.
+      // Refresh discovery only when the id is not a built-in, so the common path
+      // never awaits the network. Warm cache makes this a no-op anyway.
+      const settings = getSettings()
+      if (!supportedModels[body.model]) {
+        await ensureGatewayModels(settings)
+      }
+      const modelConfig =
+        supportedModels[body.model] ??
```
ital0, line 130: A gateway model with the same slug as a built-in is still published as a separate row. When that row is selected, this lookup chooses the built-in provider, so the request goes somewhere different from the model shown in the picker.

backend/src/inference/gateway-models.ts
```diff
@@ -0,0 +1,230 @@
...
+ * the app from booting or serving its built-in models, and the next refresh
+ * after the TTL picks it back up.
+ */
+const discover = async (settings: GatewaySettings, fetchFn: typeof fetch): Promise<GatewayModelSpec[]> => {
+  const overrides = new Map(parseGatewayModelSpecs(settings.thunderboltInferenceModels).map((s) => [s.id, s.label]))
+  // `thunderboltInferenceUrl` already includes the `/v1` prefix by convention.
+  const url = `${settings.thunderboltInferenceUrl.replace(/\/$/, '')}/models`
+
+  try {
+    const response = await fetchFn(url, {
```
ital0, line 163: This fetch has no timeout or abort signal. It is awaited by public `/config` and by the chat route on a cold cache, so an unresponsive gateway can hold those backend requests until the transport or server times out.

backend/src/inference/gateway-models.ts
```diff
@@ -0,0 +1,230 @@
...
+    cache !== null &&
+    cache.url === settings.thunderboltInferenceUrl &&
+    cache.allowlist === settings.thunderboltInferenceModels &&
+    now - cache.fetchedAt < discoveryTtlMs
+
+  if (fresh && cache) {
+    return cache.specs
+  }
+
+  const specs = await discover(settings, fetchFn)
```
ital0: Concurrent cold requests can all reach this await before the cache is written. A slower failed request can then overwrite a successful discovery with an empty catalog for the full TTL.

backend/src/config/settings.ts
```diff
@@ -177,6 +190,9 @@ const parseSettings = (): Settings => {
     exaApiKey: process.env.EXA_API_KEY || '',
     tinfoilApiKey: process.env.TINFOIL_API_KEY || '',
     tinfoilEnclaveUrl: process.env.TINFOIL_ENCLAVE_URL || 'https://inference.tinfoil.sh/v1',
+    thunderboltInferenceUrl: process.env.THUNDERBOLT_INFERENCE_URL || '',
+    thunderboltInferenceApiKey: process.env.THUNDERBOLT_INFERENCE_API_KEY || '',
+    thunderboltInferenceModels: process.env.THUNDERBOLT_INFERENCE_MODELS || '',
```
ital0, line 195: `THUNDERBOLT_INFERENCE_MODELS` is read here, but the Helm, Pulumi, preview, and workflow paths still pass only URL and key. The documented allowlist and labels are unavailable through those deployment paths.

backend/src/api/config.test.ts
```diff
@@ -60,5 +65,40 @@ describe('Config Routes', () => {
...
+
+    it('appends inference gateway models and out-versions the bundled defaults', async () => {
+      const settings = createTestSettings({
+        thunderboltInferenceUrl: 'https://gateway.example.com/v1',
+        thunderboltInferenceApiKey: 'key',
+        thunderboltInferenceModels: 'llama-3.3-70b=Llama 3.3 70B',
+      })
+      // Models are discovered from the gateway, so prime the cache with a stubbed
+      // /models response; the route then reads it without any network access.
+      await ensureGatewayModels(settings, {
```
ital0: This test warms the module cache before calling `/config`, so it never exercises the route's own discovery call. It would still pass if that call were removed.

backend/src/inference/routes.ts
```diff
@@ -125,7 +127,20 @@ export const createInferenceRoutes = (options: CreateInferenceRoutesOptions) =>
...
 
-      const modelConfig = supportedModels[body.model]
+      // Shipped models win over gateway models, so a deployment cannot shadow a
+      // built-in id by listing it in THUNDERBOLT_INFERENCE_MODELS. Gateway ids
+      // pass through unchanged: the gateway knows them by their own name.
+      // Refresh discovery only when the id is not a built-in, so the common path
+      // never awaits the network. Warm cache makes this a no-op anyway.
+      const settings = getSettings()
+      if (!supportedModels[body.model]) {
+        await ensureGatewayModels(settings)
```
ital0: The route already receives an injected `fetchFn`, but discovery falls back to the global fetch here. That leaves the new branch outside the existing test seam.

backend/src/inference/gateway-models.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      { id: 'kimi', label: 'kimi' },
+    ])
+    expect(fetchFn).toHaveBeenCalledTimes(1)
+  })
+
+  it('requests /models on the configured base URL with the api key', async () => {
+    const fetchFn = modelsResponse(['a'])
+    await ensureGatewayModels(configured(), { fetchFn: fetchFn as never })
+
+    const [url, init] = (fetchFn as unknown as { mock: { calls: [string, RequestInit][] } }).mock.calls[0]
```
ital0: This chained `as unknown as` bypasses the type check around the recorded request. It also conflicts with the repository rule against this assertion pattern.

backend/.env.example
```diff
@@ -17,6 +17,8 @@ FIREWORKS_API_KEY=
 MISTRAL_API_KEY=
 THUNDERBOLT_INFERENCE_URL=https://animal.inference.thunderbolt.io/v1
 THUNDERBOLT_INFERENCE_API_KEY=
+# Comma-separated `id` or `id=Label` entries. Nothing routes to the endpoint until set.
```
ital0: This comment says no model is routed until the list is set, while the implementation and self-hosting docs define an empty list as discover-all. The two setup paths currently describe opposite behavior.

ital0 (changes requested):
Thanks for restoring `THUNDERBOLT_INFERENCE_URL`. The backend discovery and proxy path make sense, but I do not think the PR works end to end yet.

The main issue is that discovered models are published through reconciled defaults. The current frontend drops unknown model IDs without bundled profiles, so these models never reach the database or picker. The fixed version also means later catalog changes are ignored, while an empty or failed discovery can leave the frontend and backend with different model lists.

I also found inconsistencies around keyless gateways, allowlist parsing, model ID collisions, discovery timeouts, concurrent cache refreshes, and the deployment plumbing for `THUNDERBOLT_INFERENCE_MODELS`.

Because these issues affect the self-hosting flow described by the PR, my review is request changes.

backend/src/inference/gateway-models.ts
njbrake: The reconciler dropped OTA ids lacking a bundled profile, so I moved gateway models off the version-gated `defaults.models` channel into a separate `defaults.gatewayModels` field, reconciled through a dedicated path that synthesizes a profile per row. A picker-DAL test confirms they now reach the picker.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/api/config.ts
njbrake: Removed the synthetic version bump. Gateway models no longer ride the version-gated channel, so later catalogue additions, removals, and label changes always apply.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/api/config.ts
njbrake: Discovery now returns a typed ok/failure result: a reachable-but-empty catalogue stays empty, while a failure keeps the last good catalogue instead of blanking it, so the two states no longer converge.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/client.ts
njbrake: The gateway client no longer requires a key (URL still required), so a keyless Ollama/llama.cpp gateway works on the completion path as well as discovery.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/gateway-models.ts
njbrake: Now fails closed: a configured allowlist that parses to zero valid entries exposes nothing (with a warning) instead of falling through to expose-all.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/routes.ts
njbrake: Gateway ids that collide with a built-in slug are now dropped at discovery (the built-in table moved to `supported-models.ts` to avoid an import cycle), so no shadow picker row can route to a different provider than it shows.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/gateway-models.ts
njbrake: Added `AbortSignal.timeout` to the discovery fetch so a hung gateway cannot stall `/config` or the chat cold path.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/gateway-models.ts
njbrake: Concurrent cold callers now share one in-flight request keyed by settings, and a failed refresh keeps the last good catalogue rather than overwriting it with empty.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/config/settings.ts
njbrake: THUNDERBOLT_INFERENCE_MODELS is now threaded through the k8s chart (values + backend template), Pulumi (index, services, per-pr-stack), and the stack-deploy/preview-deploy workflows, mirroring the non-secret URL handling.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/api/config.test.ts
njbrake: The config route now takes an injectable `fetchFn` and the test drives discovery through the route, so the route runs its own discovery call rather than reading a pre-warmed cache.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/routes.ts
njbrake: Discovery in the route now uses the injected `fetchFn` instead of global fetch, keeping it inside the existing test seam.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/src/inference/gateway-models.test.ts
njbrake: Replaced the chained assertion by typing the fetch mock with fetch parameters, so `.mock.calls` is typed and no `as unknown as` is needed.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

backend/.env.example
njbrake: Reworded in both `.env.example` files: empty exposes every advertised model; a set list restricts to those ids and supplies picker labels.

_Drafted by Claude Opus 4.8 via back-and-forth with @njbrake (his decisions, Claude's prose)._

PR #1204, opened 2026-08-06
fix(deploy): run the bundled Keycloak on Postgres instead of dev-mode H2

njbrake:

Runs the bundled Keycloak in production mode against Postgres instead of `start-dev` with file-backed H2.

`start-dev` implies H2, which Keycloak documents as development only. Combined with an orchestrator that mounts no volume for the service, the realm becomes ephemeral: every redeploy re-imports the realm JSON and discards everything created since. Observed on a live deployment:

- users added in the admin console vanish
- master-realm settings such as brute-force protection reset to off
- there is no login or admin event history to inspect
- `KC_BOOTSTRAP_ADMIN_*` becomes load-bearing on every boot, when it exists for the first boot only

Two-stage build so the image can run `start --optimized`: the database vendor is baked in by `kc.sh build`, while hostname, DB URL and credentials stay runtime options.

Keycloak gets its own database rather than a schema in the app's. It owns roughly 95 tables managed by Liquibase, which has no business sharing a namespace with the app's Drizzle migrations. `postgres-init` creates it alongside `powersync_storage`, so fresh stacks need nothing extra.

One behaviour change worth calling out: `--import-realm` only applies to an empty database, so realm JSON edits no longer take effect on an existing stack. That is the point, since console changes now survive redeploys, but it moves realm configuration from the file to the console.

### DIFF
--- a/deploy/docker/keycloak.Dockerfile
+++ b/deploy/docker/keycloak.Dockerfile
@@ -1,5 +1,34 @@
+# Keycloak in production mode (`start`), not `start-dev`.
+#
+# `start-dev` implies the file-backed H2 database, which Keycloak documents as
+# development-only. On an orchestrator without a persistent volume that makes the
+# whole realm ephemeral: every redeploy re-imports the realm JSON and discards
+# everything created since — users added in the admin console, master-realm
+# settings such as brute-force protection, and the login/admin event history. It
+# also leaves `KC_BOOTSTRAP_ADMIN_*` load-bearing on every boot, when it is meant
+# for the first boot only.
+#
+# Two stages so the image can run `start --optimized`: build-time options (notably
+# the database vendor) are baked in by `kc.sh build`, and startup then skips
+# re-deriving them. Runtime options — hostname, DB URL, credentials — still come
+# from the environment.
+FROM quay.io/keycloak/keycloak:26.7 AS builder
+
+# Build-time option: selects the JDBC driver and dialect compiled into the image.
+# KC_DB_URL / KC_DB_USERNAME / KC_DB_PASSWORD remain runtime options.
+ENV KC_DB=postgres
+
+RUN /opt/keycloak/bin/kc.sh build
+
 FROM quay.io/keycloak/keycloak:26.7
 
+COPY --from=builder /opt/keycloak/ /opt/keycloak/
+
+# `--import-realm` reads this on the first boot against an EMPTY database. Once
+# the realm exists the import is skipped, which is the whole point of moving off
+# H2: changes made in the admin console now survive a redeploy. The flip side is
+# that edits to this file (or to KC_SEED_*) no longer take effect on a database
+# that already holds the realm — change those in the console, or drop the schema.
 COPY deploy/config/keycloak-realm.json /opt/keycloak/data/import/thunderbolt-realm.json
 
 # The base image already runs as user 1000; keep this explicit for clarity and scanners.
@@ -8,4 +37,4 @@ USER 1000
 EXPOSE 8080
 
 ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
-CMD ["start-dev", "--import-realm"]
+CMD ["start", "--optimized", "--import-realm"]
--- a/deploy/docker/keycloak.Dockerfile
+++ b/deploy/docker/keycloak.Dockerfile
@@ -1,17 +1,20 @@
-# Keycloak in production mode (`start`), not `start-dev`.
+# Prod-capable Keycloak image. The default command is `start-dev`, which runs on
+# the file-backed H2 database — fine for local and other ephemeral use. A
+# deployment that needs state to survive a redeploy opts into production mode by
+# overriding the command with `start --optimized` and pointing KC_DB_URL /
+# KC_DB_USERNAME / KC_DB_PASSWORD at Postgres (see the Railway deployment).
 #
-# `start-dev` implies the file-backed H2 database, which Keycloak documents as
-# development-only. On an orchestrator without a persistent volume that makes the
-# whole realm ephemeral: every redeploy re-imports the realm JSON and discards
-# everything created since — users added in the admin console, master-realm
-# settings such as brute-force protection, and the login/admin event history. It
-# also leaves `KC_BOOTSTRAP_ADMIN_*` load-bearing on every boot, when it is meant
-# for the first boot only.
+# `start-dev` implies H2, which Keycloak documents as development-only: on an
+# orchestrator without a persistent volume the realm is ephemeral — every redeploy
+# re-imports the realm JSON and discards everything created since (admin-console
+# users, master-realm settings such as brute-force protection, and login/admin
+# event history), and it leaves `KC_BOOTSTRAP_ADMIN_*` load-bearing on every boot
+# when it is meant for the first boot only.
 #
-# Two stages so the image can run `start --optimized`: build-time options (notably
-# the database vendor) are baked in by `kc.sh build`, and startup then skips
-# re-deriving them. Runtime options — hostname, DB URL, credentials — still come
-# from the environment.
+# Two stages so `start --optimized` is available to deployments that opt in:
+# build-time options (notably the database vendor) are baked in by `kc.sh build`,
+# and startup then skips re-deriving them. Runtime options — hostname, DB URL,
+# credentials — still come from the environment.
 FROM quay.io/keycloak/keycloak:26.7 AS builder
 
 # Build-time option: selects the JDBC driver and dialect compiled into the image.
@@ -24,11 +27,12 @@ FROM quay.io/keycloak/keycloak:26.7
 
 COPY --from=builder /opt/keycloak/ /opt/keycloak/
 
-# `--import-realm` reads this on the first boot against an EMPTY database. Once
-# the realm exists the import is skipped, which is the whole point of moving off
-# H2: changes made in the admin console now survive a redeploy. The flip side is
-# that edits to this file (or to KC_SEED_*) no longer take effect on a database
-# that already holds the realm — change those in the console, or drop the schema.
+# `--import-realm` runs on every boot but imports only realms that do not yet
+# exist; once the thunderbolt realm is present the import is skipped. For a
+# deployment on Postgres that is the point: changes made in the admin console
+# survive a redeploy. The flip side is that later edits to this realm JSON no
+# longer take effect once the database already holds the realm — make those
+# changes in the admin console.
 COPY deploy/config/keycloak-realm.json /opt/keycloak/data/import/thunderbolt-realm.json
 
 # The base image already runs as user 1000; keep this explicit for clarity and scanners.
@@ -37,4 +41,4 @@ USER 1000
 EXPOSE 8080
 
 ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
-CMD ["start", "--optimized", "--import-realm"]
+CMD ["start-dev", "--import-realm"]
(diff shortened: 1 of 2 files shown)


deploy/docker/keycloak.Dockerfile
```diff
@@ -8,4 +37,4 @@ USER 1000
 EXPOSE 8080
 
 ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
-CMD ["start-dev", "--import-realm"]
+CMD ["start", "--optimized", "--import-realm"]
```
ital0: Blocker: every checked-in deployment bypasses this command. Compose and Helm still set `start-dev --import-realm`, and both Pulumi task definitions override the image `CMD` with the same command (`deploy/docker-compose.yml:108`, `deploy/k8s/templates/keycloak.yaml:34`, `deploy/pulumi/src/services.ts:315`, `deploy/pulumi/src/shared.ts:543`). I ran the published PR image with that override; it rebuilt itself for the dev profile and started on H2. Can we update all consumers in this PR so the production command actually runs?

ital0: Blocker: once this command is used, the current network setup still cannot start it. All stacks send plain HTTP to port 8080 behind ALB or Ingress, but none sets `KC_HTTP_ENABLED=true` or provides Keycloak TLS key material. The published image exits with `Key material not provided to setup HTTPS`. Please wire edge TLS termination explicitly.

deploy/docker/keycloak.Dockerfile
```diff
@@ -1,5 +1,34 @@
...
+# for the first boot only.
+#
+# Two stages so the image can run `start --optimized`: build-time options (notably
+# the database vendor) are baked in by `kc.sh build`, and startup then skips
+# re-deriving them. Runtime options — hostname, DB URL, credentials — still come
+# from the environment.
+FROM quay.io/keycloak/keycloak:26.7 AS builder
+
+# Build-time option: selects the JDBC driver and dialect compiled into the image.
+# KC_DB_URL / KC_DB_USERNAME / KC_DB_PASSWORD remain runtime options.
```
ital0, line 21: Blocker: none of the Compose, Helm, or Pulumi Keycloak definitions sets these values. The published image therefore tries `localhost:5432/keycloak` even though Postgres runs as another service, and it exits. Please add the URL and credentials for every deployment target that uses this image.

deploy/docker/postgres-init/01-powersync.sh
```diff
@@ -33,4 +33,11 @@ psql -v ON_ERROR_STOP=1 \
   -- Separate database for PowerSync bucket storage (avoids schema conflicts with app data).
   -- See https://docs.powersync.com/configuration/powersync-service/self-hosted-instances
   CREATE DATABASE powersync_storage OWNER postgres;
+
+  -- Keycloak's own database, for deployments that run the bundled IdP in
+  -- production mode (`start` rather than `start-dev`). Its own database rather
+  -- than a schema in this one: Keycloak owns ~95 tables and manages them with
+  -- Liquibase, which has no business sharing a namespace with the app's Drizzle
+  -- migrations. Harmless when unused -- an empty database costs nothing.
+  CREATE DATABASE keycloak OWNER postgres;
```
ital0: Blocker: `/docker-entrypoint-initdb.d` runs only when `PGDATA` is empty, so existing Compose volumes, Helm PVCs, and Fargate EFS volumes will never create this database. Helm also mounts a separate init ConfigMap that still stops after `powersync_storage`, so fresh Kubernetes installs miss it too. We need an upgrade-safe provisioning step and the Kubernetes initializer updated in this PR.

ital0: Security blocker: which login will Keycloak use? This creates the database under `postgres`, but no Keycloak role or password exists in the deployment config. Reusing the current Postgres secret would give the IdP superuser access to every database. Please provision a dedicated Keycloak role and secret.

deploy/docker/keycloak.Dockerfile
```diff
@@ -1,5 +1,34 @@
...
+
 FROM quay.io/keycloak/keycloak:26.7
 
+COPY --from=builder /opt/keycloak/ /opt/keycloak/
+
+# `--import-realm` reads this on the first boot against an EMPTY database. Once
+# the realm exists the import is skipped, which is the whole point of moving off
+# H2: changes made in the admin console now survive a redeploy. The flip side is
+# that edits to this file (or to KC_SEED_*) no longer take effect on a database
+# that already holds the realm — change those in the console, or drop the schema.
```
ital0: Non-blocking, but this comment is not accurate. Startup import runs on every boot and skips only a realm that already exists; the whole database does not need to be empty. `KC_SEED_*` is also not used in this repo, and dropping the schema would erase the users and event history this change is trying to preserve. Could we document the exact behavior and remove the destructive advice?

deploy/docker/keycloak.Dockerfile
```diff
@@ -1,5 +1,34 @@
+# Keycloak in production mode (`start`), not `start-dev`.
+#
+# `start-dev` implies the file-backed H2 database, which Keycloak documents as
+# development-only. On an orchestrator without a persistent volume that makes the
+# whole realm ephemeral: every redeploy re-imports the realm JSON and discards
+# everything created since — users added in the admin console, master-realm
+# settings such as brute-force protection, and the login/admin event history. It
+# also leaves `KC_BOOTSTRAP_ADMIN_*` load-bearing on every boot, when it is meant
```
ital0: Blocker for shared previews: this value is also used as the durable password by `per-pr-stack.ts` (`deploy/pulumi/src/shared.ts:604` and `deploy/pulumi/src/per-pr-stack.ts:217-218`). Once the master realm persists, changing `args.keycloakAdminPassword` updates Pulumi but `KC_BOOTSTRAP_ADMIN_PASSWORD` no longer changes the real account. New preview client provisioning will then fail authentication. Can we separate bootstrap credentials from the ongoing admin credential lifecycle?

ital0 (changes requested):
I think the goal is right: move Keycloak state out of the container and into Postgres so users, settings, and audit history survive a redeploy. The optimized image itself follows Keycloak's documented build pattern.

The integration is not complete yet, so I would not merge this version:

- Every checked-in deployment still bypasses or overrides the new production command with `start-dev`.
- None supplies the Postgres URL and credentials. The current HTTP topology also needs `KC_HTTP_ENABLED=true` for edge TLS termination.
- The new database is created only on first Postgres initialization, so existing volumes never get it. Kubernetes also mounts a separate stale init script.
- No dedicated Keycloak database role or secret is provisioned.
- The bootstrap admin password is also used by Pulumi as a permanent admin credential. Once the database persists, rotating that value will break preview client provisioning.
- CI builds the image but does not start this path or test persistence across a restart.

I reproduced the first two startup failures with the image published by this PR. With the deployment override it rebuilt and started in dev mode on H2. With the new default command it exited first for missing HTTPS/HTTP configuration, then for a connection attempt to `localhost:5432`.

Once the runtime wiring, upgrade path, credential lifecycle, and smoke coverage are included, this should deliver the persistence the PR describes.

deploy/docker/keycloak.Dockerfile
```diff
@@ -1,5 +1,38 @@
+# Prod-capable Keycloak image. The default command is `start-dev`, which runs on
+# the file-backed H2 database — fine for local and other ephemeral use. A
+# deployment that needs state to survive a redeploy opts into production mode by
+# overriding the command with `start --optimized` and pointing KC_DB_URL /
+# KC_DB_USERNAME / KC_DB_PASSWORD at Postgres (see the Railway deployment).
```
ital0: Should this say Render instead of Railway?

Since the Render settings live outside this repository, could we add a short note or link describing the runtime configuration here? In particular, it would help to document `start --optimized`, the `KC_DB_*` values, `KC_HTTP_ENABLED`, `KC_HOSTNAME`, and any `KC_PROXY_HEADERS` value used by Render. That would make the production setup reproducible without implying that the self-hosting configs already use it.

deploy/docker/keycloak.Dockerfile
```diff
@@ -1,5 +1,38 @@
...
+#
+# `start-dev` implies H2, which Keycloak documents as development-only: on an
+# orchestrator without a persistent volume the realm is ephemeral — every redeploy
+# re-imports the realm JSON and discards everything created since (admin-console
+# users, master-realm settings such as brute-force protection, and login/admin
+# event history), and it leaves `KC_BOOTSTRAP_ADMIN_*` load-bearing on every boot
+# when it is meant for the first boot only.
```
ital0: Could we document how Render manages the admin credential after the first boot?

I recreated the container against the same Postgres database with a different `KC_BOOTSTRAP_ADMIN_PASSWORD`. The old password still worked, while the new one did not. That behavior is expected, but any automation that needs to log in later will require a separately managed, long-lived credential.

deploy/docker/postgres-init/01-powersync.sh
```diff
@@ -14,10 +14,16 @@ set -eu
 
 : "${POWERSYNC_DB_PASSWORD:?POWERSYNC_DB_PASSWORD must be set on the postgres container}"
 
+# Dev default so a fresh local stack initializes without extra config; a
+# deployment that runs Keycloak on Postgres overrides this and sets the same
+# value as KC_DB_PASSWORD.
+: "${KEYCLOAK_DB_PASSWORD:=keycloak}"
```
ital0, line 20: Since this script is for self-hosting, would it make sense to add `KEYCLOAK_DB_PASSWORD` to `.env.example` and pass it from Compose instead of silently falling back to `keycloak`?

The default is convenient for a disposable local stack, but someone enabling Postgres in a real self-hosted installation could keep it by accident.

deploy/docker/postgres-init/01-powersync.sh
```diff
@@ -33,4 +39,15 @@ psql -v ON_ERROR_STOP=1 \
...
+
+  -- Dedicated role + database for the bundled Keycloak IdP, for deployments that
+  -- opt into production mode (`start --optimized` on Postgres rather than the
+  -- default `start-dev` on H2). A dedicated role so the IdP connects with least
+  -- privilege rather than the Postgres superuser; its own database because
+  -- Keycloak owns ~95 tables managed by Liquibase, which has no business sharing
+  -- a namespace with the app's Drizzle migrations. Harmless when unused -- an
+  -- empty database costs nothing. Opt-in prod deployments set KC_DB_USERNAME to
+  -- this role and KC_DB_PASSWORD to the value above.
+  CREATE ROLE keycloak_role WITH LOGIN PASSWORD :'keycloak_pass';
```
ital0, line 51: How should existing self-hosted installations pick up this change?

`/docker-entrypoint-initdb.d` only runs when `PGDATA` is empty, so upgrading an existing volume will not create the new role or database. If Postgres-backed Keycloak is only supported for fresh self-hosted stacks, a short upgrade note would be enough. Otherwise, we will need a provisioning step that also runs for existing volumes.

deploy/docker/postgres-init/01-powersync.sh
```diff
@@ -33,4 +39,15 @@ psql -v ON_ERROR_STOP=1 \
...
+  CREATE DATABASE keycloak OWNER keycloak_role;
```
ital0, line 52: The Helm chart mounts a separate `postgres-init` ConfigMap over `/docker-entrypoint-initdb.d`, and that version still stops after creating `powersync_storage`.

Could we update that initializer as well, or document that Helm does not support this opt-in yet? As written, a fresh Helm installation will not get the role or database added here.

ital0 (changes requested):
Thanks for the follow-up. 

The remaining questions are limited to self-hosting. `KEYCLOAK_DB_PASSWORD` still falls back silently to `keycloak`, existing Postgres volumes will not run the new initialization, and the Helm chart uses a separate initializer that does not create the new role or database. If some of those paths are intentionally unsupported, documenting the scope and upgrade steps would be enough.

PR #1205, opened 2026-08-06
fix(deploy): cache the PowerSync wasm and stop caching the SPA shell

njbrake:

Two nginx caching gaps that pull in opposite directions.

**`/@powersync/` was matched by no location block**, so PowerSync's wa-sqlite assets fell through to `location /` with no caching directives at all. That is roughly 17MB of content-hashed files revalidated on every boot, when they are as immutable as anything under `/assets/`.

**`index.html` had the opposite problem.** It is the shell that names the hashed entry chunk, so a cached copy pins a client to an old build after a deploy. It now sends `no-cache`, which still permits a conditional request, so the common case is a 304 rather than a re-download.

Deep links pick this up too: the SPA fallback is a URI, so nginx does an internal redirect and re-runs location matching, and a request for `/settings` ends up in the `location = /index.html` block.

### DIFF
--- a/deploy/config/nginx.conf.template
+++ b/deploy/config/nginx.conf.template
@@ -55,6 +55,30 @@ server {
         add_header Cache-Control "public, immutable";
     }
 
+    # PowerSync's wa-sqlite wasm/worker assets, copied into public/@powersync by
+    # the `copy-powersync-assets` Vite plugin. Content-hashed filenames like
+    # `3322bc84de986b63c2cd.wasm`, so they are as immutable as /assets/ -- and they
+    # total ~17MB, which is worth not revalidating on every boot. Without this they
+    # fall through to `location /`, which sets no caching directives at all.
+    location /@powersync/ {
+        include /etc/nginx/snippets/security-headers.conf;
+        expires 1y;
+        # nosemgrep: generic.nginx.security.header-redefinition.header-redefinition
+        add_header Cache-Control "public, immutable";
+    }
+
+    # The SPA shell decides which hashed entry chunk a client loads, so a cached
+    # copy pins the client to an old build. `no-cache` (not `no-store`) still
+    # allows a conditional request, so the common case stays a cheap 304.
+    #
+    # The `/index.html` fallback below is a URI, so nginx does an internal redirect
+    # and re-runs location matching: deep links like /settings land here too.
+    location = /index.html {
+        include /etc/nginx/snippets/security-headers.conf;
+        # nosemgrep: generic.nginx.security.header-redefinition.header-redefinition
+        add_header Cache-Control "no-cache" always;
+    }
+
     # Block source maps (they are uploaded to PostHog for error tracking, not served publicly)
     location ~* \.map$ {
         return 404;
--- a/deploy/config/nginx.conf.template
+++ b/deploy/config/nginx.conf.template
@@ -55,11 +55,25 @@ server {
         add_header Cache-Control "public, immutable";
     }
 
-    # PowerSync's wa-sqlite wasm/worker assets, copied into public/@powersync by
-    # the `copy-powersync-assets` Vite plugin. Content-hashed filenames like
-    # `3322bc84de986b63c2cd.wasm`, so they are as immutable as /assets/ -- and they
-    # total ~17MB, which is worth not revalidating on every boot. Without this they
-    # fall through to `location /`, which sets no caching directives at all.
+    # PowerSync's wa-sqlite workers, copied into public/@powersync by the
+    # `copy-powersync-assets` Vite plugin. Unlike the wasm below, the .umd.js files
+    # (worker/WASQLiteDB.umd.js, worker/SharedSyncImplementation.umd.js,
+    # index.umd.js) have STABLE names and are loaded by fixed path in
+    # src/db/powersync/database.ts, so `immutable` would pin a client to an old
+    # worker across a @powersync/web upgrade -- the same stale-build problem the
+    # index.html block below fixes. Revalidate them like index.html; the common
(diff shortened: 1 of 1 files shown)


deploy/config/nginx.conf.template
```diff
@@ -55,6 +55,30 @@ server {
         add_header Cache-Control "public, immutable";
     }
 
+    # PowerSync's wa-sqlite wasm/worker assets, copied into public/@powersync by
+    # the `copy-powersync-assets` Vite plugin. Content-hashed filenames like
+    # `3322bc84de986b63c2cd.wasm`, so they are as immutable as /assets/ -- and they
+    # total ~17MB, which is worth not revalidating on every boot. Without this they
+    # fall through to `location /`, which sets no caching directives at all.
+    location /@powersync/ {
```
ital0, line 77: One thing worries me here: not everything in this folder is content-hashed. `copy-assets` also emits `index.umd.js` and two workers (`worker/WASQLiteDB.umd.js`, `worker/SharedSyncImplementation.umd.js`) with stable names, and `src/db/powersync/database.ts:108,113` loads those workers by fixed path (the config Safari uses on the web). With `immutable` for a year, a Safari client can stay on an old worker long after a `@powersync/web` upgrade, which is the same stale-build problem the `index.html` block below is fixing.

deploy/config/nginx.conf.template
```diff
@@ -55,6 +55,30 @@ server {
         add_header Cache-Control "public, immutable";
     }
 
+    # PowerSync's wa-sqlite wasm/worker assets, copied into public/@powersync by
+    # the `copy-powersync-assets` Vite plugin. Content-hashed filenames like
+    # `3322bc84de986b63c2cd.wasm`, so they are as immutable as /assets/ -- and they
+    # total ~17MB, which is worth not revalidating on every boot. Without this they
```
ital0: Small thing: the ~17MB counts the `.map` files, and the `.map` rule below already 404s those. What can actually be served is closer to 10MB.

ital0 (changes requested):
Thanks for this one. The diagnosis is right, and the `index.html` part checks out: the 304 flow and the internal redirect for deep links both work as described.

My one concern is the `/@powersync/` block. That folder isn't all content-hashed: the two workers (`worker/WASQLiteDB.umd.js`, `worker/SharedSyncImplementation.umd.js`) have stable names and are loaded by fixed path in `src/db/powersync/database.ts:108,113`, so a year of `immutable` can leave Safari web clients on an old worker after a `@powersync/web` upgrade. That's the same stale-build problem the `index.html` half of this PR fixes. More in the line comment.

Two smaller notes: the ~17MB figure includes `.map` files the map rule already 404s (what can be served is closer to 10MB), and `oauth-callback.html` plus the other `public/` statics still have no cache policy. Both may well be intentional, just flagging.

deploy/config/nginx.conf.template
```diff
@@ -55,6 +55,57 @@ server {
         add_header Cache-Control "public, immutable";
     }
 
+    # PowerSync's wa-sqlite workers, copied into public/@powersync by the
+    # `copy-powersync-assets` Vite plugin. Unlike the wasm below, the .umd.js files
+    # (worker/WASQLiteDB.umd.js, worker/SharedSyncImplementation.umd.js,
+    # index.umd.js) have STABLE names and are loaded by fixed path in
+    # src/db/powersync/database.ts, so `immutable` would pin a client to an old
```
ital0: Small thing in the comment: I could not find any reference to `index.umd.js` in `src/`. `database.ts` only loads the two worker files. The no-cache treatment still fits it either way, since the name is stable.

deploy/config/nginx.conf.template
```diff
@@ -55,6 +55,57 @@ server {
...
+    # case stays a cheap 304. This regex overrides the prefix block below.
+    location ~ ^/@powersync/.*\.umd\.js$ {
+        include /etc/nginx/snippets/security-headers.conf;
+        # nosemgrep: generic.nginx.security.header-redefinition.header-redefinition
+        add_header Cache-Control "no-cache" always;
+    }
+
+    # The wasm alongside those workers IS content-hashed (e.g.
+    # `3322bc84de986b63c2cd.wasm`), so it is as immutable as /assets/ -- roughly
+    # 10MB of servable content worth not revalidating on every boot (the larger
```
ital0: Now that the `.umd.js` files moved to the no-cache block above, this block only serves the wasm, which is about 7.3MB. The 10MB figure still counts the umd files.

deploy/config/nginx.conf.template
njbrake: Addressed in cbfad77f. Split the block so `~ ^/@powersync/.*\.umd\.js$` serves the stable-named workers (and `index.umd.js`) `no-cache`, revalidating like `index.html`; only the content-hashed wasm keeps `immutable`. A `@powersync/web` upgrade can no longer strand a client on an old worker. Also gave `oauth-callback.html` its own `no-cache` rule; left the other `public/` statics (favicons, `capture-worklet.js`) falling through, since they aren't content-hashed and the worklet loads by fixed path.

_Note: this reply was drafted by Claude Opus 4.8 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

deploy/config/nginx.conf.template
njbrake: Fixed in cbfad77f. The comment now says ~10MB servable and notes the larger on-disk figure includes the `.map` files this rule 404s.

_Note: this reply was drafted by Claude Opus 4.8 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #1213, opened 2026-08-12
Update .env.example to include mozilladatacollective.com as waitlist …

moz-kathyreid:
…auto approve

Hi Ryan and team, 
We have Thunderbolt.io on our internal "things to take for a spin" list, I mean our R&D Roadmap ;-), and would like to bypass waitlist rather than having to set up mail sending on localhost.  Best, Kathy Reid


cjroth:
Let me know if you need anything! I'm excited for you to take it for a spin :)

PR #1228, opened 2026-08-17
perf(THU-777): run PowerSync sync stream in a dedicated worker on iOS

arienemaiara:
## Problem

On iOS the PowerSync sync stream runs on the UI thread. During a large catch-up (sync re-enabled after building a backlog on another device), message parsing and decryption block the main thread and the app freezes. [THU-777](https://linear.app/mozilla-thunderbolt/issue/THU-777/app-is-extremely-slow-while-syncing-on-ios)

## Change

Move the sync stream off the UI thread:

- `enableMultiTabs: true` at the PowerSyncDatabase level so PowerSync selects the off-thread `SharedWebStreamingSyncImplementation`. The DB factory keeps its own `enableMultiTabs: false` (SharedWorker is unavailable under `tauri://`).
- A dedicated `Worker` hosts the sync stream. iOS can't use a SharedWorker but a dedicated Worker runs off-thread. `ThunderboltDedicatedSyncImplementation.worker.ts` wires `ThunderboltSharedSyncImplementation` to a `WorkerClient` over `self`.

Parse/decrypt work moves to the `dedicated-sync-*` worker thread; the UI thread stays responsive during catch-up.

The edit lands in the `safari-tauri` database config, which serves every WebKit surface (iOS/macOS Tauri **and** desktop Safari) — they share this branch because WebKit needs `WASQLiteOpenFactory` + `OPFSCoopSyncVFS` and can't use a SharedWorker. So desktop Safari gets the same off-thread sync as a side effect. Chrome/Edge/Firefox (the `default` config) already run sync off-thread via SharedWorker and are unchanged.

## Test

- 53 PowerSync tests pass; typecheck clean.
- Manual (iOS Simulator, plus desktop Safari as the same-config proxy): build a chat backlog on one client, enable sync on the other, confirm the UI stays responsive during catch-up and the sync work runs on the `dedicated-sync-*` worker (Web Inspector → Timelines). Only the simulator/device exercises worker loading under `tauri://`.

## Evidences

https://github.com/user-attachments/assets/eabfebbf-19b0-4bad-af2b-1aca1f75a916

### DIFF
--- a/src/db/powersync/database.ts
+++ b/src/db/powersync/database.ts
@@ -88,16 +88,16 @@ export const getPowerSyncOptions = (
   }
 
   /**
-   * Safari (web) + Tauri (iOS/Desktop): Full WASQLiteOpenFactory required.
-   * OPFSCoopSyncVFS — synchronous OPFS handles. Avoids IDBBatchAtomicVFS + Asyncify
-   * (causes "Maximum call stack size exceeded" on Safari/iOS; JSC has smaller stack than V8)
-   * and SharedArrayBuffer + SharedWorker (exceeds iOS WKWebView memory, black-screen crash).
-   * Explicit UMD worker paths — bypasses import.meta.url which fails under tauri://.
-   * enableMultiTabs: false — dedicated worker, not SharedWorker (fails under tauri://).
+   * Safari (web) + Tauri (iOS/Desktop): full WASQLiteOpenFactory required.
+   * OPFSCoopSyncVFS avoids IDBBatchAtomicVFS + Asyncify ("Maximum call stack size exceeded"
+   * on Safari/iOS) and SharedArrayBuffer + SharedWorker (iOS WKWebView black-screen crash).
+   * Falls back to IDBBatchAtomicVFS when OPFS is unavailable (e.g. WebKitGTK on Tauri Linux).
+   * The DB runs in a dedicated Worker via an explicit UMD path (import.meta.url fails under
+   * tauri:// for node_modules workers).
    *
-   * Falls back to IDBBatchAtomicVFS (IndexedDB-backed, no OPFS dependency) when OPFS itself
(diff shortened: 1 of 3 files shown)


raivieiraadriano92 (dismissed):

src/db/powersync/database.ts
```diff
@@ -88,16 +88,16 @@ export const getPowerSyncOptions = (
...
+   * OPFSCoopSyncVFS avoids IDBBatchAtomicVFS + Asyncify ("Maximum call stack size exceeded"
+   * on Safari/iOS) and SharedArrayBuffer + SharedWorker (iOS WKWebView black-screen crash).
+   * Falls back to IDBBatchAtomicVFS when OPFS is unavailable (e.g. WebKitGTK on Tauri Linux).
+   * The DB runs in a dedicated Worker via an explicit UMD path (import.meta.url fails under
+   * tauri:// for node_modules workers).
    *
-   * Falls back to IDBBatchAtomicVFS (IndexedDB-backed, no OPFS dependency) when OPFS itself
-   * is unavailable — e.g. WebKitGTK (Tauri on Linux) doesn't implement `navigator.storage.getDirectory`,
-   * so OPFSCoopSyncVFS's worker throws on init instead of ever reaching this Asyncify tradeoff.
+   * Sync runs in a dedicated Worker too (THU-777): the PowerSyncDatabase-level
```
ital0: you can remove any task references (e.g. THU-777)

ital0 (changes requested):
I think that deserves deeper tests since you're changing the worker. I remember we had some problems in the past with that (unfortunately I don't recall exactly what it was).

src/db/powersync/database.ts
raivieiraadriano92: Fixed in e347fe77.

PR #1262, opened 2026-09-04
refactor: extract the confidential Pi model core into shared/agent-core

ital0:
## Summary

Extracts the confidential (Tinfoil secure-enclave) Pi model core that lived only in the CLI into `shared/agent-core`, so the web app can adopt the same engine in a follow-up PR. The CLI now consumes the shared core; its observable behaviour is unchanged.

- New `shared/agent-core/confidential-model.ts`: catalog-driven Pi compatibility resolution (vendor/model → Pi catalog entry, e.g. `zhipu`/`glm-5-2` → `zai`/`glm-5.2`, `deepseek`/`deepseek-v4-flash` → `deepseek`; unknown vendors fail at build time), attestation-failure normalisation around an already-attested `fetch`, usage-receipt capture at the provider-stream seam, and the receipt ↔ terminal-message correlation (`createReceiptLifecycle`) with a single injected `submit` transport.
- `PiModelDescriptor` gains a `kind: 'confidential'` variant and `buildAppHarness` builds it (not yet used by the app).
- `cli/src/provider-runtime/tinfoil.ts` keeps only credential handling, `SecureClient` construction, the authenticated fetch and the durable receipt outbox; `usage-receipt.ts` composes the shared lifecycle with the outbox.
- `cli/src/auth/account-client.test.ts`: the caller-deadline case no longer races a real 5 ms timer.

The shared module imports neither the `tinfoil` SDK nor `@tinfoilsh/verifier` (the app code-splits them) and compiles under the CLI's DOM-less tsconfig.

## Accounting

Unchanged: the Tinfoil proxy still issues the signed `X-Inference-Usage-Receipt`; the client captures it on the successful terminal message and posts `{ receipt, promptTokens, completionTokens, totalTokens }` to `inference-usage/receipts`. The CLI keeps its durable outbox.

## Verification

- `bun run type-check`, `bun run --cwd backend type-check`, `bun run --cwd cli typecheck`, `bun run lint`
- `bun run test`, `bun run test:backend`, `bun run test:agent-core`
- `cd cli && bun run test:5x` on Bun 1.4.0 and `ASDF_BUN_VERSION=1.3.14`
- `cd cli && bun run build`
- End-to-end CLI battery against the real backend and enclave (login, managed prompt, confidential GLM prompts with `inference_usage` rows and an empty outbox, device enforcement, outer-hop headers): see the PR comment.

### End-to-end CLI battery (real backend + real enclave)

[...]

PR #1263, opened 2026-09-04
feat: run the app's Tinfoil models on the Pi harness

ital0:
## Summary

Stacked on #1262. The web app now runs every `tinfoil` model on the in-browser Pi harness through the shared confidential core, and the legacy Vercel-AI-SDK `case 'tinfoil'` in `src/ai/fetch.ts` is gone. No feature flag, no coexistence period.

- `src/acp/built-in-adapter.ts`: `tinfoil` joins the Pi providers (with no legacy fallback, so a `toolUsage === 0` tinfoil model still runs on the harness with no active tools). `resolvePiModel` is now async and acquires the attested `SecureClient` first:
  - **System (managed) models** get a `kind: 'confidential'` descriptor: the legacy fetch behaviour is preserved exactly (SSO cookie vs bearer, read per request so a rotated session token is honoured by a cached harness; `X-App-Version`; `handleAppVersionUnsupported`; eviction of a wedged client), attestation telemetry carries `engine: 'pi'`, and receipts flow through the shared receipt lifecycle into the existing authenticated `inference-usage/receipts` POST (best effort, exactly once per terminal message, never for errored/aborted messages).
  - **BYOK (user) models** get a generic `openai-compat` descriptor over their attested client with the user's key: no receipts, no vendor compat (they have no Pi catalog entry).
  - Attestation failure and a missing BYOK key reject loudly instead of falling back.
  - The harness cache signature includes a process-local id of the attested client instance, so eviction/re-attestation rebuilds the harness instead of reusing a dead transport.
- `src/ai/fetch.ts`: legacy tinfoil branch, `includeUsage` special case and the `onStepFinish` receipt call removed; `src/ai/inference-usage-receipt.ts` is now a thin transport for the shared `InferenceUsageReceiptRequest`.
- Eval matrix/README updated: every shipped eval model runs on Pi.

## Accounting

Unchanged contract: the proxy issues `X-Inference-Usage-Receipt`; the client posts `{ receipt, promptTokens, completionTokens, totalTokens }`. The gate is now "system tinfoil model" rather than the GLM literal, so the next confidential model (DeepSeek, follow-up PR) needs no client change here.

## Verification

- `bun run type-check`, `bun run --cwd backend type-check`, `bun run --cwd cli typecheck`, `bun run lint`
- `bun run test`, `bun run test:agent-core`, `bun run test:backend`, `cd cli && bun run test`
- `bun run build`: the `tinfoil` SDK stays in its own async chunk; the entry chunk contains no SecureClient/HPKE code
- Local end-to-end run of the web app (headless Chromium + this backend + the real enclave): OTP sign-in; a managed Opus prompt (one `anthropic` usage row); GLM prompts in one chat (one `tinfoil`/`glm-5-2` `inference_usage` row and one 204 receipt POST per completed model step, the completion request carrying the session bearer and `X-App-Version` and exposing `X-Inference-Usage-Receipt`); a GLM bash tool call (`tool-ok`, `2`) as one sequential tool step plus the final answer, two metered steps; a single attestation across consecutive prompts; a blocked attestation surfacing as a clear UI error with no usage row; and a BYOK Tinfoil model added through the repaired "Test Model" probe and chatted with unmetered.


ital0:
**End-to-end web app validation (headless Chromium + real backend + real enclave)**

Rows: OTP sign-in; managed Opus prompt (one `anthropic` usage row); confidential GLM prompts in one chat (one `tinfoil`/`glm-5-2` `inference_usage` row and one 204 receipt POST per completed model step; the completion request carried the session bearer and `X-App-Version` and exposed `X-Inference-Usage-Receipt`); a GLM bash tool call (`tool-ok`, `2`) as one sequential tool step plus the final answer, two metered steps; a single attestation across consecutive prompts; a blocked attestation surfacing as a clear UI error with no usage row; BYOK Tinfoil added through the Settings "Test Model" probe and chatted with, unmetered (a regression on that probe was found by this run and fixed before opening the PR).

Not exercised: cross-device sync (the local PowerSync replicated the dev database, so the app ran with sync off), PostHog turn telemetry (data collection off on the fresh account), parallel tool calls.

PR #1264, opened 2026-09-04
feat: serve DeepSeek V4 Flash as a confidential model

ital0:
## Summary

Stacked on #1263. Every model served by Tinfoil now uses the secure enclave: DeepSeek V4 Flash becomes the second confidential model, next to GLM 5.2, and its direct (non-confidential) route is retired.

- **Catalog** (`shared/defaults/models.ts`): DeepSeek V4 Flash ships under a fresh id (`01a06dd7-67ee-75be-b957-2b746271c49d`) as `provider: 'tinfoil'`, `isConfidential: 1`, `vendor: 'deepseek'`; the old direct row (`019f227e-…`) leaves the defaults and is swept like V4 Pro was. `defaultModelsVersion` 4 → 5. The model profile keyed on the exported constant follows the new id.
- **Backend**: the proxy can't read the HPKE-encrypted body, so the client names the model out of band in `X-Inference-Model` (catalog model id). The proxy validates it against the confidential catalog (`provider tinfoil` + `isConfidential 1`), uses the resolved identity for admission, the price lookup (`inference_prices` already had both rows) and the receipt claims, and strips the header before the enclave. A missing header is treated as GLM so app builds already in the field keep working (one-line transitional default); an unknown or non-confidential value is a 400. `usage-receipt.ts` no longer carries GLM literals; `deepseek-v4-flash` is gone from `managedDirectRuntimes` and the plain Tinfoil OpenAI client is deleted from the backend. `managedGlmIdentity` had no consumers left and was removed.
- **Clients**: the app's confidential fetch and the CLI's authenticated secure fetch send the header for proxied (system) models only, never for BYOK. The CLI picker lists DeepSeek from the bundled catalog automatically. Docs that said "confidential GLM" now say "confidential models".
- **Tests**: proxy (missing header → GLM, unknown/non-confidential → 400, no upstream leak, per-model admission/receipt), receipt route for both models, direct route rejecting `deepseek-v4-flash`, direct-route fixtures moved to `opus-5`, catalog snapshot. The backend suite's PostHog cleanup was order-dependent (a cached client created by an unrelated route test); tests now own and await the clients they create through an injected, never-cached `posthogClient`.

## Deploy order

Deploy the backend first or together with the clients. A new client against an old backend has its `X-Inference-Model` ignored and is admitted and priced as GLM; an old client whose direct Flash row has not been swept yet gets "Model not found" on that row until its next reconcile. No sync-rule change, no migration, no Render roll: the PowerSync review of the catalog change passed (see the PR comment).

## UX consequence

Threads bound to the old direct Flash row lose their model row (it is swept) and fall back to the user's selected model, as with V4 Pro; they are not migrated. New DeepSeek chats are confidential and therefore created as encrypted threads.

## Verification

- `bun run type-check`, `bun run --cwd backend type-check`, `bun run --cwd cli typecheck`, `bun run lint`
- `bun run test`, `bun run test:backend` (incl. the previously failing seed 2437232382), `bun run test:agent-core`, `cd cli && bun run test:5x` on Bun 1.4.0 and 1.3.14, `cd cli && bun run build`
- Two independent reviews; a PowerSync/reconciliation review of the catalog change
- End-to-end run of both clients against the real backend and enclave (DeepSeek and GLM prompts with `inference_usage` rows and receipts, direct route retired, header validation): see the PR comment.


ital0:
**End-to-end validation of both clients (real backend + real enclave), 10/10 product rows passed**

- CLI: login + managed prompt; confidential DeepSeek prompt (`--model deepseek-v4-flash`) → one `tinfoil`/`deepseek-v4-flash` `inference_usage` row, empty receipt outbox, request carrying `X-Inference-Model: deepseek-v4-flash`, `X-App-Version` and the session bearer; confidential GLM prompt likewise; the direct route no longer serves `deepseek-v4-flash` (backend logs `Model not found`) while `opus-5` still streams 200.
- Proxy header validation on the client path `/v1/tinfoil/chat/completions`: unknown model → 400 naming `X-Inference-Model`; `opus-5` → 400; missing header → admitted as GLM.
- Web: the picker shows exactly one DeepSeek entry, marked private, next to GLM; DeepSeek prompt → one usage row and one 204 receipt with the model header; DeepSeek bash tool call → two metered steps, correct output; GLM regression prompt → one row; the retired row id is absent.

PowerSync/reconciliation review of the catalog change: PASS (same retire-and-replace path as V4 Pro, no sync-rule or migration change; a pre-existing exposure where a device booting offline on the older bundle can resurrect the retired row until the next version bump is noted in the report). Not exercised: cross-device sync with the new catalog, old accounts holding the retired row.

PR #1275, opened 2026-09-10
feat: let operators allow personal access tokens on the confidential routes

darkbanjo:
## What

Adds `CONFIDENTIAL_API_KEYS_ENABLED` (default `false`). When set to `true`, `rejectPersonalAccessToken` stops 403ing an `x-api-key` on `/v1/tinfoil/*` and the usage-receipt route. Default behaviour is unchanged.

## Why

Enables headless service callers. The first one is a Zendesk ticket-research backend using `tinfoil-python` with `base_url` pointed at `/v1/tinfoil`. Verified end to end against production with a web-session bearer before writing this.

## Corrects a wrong rationale documented in three places

The docs said confidential models "depend on session-bound cache material". They don't:

- `userCacheSecret` is generated **client-side** — `randomBytes(32)` in the CLI, `crypto.getRandomValues` in the browser — and never reaches the backend.
- Attestation and HPKE are the caller's. The proxy forwards ciphertext it cannot read.

So the guard is an **authorization** choice, not a cryptographic one — which is why it becomes a flag rather than a removal.

## What the flag actually trades

A PAT lacks a web session's device binding and lives longer, and confidential inference is metered against the account. Enabling the flag means accepting spend on that tier from a long-lived headless credential. Hence default off.

## Non-goal

The CLI still requires `thunderbolt login` for confidential models regardless of the flag, because `resolveAccountCredential` returns no cache secret on the PAT path. Documented, not changed.

## Testing

- Backend typecheck, lint and prettier clean.
- 463 backend tests pass across config / inference / tinfoil / powersync, including two new cases covering the flag-on path on both routes.


darkbanjo:
## Security review notes

A few things I want reviewers to look at deliberately rather than discover later.

**The flag is global, not per-key.** Once `CONFIDENTIAL_API_KEYS_ENABLED` is on, *every* PAT on the deployment reaches the confidential routes — including keys minted for unrelated CI. A per-key Better Auth permission would be the least-privilege design. I chose global for reviewability and to match the `CLI_DEVICE_REGISTRATION_ENABLED` precedent already in this code path, but I'm happy to change it if reviewers prefer.

**Enabling the flag widens who can exploit a pre-existing property.** Token counts on the usage-receipt endpoint are client-supplied and are *not* part of the signed receipt — the receipt signs `eventId`, `userId`, `provider`, `model` and prices only. That's unavoidable: the proxy blind-forwards ciphertext and cannot know the real token counts. The consequence is that confidential-tier quota is advisory — a caller can under-report, or never submit at all. This is reachable by any web session today; what this PR changes is that it also becomes reachable by PATs, which are longer-lived, not device-bound, and run with the api-key plugin's per-key rate limit disabled.

Backstop: the route limiter keys on `user:<id>` at 60/min and a PAT authenticates as its owner, so the account-level limit still applies, and replay is deduped via `onConflictDoNothing` on the event id. I think this is worth a separate ticket rather than folding it into this PR.

**Blast radius of a leaked PAT** grows to the confidential tier while the flag is on. Mitigation is the PAT lifecycle (`api-key/delete`), not device revocation — so short expiry is the recommendation for service keys.

**Verified unchanged:** the SSRF guard on `X-Tinfoil-Enclave-Url`, the upstream header blocklist, guard ordering (still before rate limiting and before any DB or upstream work), the cross-user receipt check, and the confidentiality property itself.

backend/docs/pat-lifecycle.md
```diff
@@ -20,10 +20,30 @@ curl --fail-with-body --silent --show-error \
...
+PAT-authenticated caller can supply its own. What a PAT lacks is the device binding a
+web session carries, and it lives longer; since confidential inference is metered
+against the account, enabling the flag means accepting spend on that tier from a
+long-lived headless credential.
+
+The CLI itself still requires `thunderbolt login` for confidential models regardless
+of this flag: `resolveAccountCredential` returns no cache secret on the PAT path, so
+`createTinfoilBinding` has nothing to namespace with. The flag exists for service
+callers that manage their own cache secret, such as a backend using `tinfoil-python`
+with `base_url` pointed at `/v1/tinfoil`.
```
ital0, line 46: Could we document the usage-receipt step for service callers here? After decrypting the inference response, the caller needs to take the signed receipt from the `X-Inference-Usage-Receipt` response header and POST it with the token counts to `/v1/inference-usage/receipts`.

Pointing `tinfoil-python` at the proxy can give callers a working response while leaving that call's usage unrecorded. The account's quota checks use that usage ledger, so this step matters even when inference succeeds. The CLI already handles it through `createUsageReceiptLifecycle` in `cli/src/provider-runtime/tinfoil.ts`.

A short example showing receipt submission and retry on failure would help someone following these instructions complete the integration.

ital0 (dismissed):

backend/src/inference/usage-receipt-routes.test.ts
```diff
@@ -182,6 +183,25 @@ describe('inference usage receipt routes', () => {
...
+    try {
+      // The flag is read when the routes are constructed, so build inside the override.
+      const enabledApp = new Elysia().use(
+        createInferenceUsageReceiptRoutes({ auth: mockAuth, database, secret, nowSeconds: () => nowSeconds }),
+      )
+
+      const response = await postRaw(enabledApp, '{', { 'x-api-key': 'valid-personal-access-token' })
+
+      // Past the guard: a malformed body is now the caller's problem, not a 403.
+      expect(response.status).not.toBe(403)
```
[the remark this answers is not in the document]
darkbanjo: ⚡ Good call — done in `f334ab3c`. You're right that `not.toBe(403)` would have passed on a 401 or a 500 and told us nothing. The route answers `emptyResponse(400)` deterministically once `parseReceiptRequest` returns null on the `'{'` body, so the assertion is now `expect(response.status).toBe(400)`, matching the tinfoil sibling.

backend/docs/pat-lifecycle.md
darkbanjo: ⚡ Agreed, and added in `f334ab3c` — a new "Reporting usage from a service caller" subsection under the confidential-models section.

It covers the point you raised directly: the proxy forwards ciphertext it cannot read, so it cannot count tokens, and a caller that skips the submission gets a working response whose usage never reaches the ledger the quota checks read. The section documents reading `X-Inference-Usage-Receipt` off the response, sending `X-Inference-Model` on the request (absent it the receipt is priced as `glm-5-3`), a curl example of the `POST /v1/inference-usage/receipts` body, and the retry dispositions per status — retry on 503/429, retain on 401, drop on 403, and 400 will never become valid. It also notes the two-hour receipt lifetime and `eventId` dedupe, so retrying a submission you're unsure landed is safe.

One correction to the pointer: `createUsageReceiptLifecycle` lives in `cli/src/provider-runtime/usage-receipt.ts`, not `tinfoil.ts` — the doc points there, and calls out the durable-outbox shape as the part worth copying.

PR #1282, opened 2026-09-16
feat(THU-781): add xlsx support 

humantorch:
Fixes THU-781.

## What

`.xlsx` files can now be attached in chat. Each sheet is extracted to text before it goes to the model, the same way `.docx` already works.

## Change

- **`package.json` / `bun.lock`**. Adds `read-excel-file` (see below for why). It brings in `saxen`, `unzipper-esm`, `node-int64`, `worker-f`, and a copy of `fflate@0.8.3` used only by `read-excel-file`. Existing lockfile entries are unchanged.
- **`src/files/transformers/xlsx-to-text.ts`** (new). Reads the workbook with `read-excel-file`, loaded only when an xlsx is actually attached, so it stays out of the entry bundle. Output is one `## Sheet: <name>` heading per sheet followed by its rows as comma-separated values. Cells containing a comma, quote, or line break are quoted, so `Stark, Inc.` stays one cell.
- **Dates are formatted in UTC.** `read-excel-file` returns cell dates as UTC midnight, so formatting them in local time showed the previous day anywhere west of UTC. Formatting in UTC reproduces what's written in the cell regardless of the user's timezone. Cells with a time of day keep it (`9/1/2026, 2:30 PM`). Output uses `sourceLocale`, matching other model-facing text.
- **`src/files/transformers/index.ts`**. Registers `xlsxMime` and the `xlsx->text` loader.
- **`src/components/chat/chat-prompt-input.tsx`**. Adds the xlsx MIME type and `.xlsx` extension to the accepted attachments.

Nothing else needed to change:
- The attachment card falls back to the generic file icon with an `XLSX` badge.
- Compression already passes non-PDF files through.
- The built-in agent and ACP agents both look up transformers generically, so they pick up xlsx automatically. The built-in agent extracts text on the first turn. The older AI SDK path tries native bytes first and falls back to text, the same as docx.

## Why `read-excel-file`

| Considered | Status |
| --- | --- |
| **SheetJS (`xlsx`)** | npm is frozen at 0.18.5 (2022). Current releases only ship as a tarball from `cdn.sheetjs.com`, and there's been no tagged release since July 2024. |
| **exceljs** | Last published October 2023. 662 open issues and 146 open PRs. |
| **read-excel-file** | MIT, published normally to npm (last release August 2026), 1 open issue. Read-only, which is all this needs, and it reads a browser `Blob` directly. |

## Limitations / non-goals

- **Hidden sheets are sent to the model.** `read-excel-file` doesn't expose which sheets are hidden, so a hidden sheet's contents are extracted like any other. Users may not know a workbook has hidden sheets, and with a cloud model that content goes to the provider. Skipping them would mean parsing the workbook XML ourselves; worth a decision before or after merge.
- **No legacy `.xls`**, only `.xlsx`, matching the existing attachment support, which accepts `.docx` but not `.doc`.
- **Formulas come through as their saved values** (what Excel shows), not recalculated.
- **No preview.** Opening a sent xlsx shows "Preview not available for this file type."
- **No size cap on extracted text.** This is also true of PDF and docx today, but spreadsheets are the format most likely to be huge. A very large workbook will parse on the main thread and may overflow the context window. `read-excel-file` has a Web Worker build if that becomes a problem.
- **Empty sheets** appear as a bare heading with no rows.

## Testing

- `xlsx-to-text.test.ts`: single sheet, multiple sheets, empty cells, CSV quoting, and dates (calendar day, time of day, seconds) run with `TZ=America/Halifax`, since `bun test` runs in UTC and would hide timezone bugs. The date tests fail against the pre-fix code.
- `index.test.ts`: registry assertions for xlsx alongside docx.
- Typecheck, lint, and Prettier clean. Full frontend suite passes across repeated randomized runs.
- Manual:
  - A single-sheet order spreadsheet with Llama 3.2 3B via Ollama (loopback path) and Claude Haiku 4.5 via `/v1/proxy`. Confirmed in the Network tab that the request body contains the extracted sheet text, and that the first turn in a fresh Haiku chat sent one request with no fallback.
  - A workbook with multiple sheets, including a hidden and an empty sheet, with Haiku. Answers drew on the second sheet, and dates match the source file after the UTC fix.


src/files/transformers/xlsx-to-text.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Extracts a plain-text table dump from an XLSX blob via a lazily-imported
+ * read-excel-file. Multiple sheets are concatenated with a heading per sheet so
+ * the model can tell them apart; rows are joined as CSV-ish text rather than a
+ * Markdown table, since spreadsheet column counts and widths vary too widely
+ * for a fixed-width table to stay readable.
+ */
+export const xlsxToText = async (file: StoredFile): Promise<{ text: string }> => {
+  const { default: readXlsxFile } = await import('read-excel-file/browser')
+  const sheets = await readXlsxFile(file.blob)
```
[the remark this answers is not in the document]
humantorch: As I see it, the behavior described above was present in pre-9.0.0 versions of `read-excel-file`, this PR installs 9.3.10. In 9.0.0 the default export returns all available sheets as an array of objects: `[{ sheet, data }, ...]`. Verified against the real package with a four-sheet workbook, unmocked, and in-app with a multi-sheet file.

package.json
```diff
@@ -154,6 +154,7 @@
     "react-pdf": "^10.4.1",
     "react-resizable-panels": "^4.10.0",
     "react-router": "^7.12.0",
+    "read-excel-file": "^9.3.10",
```
ital0: I haven't researched... But something we used to take care was to final size of the JS bundle. Also, we check the maturity and maintanability of the libs. 
Looks like this one is well-maintaned.

But just to confirm, is it really the best option?

cjroth (changes requested):
Looks good! I tested this and it works. I looked into read-excel-file and I think it makes sense to use it, though I do hate to add another third party dependency - especially one that doesn't have that much activity. It adds negligible bundle size and is lazy loaded anyways which is great.

One thought that I think is out-of-scope for this PR but worth keeping in mind for a separate task is that I think some models, possibly including the current GLM 5.3 Flash one, do support xlsx natively. I suspect that their native support would be even better than translating to text first since the AI could see the raw tokens. That said, that's a separate optimization worth exploring that spans more than just xlsx and requires us to map model <> format support which is something @darkbanjo has explored a bit.

Anyways, there's a few minor bugs that my AI code review (Fable 5.1) found which I verified with Gemeni Flash. Worth fixing these, but they aren't major. Here's the full output:

```
  The PR itself is small, consistent with the existing docx transformer pattern, and passes tests/tsc/eslint. The bundle work is correct. But I found one blocking issue in the Tauri desktop/mobile builds, one real correctness bug in date
  output, and a UX design issue inherited from the "native-first" ladder that is worse for xlsx than for anything else.

  1. Blocking: hangs forever on desktop/mobile for sheets > ~5k rows

  read-excel-file/browser unzips with fflate's async unzip(), which spawns a blob: URL Web Worker whenever any zip entry is ≥512 KB uncompressed and compresses below 80%. Sheet XML always compresses well, so any sheet XML over 512 KB (≈5–8k
  rows) takes that path. Our Tauri CSP is script-src 'self' 'wasm-unsafe-eval' with no worker-src, so blob: workers are refused.

  I verified this with Playwright against the actual production chunk (dist/assets/browser-CkmlcdZy.js), in both engines, with the Tauri CSP as a meta tag:

  chromium  nocsp  big.xlsx  ok 34ms (8001 rows)
  chromium  csp    big.xlsx  TIMEOUT — "Creating a worker from 'blob:…' violates CSP directive script-src"
  webkit    nocsp  big.xlsx  ok 40ms
  webkit    csp    big.xlsx  TIMEOUT — "Refused to load blob:… worker-src directive"

  Critically, the promise never settles — fflate sets worker.onmessage but not onerror, so a CSP-refused worker is silent. In the app that means hasExtractableText in use-attachment-remediation.ts awaits forever, remediating stays true, and
  the user gets a permanent loading state with the error suppressed. Web is unaffected (no CSP header there). Small sheets work everywhere.

  Options: (a) add "worker-src": "'self' blob:" to tauri.conf.json — smallest change, modest CSP widening; (b) don't use read-excel-file's browser entry and unzip ourselves with fflate.unzipSync (ties into §4). Either way, this needs a
  real-browser test of the large-file path; the PR's tests mock the library entirely so nothing covers it. Consider adding a timeout around transformer calls in the remediation hook regardless.

  2. Bug: dates off by one second

  Running the real library on a fixture (unmocked), a cell Excel serializes as 46266.604166666664 (= 14:30) came out as "9/1/2026, 2:29:59 PM". read-excel-file does Math.floor(86400000 * (serial - 25569)), so float error lands at
  14:29:59.999. The PR's hasTime/"seconds only when non-zero" logic amplifies it: spurious :59 seconds and off-by-one minutes will be common. Round to the nearest second before formatting.

  Related suggestion: emit ISO 8601 (2026-08-01, 2026-09-01 14:30) instead of toLocaleString(sourceLocale, …). It's unambiguous for the model (8/1/2026 reads as D/M in half the world), is simpler code, and removes the ICU narrow-NBSP quirk
  the tests have to paper over with \s.

  3. UX: xlsx always costs a failed model round-trip

  defaultDeliveryMode(xlsxMime) is undefined → native bytes go out first. No provider (Anthropic, OpenAI, Google) accepts a spreadsheet as an inline file part, so every first send of an xlsx is a guaranteed 400 → auto-remediation → re-send.
  The lazy import + parse happens inside that suppressed-error window, so it isn't a visible "weird lag" — but the whole dance is a wasted network round trip plus a parse that runs twice (once for hasExtractableText, again in hydration).
  docx has the same problem today, but the fix is a one-liner here: return 'text' from defaultDeliveryMode for xlsxMime (and arguably docx as a follow-up). That also makes the lazy chunk load at send time, which is the natural moment.

  Other observations: no size cap — my 8k-row fixture yields ~480 KB of text (≈150k tokens), which becomes a context-overflow error that remediation explicitly doesn't handle. Historical xlsx attachments re-parse on every subsequent turn
  (pre-existing for all transformers). Fully empty rows emit , , , ,  noise. Percent/currency formatting is lost (0.15, not 15%) — an inherent read-excel-file limitation.

  4. The package: read-excel-file@9.3.10

  Provenance — verified. I cloned gitlab.com/catamphetamine/read-excel-file at tag v9.3.10, rebuilt with the repo's own Babel config, and diffed against the npm tarball: browser/, node/, universal/, web-worker/, types/, modules/, commonjs/,
  and all root files are byte-identical modulo CRLF (author is on Windows; tests are excluded from publish). No install scripts anywhere in the new subtree. No fetch/eval/Function()/env access/obfuscation in the shipped code; the only
  dynamic code is Function.prototype.toString() → blob Worker (fflate's, and worker-f's).

  Trust. Author (catamphetamine / Nikolay Kuchumov) is a long-standing maintainer (libphonenumber-js etc.), package since 2018, 1.5M weekly downloads, MIT, actively maintained (Aug 2026). fflate 0.8.3 is the patched version for
  GHSA-px8p-9vwx-vf98 — the bun audit hit is posthog-js's root fflate@0.4.8, pre-existing and unrelated. Nothing malicious.

  Quality — mixed. The parsing core is solid and well-commented, handles the real edge cases (1904 epoch, sparse rows, inline strings, rich-text runs, custom numFmt date detection, missing sharedStrings). Concerns:
  • Dependency churn: v9.x pulled in two fresh forks by the same author — worker-f (created July 2026, already at 0.1.20) and unzipper-esm (June 2026). worker-f is imported in the browser path but its own JSDoc says createWorkerFunction — 
    Not used.: dead weight that can't tree-shake, yet still ships a second blob-Worker code path into our chunk.
  • It vendors saxen and lists it as a dependency (dependency unused in the browser path).
  • unzipper-esm, graceful-fs, node-int64 install but don't bundle (node-only paths). Fine, just noise.
  • A 20-year-old-style .then() codebase compiled through Babel with _wrapNativeSuper helpers ×5. Chunk is 52 KB / 16 KB gz — acceptable.

  Roll our own? The browser path is really just: fflate unzipSync (we control it → no blob worker → no CSP issue) + DOMParser (built into the browser, no SAX needed at our sizes) + workbook/rels/sharedStrings/styles/sheet parsing. Note
  jszip is already in our tree via mammoth, so we'd add zero deps. The hard 20% is date-format detection from numFmt codes (~300 lines in read-excel-file). My take: it's a reasonable pick for now given verified provenance and a tiny chunk,
  but the worker-f churn plus the CSP hang mean it's worth pinning exact (9.3.10, no caret) and I'd put "own ~250-line reader" on the table if option (b) in §1 is preferred over widening CSP. SheetJS isn't a good alternative (npm build
  stale at 0.18.5 with unpatched advisories); ExcelJS is Node-oriented and ~1 MB.

  5. Bundle / laziness — verified correct

  bun run build: entry → dynamic → xlsx-to-text-*.js (718 B) → dynamic → browser-*.js (52 KB / 16 KB gz, fully self-contained). The entry chunk contains zero read-excel-file, fflate-0.8, or worker-f code. The double await import (registry →
  module → library) is consistent with docx-to-text.ts.

  Minor

  • xlsx-to-text.test.ts mocks the library entirely, so there's no fixture-based test; a tiny committed .xlsx fixture would have caught the date bug. Happy to hand over the fixture generator I wrote if useful.
  • Side viewer will hit the 'unsupported' branch for xlsx; the card shows a generic icon with no preview. Fine as follow-ups.
  • chat-prompt-input.tsx comment update is good; consider a Sheet icon in FileCard later.

```

package.json
cjroth: This was my first reaction as well - I checked and it is both very small (tens of kb) and lazy-loaded, so I'm not concerned.

humantorch: Agreed with @cjroth below that I don't love adding yet another external dependency here, but it was the best of all the options I looked at. Aside from the other two I mentioned above (SheetJS, exceljs) there were a number of others that I found, none of which had been updated in 5+ years. 

I briefly explored rolling our own file handler here instead of using a third-party kit, that's possibly a good option in the long run but definitely has a steeper investment cost up front.

darkbanjo (dismissed):

humantorch:
Added some fixes for the issues @cjroth posted: times are now accurate to the second, xlsx files are sent as text from the start to avoid wasting a round-trip, and empty rows are skipped. `read-excel-file` is also version-pinned to the current release (9.3.10).

The bigger question of rolling our own xlsx parser is still unresolved and probably out of scope for this PR. Worth having separately, IMO. This also impacts the desktop/mobile hang on large sheets: web is fine (tested locally with a 10k-row sheet), but in Tauri it needs either a CSP change (worker-src 'self' blob:) or our own parser. 

I'd suggest merging this as-is and fixing the Tauri hang in a follow-up, since web works today.

src/components/chat/chat-prompt-input.tsx
```diff
@@ -86,6 +87,7 @@ const acceptedAttachmentExtensions = [
   '.webp',
   '.gif',
   '.docx',
+  '.xlsx',
```
ital0, line 90: Could we also normalize the MIME type for `.xlsx` when `File.type` is empty or `application/octet-stream`? The extension check accepts the file, but `resolveTextMimeType` leaves those values unchanged, so the XLSX transformer isn't found. I reproduced the helper behavior in isolation, though I haven't tested this through the browser. A fallback to the XLSX MIME type and a test for these two cases would cover it.

ital0 (changes requested):

src/components/chat/chat-prompt-input.tsx
humantorch: Looks like the .docx parser has the same issue so I'll include a fix for that one here as well.

ital0 (approved):

darkbanjo (approved):

